PR #5, opened 2016-11-25
common-SLES12-2016.10: sles12 deps for scorep

jgphpc:
## Logs
```  
X=/apps/common/UES/sandbox/jgp/easybuild.dom/gpu/easybuild/software

$X/Score-P/3.0-CrayCCE-2016.11/easybuild/easybuild-Score-P-3.0-20161125.170105.log
$X/Score-P/3.0-CrayGNU-2016.11/easybuild/easybuild-Score-P-3.0-20161125.165525.log
$X/Score-P/3.0-CrayPGI-2016.11/easybuild/easybuild-Score-P-3.0-20161125.170009.log
$X/Score-P/3.0-CrayIntel-2016.11/easybuild/easybuild-Score-P-3.0-20161125.170155.log
```


jenkins-builds/common-SLES12-2016.10
```diff
@@ -14,4 +14,13 @@ likwid-4.1.2.eb
 HDFView-2.13.0.eb
 ctags-5.8.eb
 lynx-2.8.8rel.2.eb
-tmux-2.3.eb
\ No newline at end of file
+tmux-2.3.eb
+# --- scorep
```
jgphpc: Is it because of jenkins ?

jenkins-builds/common-SLES12-2016.10
```diff
@@ -14,4 +14,13 @@ likwid-4.1.2.eb
...
+zlib-1.2.8-scorep.eb
```
jgphpc: score-p depends on compiler&mpich, there is no such thing as a generic version.

gppezzi: then what is the point of having generic dependencies on a separate software stack? just to save build time?

gppezzi:
```
perettig@daint101:~> ls -ld /apps/common/UES/sandbox/jgp/
drwx--S--- 127 piccinal csstaff 8192 Nov 25 12:51 /apps/common/UES/sandbox/jgp/
```

jenkins-builds/common-SLES12-2016.10
jgphpc: > we need to build 'common' and then the system-specific one
This does not make sense: the common stack is compatible with SLES12 based systems and all compilers hence i do not get why you would need to rebuild it for a specific system...

jgphpc: > we need to change all jenkins projects.

why ? just need modulepath to point to /apps/common/UES/SLES12/easybuild/modules/all

gppezzi:
Tests passed on dom and daint by doing:
```
module rm PrgEnv-cray/6.0.3
module use /opt/cray/pe/craype/2.5.8/modulefiles
```

Looks good, thanks @jgphpc !

PR #7, opened 2016-11-30
Improved EasyBuild-custom and production scripts

victorusu:
Changes to EasyBuild-Custom and EasyBuild modules

- The following variables were set to use the /dev/shm folders (improving performance)
EASYBUILD_TMPDIR            /dev/shm/$USER/easybuild/stage/tmp
EASYBUILD_BUILDPATH         /dev/shm/$USER/easybuild/stage/build

- $CRAY_CPU_TARGET added to the default user EASYBUILD installation folder.
This allows the user to know for which system he installed his own EB configfiles.
This change only affects CRAY systems.

- Removed the definition of EASYBUILD_TMP_LOGDIR. This flag definition was unnecessary.

BUGFIX:
- The MODULEPATH now depends on the EASYBUILD_INSTALLPATH variable, as it should be.

- Updated the jenkins production scripts to use the module EasyBuild.
This removes the need of `setup.sh`. Which is incompatible with the EasyBuild modules.
Now one only needs to maintain a single language to perform the EasyBuild installation.

### DIFF
--- a/easybuild/easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb
+++ b/easybuild/easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb
@@ -66,19 +66,24 @@ set host_machine [regsub -all {[-]$} $host_machine ""]
 # the following variables should depend on it
 #
 # EASYBUILD_PREFIX    
-# EASYBUILD_BUILDPATH 
-# EASYBUILD_TMPDIR    
-# EASYBUILD_TMP_LOGDIR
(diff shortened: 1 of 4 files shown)


lucamar:
Great..., but why so many subfolder under /dev/shm/$USER? Is `easybuild/stage` really necessary?

victorusu:
The use of stage is just to avoid any pollution if someone already has a easybuild folder inside the /dev/shm/$USER. Because easybuild does not have a stage folder

gppezzi:
Let's test this on the regression builds.
Thanks @victorusu!

PR #9, opened 2016-12-01
Improving EasyBuild-custom module ecosystem

gppezzi:
The commit applies the following changes to the `module/Easybuild` and
`easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb` files:
 - The EB_CUSTOM_PREFIX variable was removed and the functionality now
   depends on the EASYBUILD_PREFIX variable
 - The EASYBUILD_INSTALLPATH is set to EASYBUILD_PREFIX
 - In order to change the install dir one needs to set the
   EASYBUILD_PREFIX variable before the module load
 - EASYBUILD_PREFIX is unset automatically after module unload
 - EASYBUILD_INSTALLPATH does not append the easybuild folder
   automatically anymore. It has to be set when defining the
   EASYBUILD_PREFIX variable
 - Fixed bug when defining the EASYBUILD_PREFIX on non-cray systems
 - CRAY specific section was moved to the system dependent variable
   definition inside the module
 - Improved code style

General commit changes:
 - Removed the symbolic link `setup.sh` that used to point to
   `scripts/setup.sh-stable`
 - Removed the old daint-modulefile
 - Updated the `jenkins-builds/production_builds-dom.sh` and
   `jenkins-builds/production_builds.sh` scripts to account for the
   `module/Easybuild` changes

Now the EB_CUSTOM_REPOSITORY variable is the only one that is not unset
after module unload.


gppezzi:
@lucamar @jgphpc could you please review the new setup?

PR #12, opened 2016-12-02
sles11: my .ebs

jgphpc:
sles11: my .ebs


gppezzi:
do you want also to update the [sles12 production list](https://github.com/eth-cscs/production/blob/master/jenkins-builds/common-SLES12-2016.10)?

jgphpc:
see #48 #49 #50

PR #15, opened 2016-12-06
Use PyRegression updated daily by jenkins

gppezzi:
instead of the static tarball from master

@vkarak  @victorusu


vkarak:
Looks good to me.

PR #26, opened 2016-12-13
new ddt version (#ALLINEA-988)

jgphpc:
ddt/612-1 fixes a slurm issue on daint.
ddt is already installed, just need to install the new modulefile using jenkins (?)


gppezzi:
Can you point out what is the difference/fix? Since this is a new file we cannot easily spot the change on the PR

jgphpc:
Can you see [#988](https://allinea.atlassian.net/servicedesk/customer/portal/1/ALLINEA-988?sda_source=notification-email) ? ( _just needs to sign up for an account_ )

PR #28, opened 2016-12-15
new modules for simple installs of VMD

jfavre:
VisIt and VMD modules should be installed along with the other EasyBuild packages


easybuild/easyconfigs/v/VMD/VMD-1.9.3.eb
```diff
@@ -0,0 +1,17 @@
...
+name = 'VMD'
+version = '1.9.3'
+
+homepage = 'http://www.ks.uiuc.edu/Research/vmd/'
+description = """VMD is a molecular visualization program for displaying, animating, and analyzing large biomolecular systems using 3-D graphics and built-in scripting"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+modtclfooter = """
+append-path     PATH   /apps/daint/UES/VMD/1.9.3/bin
```
gppezzi: @jfavre were these binaries compiled on Daint? 

I'm just mentioning because we usually prefix builds with OS version (such as `/apps/daint/UES/6.0.UP02/`). 

The current location should be fine for pre-packaged binaries.

jfavre:
These binaries were extracted from tar files I got from the VMD site. There were not build on daint, thus should not be prefix-ed by the OS.

PR #30, opened 2016-12-16
Adding of Visit 2.12.0 recipe

victorusu:
Adding the recipe for visit 2.12.0.
In order to install visit, one need to install mesa.
And it depends on the libxml2 with python support. Therefore
the recipe for libxml2 with python support was added.

On daint, visit with easybuild sometimes fails to install when using more
than 16 processors. Not sure why.


gppezzi:
please also add `Visit-2.12.0-CrayGNU-2016.11.eb` to the production lists:
* https://github.com/eth-cscs/production/blob/master/jenkins-builds/6.0.UP02-2016.11-gpu and
* https://github.com/eth-cscs/production/blob/master/jenkins-builds/6.0.UP02-2016.11-mc

easybuild/easyconfigs/v/Visit/Visit-2.12.0-CrayGNU-2016.11.eb
```diff
@@ -0,0 +1,70 @@
...
+configopts += ' ${CRAY_PMI_POST_LINK_OPTS}"; '
+
+prebuildopts  = ' cd build; '
+buildopts = 'install'
+
+dependencies = [ ('Python', '2.7.12'),
+                 ('Boost', '1.61.0', '-Python-2.7.12'),
+                 ('libxml2', '2.9.3', '-Python-2.7.12') ]
+
+# CMake and libxml2 dependencies with dummy toolchain
```
gppezzi: Here you mention that you use the dummy toolchain, but you're not actually using it. 

FYI the magic syntax for using dummy is:
```
('libxml2', '2.9.3', '-Python-2.7.12', 'True')
```

My suggestion is to leave as it is now to avoid re-testing everything and just fix the comment.

victorusu: Ok. I have done that.

easybuild/easyconfigs/v/Visit/Visit-2.12.0-CrayGNU-2016.11.eb
```diff
@@ -0,0 +1,106 @@
...
+# rm -rf visit2.12.0.tar.gz
+# tar cfz visit-2.12.0.tar.gz visit2.12.0
+# mkdir -p $EASYBUILD_SOURCEPATH/v/Visit
+# cp visit-2.12.0.tar.gz $EASYBUILD_SOURCEPATH/v/Visit/
+#
+# NOTES:
+# * The new visit package is named visit-2.12.0.tar.gz, not visit2.12.0.tar.gz
+#   as the original file.
+# * The download-only option must not be included in the actual build.
+# * The configuration options are the same as the ones below. Execept for the
```
gppezzi: "Except"

easybuild/easyconfigs/v/Visit/Visit-2.12.0-CrayGNU-2016.11.eb
```diff
@@ -0,0 +1,106 @@
...
+# In order to use the pre-downloaded version use the source below
+sources = [SOURCELOWER_TAR_GZ]
+
+srcdir='../src'
+preconfigopts  = ' export PAR_COMPILER=`which cc`; '
+preconfigopts += ' export PAR_INCLUDE="-I$MPICH_DIR/include"; '
+preconfigopts += ' ./src/svn_bin/build_visit --console --no-visit '
+preconfigopts += ' --system-cmake --h5part --no-pyside --system-python --alt-boost-dir $EBROOTBOOST '
+preconfigopts += ' --vtk --mesa --parallel --hdf5 --silo --szip --makeflags -j20 '
+preconfigopts += ' --netcdf --qt <<< "yes" && '
```
gppezzi: 👍  for the `yes` hack

easybuild/easyconfigs/v/Visit/Visit-2.12.0-CrayGNU-2016.11.eb
```diff
@@ -0,0 +1,103 @@
...
+
+# In order to use the original version uncomment these lines
+#source_urls = ['http://portal.nersc.gov/project/visit/releases/%(version)s' ]
+#sources = ['visit2.12.0.tar.gz']
+
+# In order to use the pre-downloaded version use the source below
+sources = [SOURCELOWER_TAR_GZ]
+
+srcdir='../src'
+preconfigopts  = ' export PAR_COMPILER=`which cc`; '
```
gppezzi: Do you really need the full path to the cc wrapper? 

Normally we should use the variables defined by easybuild for specifying the compilers:
```
  export CC="cc"
  export CXX="CC"
  export F77="ftn"
  export F90="ftn"
```

gppezzi:
Awesome, thanks @victorusu and @jfavre! Let's try this out.

PR #37, opened 2016-12-20
amber qe vasp

lucamar:
Several users required the non CUDA version of Amber, QuantumESPRESSO and VASP to run on the hybrid XC50, as they need these versions for their simulations and they can't access the multicore XC40.


gppezzi:
👎  but we have no better alternative

PR #44, opened 2017-01-04
Adding Tinker + OpenMM with cuda support

victorusu:
Due to a bug on intel 17 compilers, one cannot install TINKER using
the current CrayIntel-2016.11 toolchain. Therefore I am adding the
CrayIntel-2016.11-intel16 toolchain.

Important aspects of this commit:
* Installation of TINKER + OpenMM (dynamic_omm program) and *pure TINKER*
programs are performed. One can compile *pure TINKER* only but setting
the openmm flag to False.
* It is not possible to install only TINKER+OpenMM with this release.
* OpenMM is cuda only and has *NO* python wrappers

### DIFF
--- a/easybuild/easyblocks/tinker.py
+++ b/easybuild/easyblocks/tinker.py
@@ -1,0 +1,227 @@
+##
+# Copyright 2009-2016 Ghent University
+#
+# This file is part of EasyBuild,
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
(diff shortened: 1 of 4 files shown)


victorusu:
Compilation logs can be seen using:
```
cat /apps/daint/UES/6.0.UP02/sandbox-vh/easybuild/software/TINKER/7.1.3-CrayIntel-2016.11-intel16-OpenMM-7.0.1-cuda-8.0/easybuild/easybuild-TINKER-7.1.3-20170104.134739.log
```

PR #46, opened 2017-01-09
Gsl cce

victorusu:
Compilation log can be see at:
/apps/daint/UES/6.0.UP02/sandbox-vh/easybuild/software/GSL/2.2.1-CrayCCE-2016.11/easybuild/easybuild-GSL-2.2.1-20170109.085043.log


gppezzi:
ok to test

PR #62, opened 2017-01-20
Netcdf python version bump

gppezzi:


jgphpc:
retest this please

gppezzi:
@jgphpc CI will only test new/modified ebs. This PR is only for checking if you are ok with removing the old versions from the list and replacing with the new ones ( if possible we want to avoid keeping multiple versions on the production list)

PR #66, opened 2017-01-23
WIP: Perftools-lite with Amber, GROMACS and LAMMPS

victorusu:
This commit adds the CUDA and non-CUDA versions. Even though
not all of these options are working prorpely. GROMACS and LAMMPS
non-CUDA versions are able to produce performance analysis.
While the rest no.


jgphpc:
Note for next time: split into multiple PRs so we can build in //

victorusu:
This PR is old and is not correct any more. Can you delete it please?

PR #68, opened 2017-01-24
adding XDMF reader version 3 in ParaView and the alternate reader in …

jfavre:
enabling new generation Xdmf readers in visualization softwares ParaView and Visit


gppezzi:
@jfavre there's a syntax error on your ParaView easyconfig file, my first guess would be the Boost dependency, so I suggest:
```
    ('Boost', '1.61.0', '-Python-2.7.12'),
```

gppezzi:
Syntax now looks good, but currently we don't have Boost/1.61.0-CrayGNU-2016.11 (without python) available on the repository.

You would need to either create Boost-1.61.0-CrayGNU-2016.11.eb or to add the `-Python-2.7.12` suffix to this dependency (as I suggested above).

PR #85, opened 2017-02-02
Boost 1.63.0 with python 2 and python 3 support

victorusu:


gppezzi:
thanks @victorusu 

Don't forget to delete boost.py once we upgrade EB.

PR #86, opened 2017-02-02
TensorFlow 0.12.0 without GPU support

gppezzi:


gppezzi:
@mschoengens @victorusu I've managed to find out the offending export by running the EB extended dry run (-x), copying the exports to the terminal and trying the build manually.

It seems that the programming environment gets messed up if we tell bazel to use the cray wrappers (see fix [here](https://github.com/eth-cscs/production/pull/86/commits/3ee5b5e60c708dc3f7a6091f2da0cd7984b5aec4#diff-54d942a4b96d6e2c1d83ad2a1a1874bbR35)).

PR #87, opened 2017-02-03
added h5py which I need in the Programmable Sources of ParaView

jfavre:


gppezzi:
@jfavre I've did the sync with master just to speedup the CI testing

PR #93, opened 2017-02-10
Adding GPAW-1.2.0 easyconfig and patch files with updated dependency list

lucamar:
Tested locally


lucamar:
Thanks @vkarak and @gppezzi , I'm going to merge the request.

PR #94, opened 2017-02-10
Refactored TensorFlow patches such that library paths are dynamically…

mschoengens:
… created depending on which module is loaded.


gppezzi:
ok to test

gppezzi:
@mschoengens this does not work, as Bazel doesn't seem to accept environment variables here.

I think we need to take a deeper look on how to pass arguments to Bazel if we want to avoid hardcoded paths.

mschoengens:
The environment variables are substituted with the real path by sed (see the patch file)

When I run it in the terminal everything works fine -- can you send me the Jenkins error log?

gppezzi:
Sorry I've missed the sed...

The error was on the multicore version (branch was not in sync with master), so it was unrelated with this PR.

I've synced with master and now it looks good. Thanks @mschoengens

PR #97, opened 2017-02-14
Cp2k fix

lucamar:
Adding CP2K-trunk-CrayGNU-2016.11-cuda-8.0.54 to fix a bug in CP2K release 4.1: the fix is available only in the trunk SVN, therefore I needed to package the source tarball separately.

### DIFF
--- a/easybuild/easyconfigs/c/CP2K/CP2K-svn-CrayGNU-2016.11-cuda-8.0.54.eb
+++ b/easybuild/easyconfigs/c/CP2K/CP2K-svn-CrayGNU-2016.11-cuda-8.0.54.eb
@@ -2,7 +2,7 @@
 easyblock = 'MakeCp'
 
 name = 'CP2K'
-version = "trunk"
+version = "svn"
 
 homepage = 'http://www.cp2k.org/'
 description = """CP2K is a freely available (GPL) program, written in Fortran 95, to perform atomistic and molecular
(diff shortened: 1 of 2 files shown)


lucamar:
Successful after shortening the buildpath of user jenscscs! Thanks @gppezzi

gppezzi:
Good but I suggest you add a date/tag to the version (if that does not break the build 😄 )

easybuild/easyconfigs/c/CP2K/CP2K-svn-CrayGNU-2016.11-cuda-8.0.54.eb
```diff
@@ -0,0 +1,59 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'MakeCp'
+
+name = 'CP2K'
+version = "svn"
```
gppezzi: how about:
version = "2017.02"? 

Unless you don't plan to make it public (in that case you can leave it like this).

lucamar: Right now I was not planning to build it with Jenkins for the general public in production (it is not a stable official release), but just send it to two users to fix their bug. I hope the CP2K developers will soon publish a new stable release with the bug fix.

PR #104, opened 2017-02-23
perftools-cscs/645 for daint

jgphpc:


lucamar:
Awesome, thank you @jgphpc !

PR #107, opened 2017-02-27
Adding Qt-5.8.0 easyconfigs with updated easyblock

lucamar:
Qt release 5.8.0 accepts '-prefix installdir' in place of '--prefix=installdir', therefore a change in the easyblock qt.py was necessary in order to check if the easyblock specific variable 'prefix_opt' is defined in the easyconfig file.

### DIFF
--- a/easybuild/easyblocks/qt.py
+++ b/easybuild/easyblocks/qt.py
@@ -1,0 +1,134 @@
+##
+# Copyright 2013 Ghent University
+#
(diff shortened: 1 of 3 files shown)


lucamar:
I need to check why the build failed for Jenkins (gzip error unzipping .tar.gz), as I managed to build it successfully with my `$USER`.

lucamar:
Test again

PR #110, opened 2017-02-28
fixed problem due to renaming Python to Python-bare

gppezzi:
Plus some cleaning...

Also see: https://github.com/hpcugent/easybuild-easyblocks/issues/1063


gppezzi:
I've successfully installed packages with pip on both daint and kesch using this recipe.

I suppose that the PIP certificates workaround is not needed anymore and we can also remove it from the other python recipes by the next time we upgrade the software stack.

lucamar:
Thanks @gppezzi

PR #117, opened 2017-03-07
ray tracing libraries used by ParaView (5.3 and above)

jfavre:
These two libraries provided by Intel are used by ParaView for real-time raytracing. This will be the first version on daint (v5.3) to support it.


gppezzi:
Looks good, thanks @jfavre

PR #121, opened 2017-03-13
PyRegression 2.1

gppezzi:


jenkins-builds/common-SLES12-2016.10
```diff
@@ -17,9 +17,8 @@
  mc-4.8.18.eb
  nano-2.7.0.eb
  numactl-2.0.11.eb
- PyRegression-2.0.1.eb
  PyRegression-dev.eb
- PyRegression-2.0.2.eb
```
vkarak: I would keep previous releases just in case sth unexpected comes up and we need to use the older version.

The dev version though could be removed.

gppezzi: As discussed, the old module won't be immediately removed from production with this change. 
This is just to keep the list of supported versions updated for the moment when we will proceed with the cleaning.

gppezzi: The module file for the dev version is now deleted.

PR #126, opened 2017-03-20
Para view leone

jfavre:
a long overdue update of ParaView on Leone to support ticket #26917


gppezzi:
I'm merging since the CI does not test builds on Leone. @jfavre will add a reference to the build log.

jfavre:
/scratch/leone/jfavre/leone/software/ParaView/5.3.0-foss-2016.04/easybuild/

PR #131, opened 2017-03-22
TF 1.0 CPU + haswell optimizations

gppezzi:


easybuild/easyconfigs/t/TensorFlow/TensorFlow-1.0.0-CrayGNU-2016.11-Python-3.5.2.eb
```diff
@@ -0,0 +1,63 @@
...
+
+toolchain = {'name': 'CrayGNU', 'version': '2016.11'}
+
+builddependencies = [
+    ('wheel', '0.30.0a0', '-Python-%(pyver)s'),
+    ('Bazel', '0.4.5', '-Python-%(pyver)s'),
+]
+
+dependencies = [
+    ('Python','3.5.2'),
```
victorusu: I would use `pyver` here.

PR #135, opened 2017-03-27
unpacking VMD with EB

gppezzi:
Actually install package with EB instead of only creating the module file.


jfavre:
Thanks @gppezzi  for suggesting this new method

PR #138, opened 2017-03-28
Adding xalt to production build

victorusu:
This adds xalt to production build. It will be called only by Jenkins and on daint.

Since we do not have any mechanism to capture if there was a new build we __always__ perform the reverseMap operation.
But we do it in a different folder and then, copy it on top of the production one.


jgphpc:
New permission for jenkins: https://webrt.cscs.ch/Ticket/Display.html?id=27062

victorusu:
Changed to WIP because we may need to change the `xalt` folder.

jgphpc:
Only for SCS modulefiles.

PR #141, opened 2017-03-30
VASP with perftools-cscs/645-cuda and perftools-cscs/645-nogpu

lucamar:
VASP recipes, CUDA and non-GPU as both are available in production on `daint-gpu`

### DIFF
--- a/easybuild/easyconfigs/v/VASP/VASP-5.4.1-CrayIntel-2016.11-cuda-8.0.54-pat-645-cuda.eb
+++ b/easybuild/easyconfigs/v/VASP/VASP-5.4.1-CrayIntel-2016.11-cuda-8.0.54-pat-645-cuda.eb
@@ -1,0 +1,47 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'MakeCp'
+
+name = 'VASP'
+version = "5.4.1"
+cudaversion = '8.0.54'
+patversion = "645-cuda"
+versionsuffix = '-cuda-%s-pat-%s' %(cudaversion, patversion)
+
+homepage = 'http://www.vasp.at'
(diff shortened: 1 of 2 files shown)


jgphpc:
> If you use a scientific application supported by CSCS, we will provide the EasyBuild configuration file to 
> create the modulefile with the instrumented executable: 
> please contact us if the EasyBuild recipe of a supported applications is not yet available.

Why not installing precompiled supported apps ? It will save time for users who don't want to recompile. The other users can always use the recipe if they want to.

gppezzi:
I don't want these modules to be visible by default. 
We can install all apps on a sandbox and tell users to do module use, how about that?

lucamar:
@gppezzi Can we set a common folder to build the instrumented modulefiles? 
Having a single place would make easier to verify that all the required modules have been provided before each production project submission deadline.

lucamar:
As discussed during the SCS meeting on Tue Apr 4th, I have created the folder `/apps/daint/UES/6.0.UP02/craypat/easybuild` to build the modulefiles with the instrumented executables of supported applications, as I did with CP2K (recipe already in the production repository):
```bash
export EB_CUSTOM_REPOSITORY=/users/lucamar/production/easybuild
export EASYBUILD_PREFIX=$APPS/UES/6.0.UP02/craypat/easybuild
module load daint-gpu && module load EasyBuild-custom
eb CP2K-4.1-CrayGNU-2016.11-cuda-8.0.54-pat-645-cuda.eb
```
Please feel free to use `CP2K-4.1-CrayGNU-2016.11-cuda-8.0.54-pat-645-cuda.eb` as a template for your recipe: the version of perftools-cscs used as build dependency is inserted in the module name.

PR #149, opened 2017-04-07
cpmd with mkl

ajocksch:
linked with mkl; also craypat included in the 2nd recipe

### DIFF
--- a/easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-2016.10-intel17.eb
+++ b/easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-2016.10-intel17.eb
@@ -3,7 +3,6 @@ easyblock = 'MakeCp'
 
 name = 'CPMD'
 version = '4.1'
-versionsuffix = ''
 
 homepage = 'http://www.cpmd.org/'
 description = """CPMD The CPMD code is a parallelized plane wave / pseudopotential
@@ -21,8 +20,7 @@ dependencies = [('fftw/3.3.4.10', EXTERNAL_MODULE)]
 
 skipsteps = ['configure']
 
-common_prebuildopts = ''
-common_prebuildopts += './configure.sh PREFIX=%(installdir)s CRAY-XE6-INTEL-MKL &&'
+common_prebuildopts = './configure.sh PREFIX=%(installdir)s CRAY-XE6-INTEL-MKL &&'
 
 prebuildopts = [
                          common_prebuildopts,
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-2016.10-intel17.eb
```diff
@@ -3,6 +3,7 @@ easyblock = 'MakeCp'
 
 name = 'CPMD'
 version = '4.1'
+versionsuffix = ''
```
gppezzi: The MKL version should replace the libsci version or they are supposed to co-exist?
If it will replace you can remove this empty versionsuffix variable, if they will co-exist I suggest you add something like `-MKL` to the version suffix

easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-2016.10-intel17.eb
```diff
@@ -13,10 +14,15 @@ toolchainopts = {'usempi': True}
...
+#patches = ['patch_cpmd_intel17_mkl.patch']
+patches = [('CRAY-XE6-INTEL-MKL', 'CPMD/configure')]
+
+dependencies = [('fftw/3.3.4.10', EXTERNAL_MODULE)]
+
 skipsteps = ['configure']
 
-common_prebuildopts = 'sed -i \'s#/usr/bin/cpp -P -C -traditional#cpp -P -traditional#\' ./configure/CRAY-XE6-INTEL && '
-common_prebuildopts += './configure.sh PREFIX=%(installdir)s CRAY-XE6-INTEL &&'
+common_prebuildopts = ''
```
gppezzi: You can remove this line and replace `+=` by `=` on the line below.

easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-2016.10-intel17_craypat.eb
```diff
@@ -0,0 +1,44 @@
...
+sources = ['cpmd-v%s.tar.gz' %version]
+
+#patches = ['patch_cpmd_intel17_mkl.patch']
+patches = [('CRAY-XE6-INTEL-MKL', 'CPMD/configure')]
+
+dependencies = [('fftw/3.3.4.10', EXTERNAL_MODULE)]
+
+skipsteps = ['configure']
+
+common_prebuildopts = ''
```
gppezzi: remove line

easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-2016.10-intel17.eb
ajocksch: it should replace the libsci version

PR #153, opened 2017-04-07
Xalt reverseMapD execution

victorusu:


jenkins-builds/production_builds.sh
```diff
@@ -92,9 +92,15 @@ EOF
 done
 
 # update xalt table of modulefiles
+echo "loading PrgEnv-cray"
+module load PrgEnv-cray/6.0.3
```
jgphpc: remove 6.0.3 ?

PR #154, opened 2017-04-07
Updating CPMD recipes

lucamar:
Updated CPMD recipes (with and without perftools) using CrayIntel-2016.11: @ajocksch tested both executables, thank you!


lucamar:
After the pull request will be merged into the master, I will update the production build lists on Piz Daint

PR #155, opened 2017-04-12
Namd perftools easyconfig renaming

ChristopherBignamini:
@lucamar , could you please have a look? Thank you.


easybuild/easyconfigs/n/NAMD/NAMD-2.11-CrayIntel-2016.11-cuda-8.0.54-pat-645.eb
```diff
@@ -2,7 +2,7 @@
 easyblock = 'MakeCp'
 
 name = 'NAMD'
-patversion = "645"
+patversion = "645-cuda"
 version = '2.11'
 versionsuffix = '-cuda-8.0.54-pat-%s' %(patversion)
```
lucamar: I would also define `cudaversion=8.0.54`

lucamar:
Thanks @ChristopherBignamini

PR #159, opened 2017-04-18
CNTK & dependencies

gppezzi:


easybuild/easyconfigs/o/OpenCV/OpenCV-3.1.0-CrayGNU-2016.11-Python-3.5.2.eb
```diff
@@ -0,0 +1,69 @@
...
+ the use of machine perception in the commercial products."""
+
+toolchain = {'name': 'CrayGNU', 'version': '2016.11'}
+toolchainopts = {'pic': True, 'verbose': False}
+
+source_urls = ['https://github.com/opencv/opencv/archive/']
+sources = ['%(version)s.zip']
+
+builddependencies = [
+    ('CMake', '3.6.0', '', True),
```
jgphpc: Why not using /apps/common/UES/jenkins/SLES12/easybuild/modules/all/CMake/3.6.2 ?

gppezzi: CMake is a hidden dependency now, so it won't use that version from apps/common..

For the next software stack update I think we should remove cmake from hide deps and try to consistently use the same (dummy) version.

PR #161, opened 2017-04-19
adding ptl645-loops

jgphpc:
for openacc course.


jgphpc:
https://madra.cscs.ch/ues/userportal/merge_requests/137

PR #163, opened 2017-04-21
mympingpong

jgphpc:
Fixes https://madra.cscs.ch/eth-cscs/regression/issues/66

### DIFF
--- a/easybuild/easyconfigs/h/h5py/h5py-2.5.0-CrayGNU-2016.11-Python-2.7.12-parallel18.eb
+++ b/easybuild/easyconfigs/h/h5py/h5py-2.5.0-CrayGNU-2016.11-Python-2.7.12-parallel18.eb
@@ -1,0 +1,37 @@
+easyblock = "PythonPackage"
+
+name = 'h5py'
+version = '2.5.0'
+python = 'Python'
+pyver = '2.7.12'
+pyshortver = '.'.join(pyver.split('.')[0:2])
+hdf5suff = '-parallel18'
+versionsuffix = '-%s-%s%s' % (python, pyver, hdf5suff)
+
+homepage = 'http://www.h5py.org/'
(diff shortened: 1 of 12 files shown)


easybuild/easyconfigs/h/h5py/h5py-2.5.0-CrayGNU-2016.11-Python-2.7.12-parallel18.eb
```diff
@@ -0,0 +1,37 @@
...
+
+source_urls = [PYPI_SOURCE]
+sources = [SOURCE_TAR_GZ]
+dependencies = [
+    (python, pyver),
+    ('cray-hdf5-parallel/1.8.16', EXTERNAL_MODULE),
+    ('mpi4py', '1.3.1', '-Python-%(pyver)s' + '-timed-pingpong'),
+]
+
+prebuildopts = ' python setup.py configure --mpi --hdf5-version=1.8.16 --hdf5=/opt/cray/pe/hdf5-parallel/1.8.16/gnu/51/ && '
```
gppezzi: use $EBROOTHDF5 instead of hardcoded path?

jgphpc: /opt is from Cray, hence $EBROOTHDF5 is not set.

gppezzi: It should.

If it is not set, it should be because this version is missing here: https://github.com/eth-cscs/production/blob/master/easybuild/cray_external_modules_metadata.cfg

jgphpc:
Solved in https://github.com/eth-cscs/production/pull/297

PR #165, opened 2017-04-24
cleaning up bazel workspace after successful build

gppezzi:
Addressing ticket #27402

TensorFlow builds keep bazel workspace after successful builds and this change should fix this problem.

Failed builds will stay anyway on the temporary folder (as expected).


gppezzi:
This pull request was built on daint101 and the bazelout workspace was deleted: 
```
jenscscs@daint101:/dev/shm/jenscscs> ls /dev/shm/jenscscs/bazelout
ls: cannot access /dev/shm/jenscscs/bazelout: No such file or directory
```

PR #172, opened 2017-05-03
Adding recipe for CMake-3.8.1

lucamar:
... and changing list common-SLES12 accordingly

### DIFF
--- a/easybuild/easyconfigs/c/CMake/CMake-3.8.1.eb
+++ b/easybuild/easyconfigs/c/CMake/CMake-3.8.1.eb
@@ -1,0 +1,24 @@
+# Contributed by Jean-Guillaume Piccinali and Luca Marsella (CSCS)
+easyblock = 'ConfigureMake'
+
+name = 'CMake'
+version = "3.8.1"
+
+homepage = 'http://www.cmake.org'
+description = """CMake, the cross-platform, open-source build system.
+ CMake is a family of tools designed to build, test and package software."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/c/CMake/CMake-3.8.1.eb
```diff
@@ -0,0 +1,24 @@
...
+easyblock = 'ConfigureMake'
+
+name = 'CMake'
+version = "3.8.1"
+
+homepage = 'http://www.cmake.org'
+description = """CMake, the cross-platform, open-source build system.
+ CMake is a family of tools designed to build, test and package software."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
gppezzi: I'd suggest to use: 
```
```
since it could fail to load dependencies if the version 'dummy' is used.

easybuild/easyconfigs/c/CMake/CMake-3.8.1.eb
```diff
@@ -0,0 +1,24 @@
...
+#toolchainopts = {'dynamic': True}
```
gppezzi: please delete comment (dynamic is true by default).

lucamar: Comment removed

easybuild/easyconfigs/c/CMake/CMake-3.8.1.eb
lucamar: Thanks @gppezzi , done!

PR #173, opened 2017-05-03
Reframe packages

vkarak:


vkarak:
I can't understand what to check exactly for the `-dev` package, but I updated accordingly.

easybuild/easyconfigs/r/reframe/reframe-2.2.eb
```diff
@@ -0,0 +1,26 @@
...
+description = "A regression framework for writing portable tests for HPC systems."
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = [ ('archive.tar.gz?ref=v2.2', "tar xfz %s" )]
+source_urls = ['https://madra.cscs.ch/scs/reframe/repository/']
+
+dependencies = [('Python-bare', '3.5.2')]
+
+modextrapaths = {'PATH': ['bin']}
```
gppezzi: This line can probably be eliminated (bin is added by default to the path).

gppezzi:
As discussed, the '-dev' is not used anymore, you can please remove it from the PR.

gppezzi:
If you want do deploy 2.2 on production, you should also update the list here:
https://github.com/eth-cscs/production/blob/master/jenkins-builds/common-SLES12-2016.10

And also remove the old one from the list.

vkarak:
Ok, I will add it.

PR #175, opened 2017-05-04
Recipe for Yambo 4.1.3 with CrayIntel 2016.11

lucamar:
Yambo is a FORTRAN/C code for Many-Body calculations in solid state and molecular physics: 
I had to insert `-qopenmp` in FCFLAGS as the `toolchainopts` flag `openmp: True` was not adding that there and the link would fail otherwise.


easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
```diff
@@ -0,0 +1,46 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'MakeCp'
```
victorusu: Why use `MakeCp` if you are setting `with_configure = 'True'`?
Why not `ConfigureMake`?
What do you think?

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
```diff
@@ -0,0 +1,46 @@
...
+    ('fftw/3.3.4.10', EXTERNAL_MODULE),
+    ('libxc', '3.0.0')
+]
+
+with_configure = 'True'
+
+preconfigopts = ' sed -i "s/\*ftn\* |//" configure && '
+
+configopts = ' CC=cc FC=ftn MPICC=cc CPPFLAGS="" FCFLAGS="-nofor_main -qopenmp" --enable-iotk --enable-open-mp '
+configopts += ' --with-hdf5-includedir="$HDF5_DIR/include" --with-hdf5-libdir="$HDF5_DIR/lib" '
```
victorusu: `$HDF5_DIR` is defined as the mapping of the external hdf5 module from Cray.
Why not use `$EBROOTHDF5`, since it is also defined when we do the map?
This would be a bit more generic.
What do you think?

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
```diff
@@ -0,0 +1,46 @@
...
+configopts += ' --with-netcdf-includedir="$NETCDF_DIR/include" --with-netcdf-libdir="$NETCDF_DIR/lib" '
```
victorusu: `$NETCDF_DIR` is defined as the mapping of the external netcdf module from Cray.
Why not use `$EBROOTNETCDF`, since it is also defined when we do the map?
This would be a bit more generic and we use this one for Amber.
What do you think?

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
```diff
@@ -0,0 +1,46 @@
...
+configopts += ' --with-netcdff-includedir="$NETCDF_DIR/include" --with-netcdff-libdir="$NETCDF_DIR/lib" '
+configopts += ' --with-fft-includedir="$FFTW_INC" --with-fft-libdir="$FFTW_DIR" '
```
victorusu: `$FFTW_DIR` is defined as the mapping of the external fftw module from Cray.
Why not use `$EBROOTFFTW`, since it is also defined when we do the map?
This would be a bit more generic.
What do you think?

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
lucamar: The application does not provide `make install`, so we need to copy the executables anyway like you do with `MakeCp`, which provides the useful variable `files_to_copy`.

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
lucamar: Thanks @victorusu, I have used successfully `$EBROOTHDF5` in the recipe!

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
lucamar: Thanks @victorusu, I have used successfully `$EBROOTNETCDF` in the recipe!

easybuild/easyconfigs/y/Yambo/Yambo-4.1.3-CrayIntel-2016.11.eb
lucamar: Thanks @victorusu, I have used successfully `$EBROOTFFTW` in the recipe!

lucamar:
Thanks @victorusu, I have implemented the corrections that you have suggested with respect to the `$EBROOT` variables of the `EXTERNAL_MODULEFILE`, including `$EBROOTLIBSCI`: the build has worked in my local folder!

PR #176, opened 2017-05-05
Adding Yambo 4.3.1. with perftools-cscs/645-nogpu

lucamar:
It was necessary to unload the module `perftools-cscs` during the configure step:
`preconfigopts += ' module unload perftools-cscs/%s && ' % patversion`
I have also disabled the build of `iotk` (I/O toolkit) since the Makefile would build an internal version running configure again during the build step and thus fail.


victorusu:
`lgtm`

PR #183, opened 2017-05-22
Making Bison hidden by default

victorusu:
If you start the build from scratch (no preloaded module) the Bison module is not hidden, but if a comma is added in the beginning of the hide-deps flag, then it becomes hidden.
This bug has a minimal effect... Since Bison is already installed.


boegel:
This is strange... What does `eb --show-config` print?

easybuild/easybuild.d/system_wide.cfg
```diff
@@ -1,4 +1,4 @@
...
+hide-deps=",Bison,Doxygen,JasPer,NASM,SQLite,Szip,Tcl,bzip2,cURL,flex,freetype,libjpeg-turbo,libpng,libreadline,libtool,libxml2,ncurses,zlib,M4,Serf,APR,APR-util,expat,SCons,binutils,Coreutils,GLib,Qt,SCOTCH,Tk,hwloc,libffi,libunwind,make,numactl,pkg-config,gettext,Autotools,Automake,Autoconf,OPARI2,OTF2,UDUNITS,ZeroMQ,OpenPGM,util-linux,libsodium,libQGLViewer,Eigen,GTS,GL2PS,PyGTS,PyQt,IPython,Python-Xlib,LOKI,SIP,NASM,PIL,libjpeg-turbo,libxcb,libX11,libXau,xproto,kbproto,inputproto,libpthread-stubs,xextproto,libXdmcp,xcb-proto,xtrans,LibTIFF,byacc,guile,libunistring,CMake,PCRE,XZ,PROJ,libutempter,libevent,libiberty,SIONlib,PDT,Cube,Python-bare,fontconfig,arpack-ng,gperf,libgd,libcerf,gnuplot,GraphicsMagick,qrupdate,Mesa,LLVM,nettle,libGLU,libdrm,Mako,GLPK,FLTK,gc,SWIG,wheel,protobuf,protobuf-core,Bazel,cuDNN,nose-parameterized,mock,funcsigs,"
```
boegel: don't use quotes here, they'll be part of the first value, i.e. `"Bison` vs `Bison`

gppezzi: Thanks for spotting that @boegel 

We didn't have quotes on our old [setup](https://github.com/eth-cscs/production/blob/master/easybuild/scripts/setup.sh-stable), but at some point we added quotes a my memory is not helping me to remember why&when..

PR #184, opened 2017-06-01
finally tested 5.3 within a VNC desktop. By-passing OpenGL checks to …

jfavre:
…avoid crash

log files are in:

/scratch/leone/jfavre/leone/software/ParaView/5.3.0-foss-2016.04/easybuild/easybuild-ParaView-5.3.0-20170601.*


gppezzi:
Thanks @jfavre

PR #186, opened 2017-06-02
OpenFOAM 4.1 for leone

gppezzi:
@ajocksch can you please test this?
```
module use /apps/leone/UES/RHEL6.7EUS/sandbox/modules/all
```

### DIFF
--- a/easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-4.1-foss-2016b.eb
+++ b/easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-4.1-foss-2016b.eb
@@ -1,0 +1,36 @@
+name = 'OpenFOAM'
+version = '4.1'
+
+homepage = 'http://www.openfoam.com/'
+description = """OpenFOAM is a free, open source CFD software package.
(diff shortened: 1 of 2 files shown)


ajocksch:
Checks: dambreak, hotroom, tjunction and motorbike passed successfully

minor: error message since paraview is not integrated into the installation

jfavre:
Anton Kidess of HILTI has replied to me the following:

_Long story short, the OpenFOAM install does not require its own build of Paraview with the CFDdirect plugin. Instead, the native reader with a system-wide Paraview build makes sense. The native reader is always included in the binaries I download from Kitware, so I assume you don’t need to do anything special when compiling Paraview to enable it._

My translation of this is: We should not compile install OpenFOAM with an integrated ParaView. I.e. Compile OpenFOAM only. And Anton Kidess and HILTI folks rely on the **native** reader interface of ParaView. Indeed, I have verified that my current Release Candidate installation of ParaView on Leone offers an OpenFoam reader option.

Jean

gppezzi:
Thanks @jfavre for the heads up and @ajocksch for the tests. 

If you don't have any remark to add, I'll merge this recipe without ParaView.

PR #188, opened 2017-06-07
version 5.4 for Leone

jfavre:
I have tested my private EB build on leone, both in client-server mode and inside a VNC desktop

log files are in /user/jfavre/easybuild/leone/software/ParaView/5.4.0-foss-2016.04/easybuild

5.4 Final Release is ready to be deployed on Leone


jfavre:
have recompiled with 2016b and checked client-server execution without any issues.
the 2016.04 version can be abandoned and the 2016b can be used instead.

gppezzi:
thanks @jfavre 

I've also added our custom CSCS python version to your PR (instead of using the stock version from easybuild)

PR #190, opened 2017-06-08
Reframe 2.3

vkarak:
This should not be made the default version, since it has broken some of our regression tests.


gppezzi:
Do you want to install it anyway? If so, please add it to the jenkins build list.

PR #191, opened 2017-06-09
Updating set eb_config_dir in recipe

lucamar:
The variables is now defined as follows:
```
set eb_config_dir          /apps/daint/UES/jenkins/production/easybuild
```


gppezzi:
please also update the path here
https://github.com/eth-cscs/production/blob/master/easybuild/module/Easybuild

lucamar:
Updated https://github.com/eth-cscs/production/blob/master/easybuild/module/Easybuild as well and added to this pull request

gppezzi:
looks good, let's test!

PR #193, opened 2017-06-13
[daint-gpu] remove confict with dwstat

jgphpc:


easybuild/easyconfigs/d/ddt/ddt-7.0.2.eb
```diff
@@ -36,28 +36,20 @@ set     url                        "http://www.allinea.com"
...
-}
-
-if  { $hostname=="tave" } {
+} elseif { $hostname=="tave" } {
           set     os      "SLES12"
           set     version "7.0.2"
-}
-  
-if { "$os"=="" } {
-        puts stderr "L60: host=$host"
```
jgphpc: This comment was making dwstat/eswrap to fail.

lucamar:
An important remark for the Jenkins project to test the pull request: daint-gpu and daint-mc are mutually exclusive, therefore you can use only one per time in the test. 
Currently, if `gpu` is matched in the title of the pull request, then the module `daint-gpu` is loaded and your EasyBuild configuration file will be built only with that one. You need to open a separate request for `daint-mc`.

lucamar:
Multiple labels will work with different systems names instead, as Jenkins will log on each system and check the matching keywords on the title of the pull request. E.g.:
`[leone] [daint-mc] ...` will test the recipe on Leone and Piz Daint (Multicore)

PR #198, opened 2017-06-14
Updating EasyBuild-custom-cscs

lucamar:
Using the variable `CRAYPE_NETWORK_TARGET` to define `EASYBUILD_PREFIX`


lucamar:
The checks fail because the module `EasyBuild-custom` is already loaded, so its build can't be tested.

PR #200, opened 2017-06-16
[monch] sync dep versions for zlib, ncurses, libreadline

gppezzi:
cc @minduni


gppezzi:
Merging as agreed with @minduni (this should work only on monch04, so it is ok if CI fails).

PR #201, opened 2017-06-19
[kesch] Update R 3.3.2 for Kesch/Escha

gppezzi:


gppezzi:
test this please

gppezzi:
@lucamar ncurses and other libs will be updated when we provide the new software stack, for now I'd like to avoid having different versions that could potentially cause conflict with the existing modules.

PR #213, opened 2017-06-23
Adding Dom to production script

lucamar:
I have inserted the system Dom. Do we need to update xalt table of modulefiles on Dom as well?


victorusu:
`lgtm`

lucamar:
Thanks @victorusu ... so we don't need to update xalt table of modulefiles on Dom as well?

PR #217, opened 2017-06-27
[daint-gpu] new Intel ray-tracing libs for use in ParaView. They fix the blank (n…

jfavre:
Testing OSPRAY-rendering of particles in ParaView was broken with the previous Intel versions. These two updates fix the problem. ParaView 5.4 will have to be re-linked as well.


gppezzi:
retest this please

PR #219, opened 2017-07-07
[kesch] Subversion 1.9.6

gppezzi:
Subversion 1.7 provided by rpm should be enough for the users, I'm just adding this in case it is requested.

This should work on the node with RH7.3 and it should replace the dummy version (that does not work on the new OS).


gppezzi:
I'm merging anyway since the PR is tested only on the old OS. Thanks @lucamar for the review.

PR #224, opened 2017-07-10
Change from $CRAYPE_NETWORK_TARGET to $PE_ENV

victorusu:
This PR is about changing the variable we use to separate from CRAY system from non-Cray ones.


gppezzi:
I'm closing this, let's fix this on the same PR as https://github.com/eth-cscs/production/pull/225

PR #225, opened 2017-07-10
Change from CRAYPE_NETWORK_TARGET to system name check

victorusu:
This PR is about how we define if the system is a CRAY or not


victorusu:
retest this please

gppezzi:
Don't expect this test to pass (EasyBuild-custom is used by jenkins, so we have chicken&egg problem).

lucamar:
We need to find another variable, see my comment in #224

lucamar:
The title of the pull request has not matched `leone` and `monch`, due to an upper case letter in `Leone` and a typo in `month`.

gppezzi:
@victorusu @lucamar Since the new CN from kesch breaks again the system auto-guessing, let's please stop doing that and use host names for identifiying kesch/dom/daint.

lucamar:
I have added the check of the system name, taking into account `*esch*` compute nodes as well. 
The CI on Jenkis failed as expected, since I have removed system names from the pull request title.

gppezzi:
I'm feeling lucky, so I'm merging this.

PR #229, opened 2017-07-12
[dom] adds support for Cray PE 17.06 and sample .ebs for testing

gppezzi:
I'm proposing to also adopt the versioning scheme as defined by [Cray](https://ditaweb.com/cray.deliver/browse/xc/article/released-cray-xc-programming-environments-1706---june-29-2017): 
* `17.06` instead of `2017.06`


lucamar:
As discussed, we can temporarily use a module footer with the following `setenv` instruction:
```
setenv  LD_LIBRARY_PATH         "$::env(CRAY_LD_LIBRARY_PATH):$::env(LD_LIBRARY_PATH)"
```
The footer could be located in $APPS/UES/login and can be called by EasyBuild using `--modules-footer=...` in the production list. As an alternative, we could add the footer instruction within the production script for the supported applications only. What's your opinion?

gppezzi:
@lucamar If we use the `--modules-footer=` option, this will systematically append `$CRAY_LD_LIBRARY_PATH` to `$LD_LIBRARY_PATH` for every dependency we load? We would end up with a very long variable with duplicate entries?

lucamar:
@gppezzi You're right, the environment variable `LD_LIBRARY_PATH` would contain duplicates every time we load a dependency. In order to avoid that, the simplest way would be to append this footer:
```
remove-path   LD_LIBRARY_PATH         "$::env(CRAY_LD_LIBRARY_PATH)"
prepend-path  LD_LIBRARY_PATH         "$::env(CRAY_LD_LIBRARY_PATH)"
```

lucamar:
retest this please

lucamar:
retest this please

jgphpc:
Would you mind adding more deps as in https://github.com/eth-cscs/production/commit/bf39249fd571289cd7a404c0813a98bccc52da5a ?
This would be closer to what `module load cdt/17.06` is doing.

jgphpc:
Should i put this https://github.com/eth-cscs/production/pull/229#issuecomment-316636625 in my .eb ?

jgphpc:
Could we also move these .eb ... :

* production.git/jenkins-builds/MCH-RH7.3-17.02
```
 Anaconda3-4.4.0.eb
 Autotools-201602.eb
 CMake-3.7.2.eb
 NCL-6.4.0.eb
 rstudio-1.0.143.eb
```

to `common-RH7-17.02` instead ?

gppezzi:
@jgphpc `Would you mind adding more deps as in bf39249 ?`

When discussing with Cray it was recommended that we use the default (latest) version of craype, that's why we didn't add it. Same for PMI.

Now they pin all the versions with this cdt module (as we would have preferred to do initially), but I'm not sure if they update LD_LIBRARY_PATH too (all these modules generally rely on defaults), so we need to double check with Cray.

gppezzi:
@jgphpc as discussed: for the common builds, the initial idea was to have generic builds shared by many systems. 

Currently there's very little re-use of these stacks so we're merging the 'common' builds with the main build list by the next software stack deployments.

jgphpc:
> I'm not sure if they update LD_LIBRARY_PATH too

Looks like they do not:

mll cdt/17.06
=> Switching to cray-mpich/7.6.0.

ftn -g 0.F90  -dynamic
srun -n1 -Cgpu -t1 ./a.out
=> MPI VERSION    : CRAY MPICH version 7.6.0 (ANL base 3.2)
:-(


export LD_LIBRARY_PATH=$CRAY_LD_LIBRARY_PATH:$LD_LIBRARY_PATH
srun -n1 -Cgpu -t1 ./a.out
=> MPI VERSION    : CRAY MPICH version 7.6.0 (ANL base 3.2)
:-)

Conclusion:
export LD_LIBRARY_PATH=$CRAY_LD_LIBRARY_PATH:$LD_LIBRARY_PATH
is mandatory...

gppezzi:
@jgphpc could you please push the commits from your branch `cdt1706-jg` to this one, so that we can review the new toolchains altogether?

lucamar:
I'm closing this pull request, as we are now working on `17.08`: 
https://github.com/eth-cscs/production/pull/262

PR #231, opened 2017-07-13
WIP Lowercase packages

victorusu:
Since there is no way to have an if statement on the file.
We need to provide the package list twice, one uppercase and one lowercase.
This way the hidden packages can be hidden.


gppezzi:
I'm closing this since it is not needed. See [here](https://gist.github.com/gppezzi/9bb33ace4a6adc582dd9c45d31cbb174). 

Thanks anyway @victorusu and enjoy your holidays.

PR #235, opened 2017-07-14
[leone] updating prgenv and hidden modules

gppezzi:
We should rebuild the software stack with the new settings before it gets announced to the users.


gppezzi:
retest this please

PR #238, opened 2017-07-19
[daint dom kesch leone monch] Regression eb

lucamar:
Adding flag `-u/--unuse` to production script `production.sh` to allow unuse of selected PATH in MODULEPATH. The new flag will be used by the script within the jenkins project `RegressionEB`.


lucamar:
The script will be used by `https://jenkins.cscs.ch/view/RegressionEB/job/RegressionEB`

jenkins-builds/production.sh
```diff
@@ -7,18 +7,19 @@ scriptname=$(basename $0)
...
-    -f,--force    Force build of given package  (optional: double quotes for a list)
+    -a,--arch     Architecture (gpu or mc)           (mandatory: Dom and Piz Daint only)
+    -f,--force    Force build of given package       (optional: double quotes for a list)
     -h,--help     Help message
-    -l,--list     Production list file          (mandatory: EasyBuild production list)
-    -p,--prefix   EasyBuild prefix folder       (mandatory: installation folder)
-    -x,--xalt.    [yes|no] update XALT database (optional, default is yes)
+    -l,--list     Production list file               (mandatory: EasyBuild production list)
+    -p,--prefix   EasyBuild prefix folder            (mandatory: installation folder)
+    -u,--unuse    Module unuse colon separated PATH  (optional: default is null)
```
gppezzi: Is there a case where -u and -x will not be used together? 

If not, then I'd prefer to have only one option to that replaces both options (such as `-r,--regression`).

But I'm ok with this solution as well.

lucamar: I have implemented this flag to treat every system in the same way within the Jenkins project `RegressionEB`: this approach has the advantage to select explicitly the path to be unused, avoiding the need to have it hard coded for Piz Daint only within the script `production.sh`.
The flags `-u $PATH -x no` are used together for Dom and Piz Daint only.

PR #242, opened 2017-07-25
[daint dom] Module purge in production on Piz Daint/Dom

lucamar:
I have updated the script `production.sh` to use absolute paths, in order to let us call it from arbitrary locations. 

I have also inserted a `module purge` instruction to clean the environment on Dom and Piz Daint before starting the build, loading the necessary modulefiles in place of the previous `module use` command:
`module load craype craype-network-aries modules`

Finally, I set `EB_CUSTOM_REPOSITORY=/apps/common/UES/jenkins/production/easybuild`, since the latter is always up to date for production builds. We can add later a flag to define this variable as a command line argument or check if it is already defined in the environment, in case we would like to use the script for testing new recipes: in fact, in that case one cannot use the master branch of the production repository, but needs to sync a local branch.

### DIFF
--- a/jenkins-builds/production.sh
+++ b/jenkins-builds/production.sh
@@ -72,21 +72,11 @@ done
 # optional EasyBuild arguments
 eb_args=""
 
-# check prefix folder
-if [ -z "$PREFIX" ]; then
-    echo -e "\n Prefix folder not defined. Please use the option -p,--prefix to define the prefix folder \n"
-    usage
-fi
-
 # system name (excluding node number)
 if [[ "$HOSTNAME" =~ esch ]]; then
  system=${HOSTNAME%%[cl]n-[0-9]*}
 else
  system=${HOSTNAME%%[0-9]*}
-fi
-
-if [ -z "$update_xalt_table" ]; then
-    update_xalt_table=yes
 fi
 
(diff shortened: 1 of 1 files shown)


jenkins-builds/production.sh
```diff
@@ -96,17 +96,15 @@ if [[ "$system" =~ "daint" || "$system" =~ "dom" ]]; then
...
+        module purge
+        module load craype craype-network-aries modules
         module load daint-${ARCH}
         eb_args="${eb_args} --modules-header=${scriptdir%/*}/login/daint-${ARCH}.h"
     fi
 fi
 
 # --- COMMON SETUP ---
-export EB_CUSTOM_REPOSITORY=$PWD/easybuild
+export EB_CUSTOM_REPOSITORY=/apps/common/UES/jenkins/production/easybuild
```
victorusu: Why not wrap the variable definition with a `if` statement. So that if it is defined, one can use the defined one, if not defined, one could define this default?
Some think like:
```bash
if [ -z $EB_CUSTOM_REPOSITORY ]; then
    export EB_CUSTOM_REPOSITORY=/apps/common/UES/jenkins/production/easybuild
fi
```

lucamar: Sure, I have already mentioned this possibility in the description above.

PR #244, opened 2017-07-26
[daint dom gpu] Updating NCO with ANTLR support for ncap2 executable

lucamar:
I have inserted the dependency ANTLR to allow the build of the `ncap2` executable, which is otherwise disabled: I have also added the library ANTLR to the `hide-deps` list, as it is not used otherwise.


victorusu:
`latm`

PR #247, opened 2017-07-28
[daint-gpu] Remove vtk

lucamar:
Removing VTK from `daint-gpu` production list.


gppezzi:
As agreed with @jfavre, we'll remove this (deprecated version) from the production list in order to avoid failures on dom (CNs) and we'll provide a more recent build of VTK that does not depend on X.

PR #249, opened 2017-07-29
[daint dom kesch leone monch] ddt/705

jgphpc:


jgphpc:
What should be the title for building/installing in /apps/common/UES/RHAT6/ and /apps/common/UES/SLES12/ ?

jgphpc:
retest this please

jgphpc:
merge this please.

easybuild/easyconfigs/d/ddt/ddt-7.0.5.eb
```diff
@@ -10,64 +10,74 @@ toolchain = {'name': 'dummy', 'version': 'dummy'}
...
+        set osv "6"
+    }
 }
-  
-if { "$os"=="" } {
-        puts stderr "L60: host=$host"
-        set     os      "SLES12"
-        set     version "7.0"
+
+if { ${OS_ID}=="generic" } {
```
victorusu: What do you think about change the concept  from `if { ${OS_ID}=="generic" } {` to actually checking if it is redhat 6?

Something like 

```
set OS_ID "NULL"
if { [ file exist /etc/os-release ] } {
...
} elseif  { [ file exist /etc/redhat-release ] } {
...
} 
...
if { ${OS_ID}=="NULL" } {
error "OS not supported"
}
```

easybuild/easyconfigs/d/ddt/ddt-7.0.5.eb
```diff
@@ -10,64 +10,74 @@ toolchain = {'name': 'dummy', 'version': 'dummy'}
...
+set     url2         "http://user.cscs.ch"
+set     modulename   "ddt"
+set     version      "7.0.5"
+set     packageroot  "/apps/common/UES/${ost}${osv}/${modulename}/${version}"
 
 # ============================================================================
-# module-info mode ==> load, remove, display, help, whatis, switch, switch1, switch2, or switch3.
-# load
-# ============================================================================
 #if { [ module-info mode load ] } {
```
victorusu: You do not need the `if { [ module-info mode load ] } {`. It is even commented out.
Why not delete it?

easybuild/easyconfigs/d/ddt/ddt-7.0.5.eb
jgphpc: we don't build ebs on rhat6 anymore.

victorusu: 👍

easybuild/easyconfigs/d/ddt/ddt-7.0.5.eb
jgphpc: Was needed with older versions of the modulecmd, keeping it for debug.

victorusu: 👍

PR #252, opened 2017-08-04
[leone] adding boost for ESyS

gppezzi:
So that the user finds all the dependencies on the modulepath when building ESyS


jgphpc:
Request is from https://webrt.cscs.ch/Ticket/Display.html?id=28458

PR #253, opened 2017-08-04
[daint dom kesch leone monch] Define BUILDPATH and TMPDIR only if not set

lucamar:
The environment variables `EASYBUILD_BUILDPATH` and `EASYBUILD_TMPDIR` will be defined only if the `$USER` has not previously set these varialbes in the environment.

### DIFF
--- a/easybuild/easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb
+++ b/easybuild/easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb
@@ -83,11 +83,15 @@ setenv EASYBUILD_INSTALLPATH                    $::env(EASYBUILD_PREFIX)
 #
 # If not set, EASYBUILD_BUILDPATH and EASYBUILD_TMPDIR will point to /dev/shm/$USER/easbuild/stage
 #
-if { ! [ info exists ::env(EASYBUILD_TMPDIR) ] } {
-    setenv EASYBUILD_TMPDIR                         /dev/shm/$::env(USER)/easybuild/stage/tmp
+if { [ info exists ::env(EASYBUILD_TMPDIR) ] } {
+    setenv EASYBUILD_TMPDIR                     $::env(EASYBUILD_TMPDIR)
+} else {
+    setenv EASYBUILD_TMPDIR                     /dev/shm/$::env(USER)/easybuild/stage/tmp
 }
-if { ! [ info exists ::env(EASYBUILD_BUILDPATH) ] } {
-    setenv EASYBUILD_BUILDPATH                      /dev/shm/$::env(USER)/easybuild/stage/build
+if { [ info exists ::env(EASYBUILD_BUILDPATH) ] } {
+    setenv EASYBUILD_BUILDPATH                  $::env(EASYBUILD_BUILDPATH)
+} else {
+    setenv EASYBUILD_BUILDPATH                  /dev/shm/$::env(USER)/easybuild/stage/build
 }
 ###############
 # FINAL TOUCH #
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb
```diff
@@ -81,12 +81,14 @@ if { ! [ info exists ::env(EASYBUILD_PREFIX) ] } {
...
-# These variables are set to /dev/shm/$USER/easbuild/stage
-# EASYBUILD_BUILDPATH
-# EASYBUILD_TMPDIR
+# If not set, EASYBUILD_BUILDPATH and EASYBUILD_TMPDIR will point to /dev/shm/$USER/easbuild/stage
 #
-setenv EASYBUILD_TMPDIR                         /dev/shm/$::env(USER)/easybuild/stage/tmp
-setenv EASYBUILD_BUILDPATH                      /dev/shm/$::env(USER)/easybuild/stage/build
+if { ! [ info exists ::env(EASYBUILD_TMPDIR) ] } {
+    setenv EASYBUILD_TMPDIR                         /dev/shm/$::env(USER)/easybuild/stage/tmp
+}
```
victorusu: I think you need to add an `else` clause. This is important for the `module unload`. 
Because the `module unload` will rerun the the tcl script and change all the `setenv` to `unset`. But it will never reach the `setenv` without the `else` statement. Because when loaded the `::env(EASYBUILD_TMPDIR)` exists.

Thus, it should look like:
```
if { ! [ info exists ::env(EASYBUILD_TMPDIR) ] } {
    setenv EASYBUILD_TMPDIR                         /dev/shm/$::env(USER)/easybuild/stage/tmp
} else {
    setenv EASYBUILD_TMPDIR                         $::env(EASYBUILD_TMPDIR)
}
```

easybuild/easyconfigs/e/EasyBuild-custom/EasyBuild-custom-cscs.eb
```diff
@@ -81,12 +81,14 @@ if { ! [ info exists ::env(EASYBUILD_PREFIX) ] } {
...
+if { ! [ info exists ::env(EASYBUILD_BUILDPATH) ] } {
+    setenv EASYBUILD_BUILDPATH                      /dev/shm/$::env(USER)/easybuild/stage/build
+}
```
victorusu: The same `else` as before.

easybuild/module/Easybuild
```diff
@@ -87,12 +87,14 @@ if { ! [ info exists ::env(EASYBUILD_PREFIX) ] } {
...
-# These variables are set to /dev/shm/$USER/easbuild/stage
-# EASYBUILD_BUILDPATH
-# EASYBUILD_TMPDIR
+# If not set, EASYBUILD_BUILDPATH and EASYBUILD_TMPDIR will point to /dev/shm/$USER/easbuild/stage
 #
-setenv EASYBUILD_TMPDIR                         /dev/shm/$::env(USER)/easybuild/stage/tmp
-setenv EASYBUILD_BUILDPATH                      /dev/shm/$::env(USER)/easybuild/stage/build
+if { ! [ info exists ::env(EASYBUILD_TMPDIR) ] } {
+    setenv EASYBUILD_TMPDIR                         /dev/shm/$::env(USER)/easybuild/stage/tmp
+}
```
victorusu: Same as before.

easybuild/module/Easybuild
```diff
@@ -87,12 +87,14 @@ if { ! [ info exists ::env(EASYBUILD_PREFIX) ] } {
...
+if { ! [ info exists ::env(EASYBUILD_BUILDPATH) ] } {
+    setenv EASYBUILD_BUILDPATH                      /dev/shm/$::env(USER)/easybuild/stage/build
+}
```
victorusu: Same as before.

lucamar: Done, thanks @victorusu

easybuild/module/Easybuild
lucamar: Done, thanks @victorusu

victorusu:
`lgtm`

PR #254, opened 2017-08-04
[daint dom] Update production with footer

lucamar:
I'm inserting the `--modules-footer` to update the library path on Cray systems.
Meanwhile, I'm also changing the way we force a build to address a bug.


lucamar:
I have fixed the matching of the list of files give with the force flag within multiple production files. 
The script works, but I still get the following messages: 
```
line 0: unalias: sd_log: not found
line 0: unalias: sc_log: not found
```

PR #256, opened 2017-08-09
[leone] Updating ESyS

gppezzi:
- Updated `$PYTHONPATH` as requested by the user
- Adding Python as explicit dependency and using `pyver/pyshortver` variables

### DIFF
--- a/easybuild/easyconfigs/e/ESyS/ESyS-2.3.5-foss-2016b-Python-2.7.12.eb
+++ b/easybuild/easyconfigs/e/ESyS/ESyS-2.3.5-foss-2016b-Python-2.7.12.eb
@@ -1,8 +1,10 @@
+# @author: jgp, gppezzi (CSCS)
+
 easyblock = 'ConfigureMake'
 
 name = 'ESyS'
 version = '2.3.5'
-versionsuffix = '-Python-2.7.12'
+versionsuffix = '-Python-%(pyver)s'
 
 homepage = 'https://launchpad.net/esys-particle'
 description = """ESyS-Particle is Open Source software for particle-based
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/e/ESyS/ESyS-2.3.5-foss-2016b-Python-2.7.12.eb
```diff
@@ -1,8 +1,10 @@
+# @author: jgp, gppezzi (CSCS)
```
jgphpc: why not using `git blame` ?

easybuild/easyconfigs/e/ESyS/ESyS-2.3.5-foss-2016b-Python-2.7.12.eb
```diff
@@ -1,8 +1,10 @@
...
+
 easyblock = 'ConfigureMake'
 
 name = 'ESyS'
 version = '2.3.5'
-versionsuffix = '-Python-2.7.12'
+versionsuffix = '-Python-%(pyver)s'
```
jgphpc: `pyver` is defined by the dependency with ('Python', '2.7.12') right ?

gppezzi: Yes. But I don't know if that is a bug or feature..

gppezzi: (in the sense that Python is already loaded by Boost, but `pyver` is not defined in this case)

jgphpc:
jenkins failed because dom is down but leone passed ok:
https://jenkins.cscs.ch/job/ProductionEBtestingPR/191/console
=> merging.

easybuild/easyconfigs/e/ESyS/ESyS-2.3.5-foss-2016b-Python-2.7.12.eb
gppezzi: we can use `git blame` as a last resort, but it is nicer if people claim authorship. Besides, the latter option will also work if the file is moved to another git repo.

PR #258, opened 2017-08-11
Adding moose with VTK

victorusu:
Since the code takes forever to compile, you can access the logs here:
```
/apps/daint/UES/6.0.UP02/sandbox-vh/easybuild/software/moose/d2193ae-CrayGNU-2016.11-VTK-7.1.0/easybuild/easybuild-moose-d2193ae-20170810.163809.log
```

I am also renaming the VTK file to make it compatible with Easybuild file name pattern. If not, one cannot use VTK as a dependency for another package.
```
VTK-7.1.0-CrayGNU-2016.11.eb → VTK-7.1.0-CrayGNU-2016.11-Python-3.5.2.eb
```


victorusu:
I have not added any system to Jenkins, because we should not re-test this build, since it takes several hours and blocks Jenkins slaves.

victorusu:
It does not work... moose requires Python 2 and this build links to VTK with Python3. I am changing it...

lucamar:
Is this still work in progress?

PR #264, opened 2017-08-17
[daint dom] Fixing footer for Cray systems

lucamar:
Fix footer for modules built on Cray systems


lucamar:
`::env` is the fully qualified built-in global array `env` accessible within a Tcl script: http://wiki.tcl.tk/1624

PR #269, opened 2017-08-23
[daint dom kesch leone monch] Proposal to define sourcepath

lucamar:
It will be useful for Jenkins projects where a different `EASYBUILD_SOURCEPATH` might be needed.


lucamar:
Thanks @victorusu I have fixed the indentation following your review

PR #270, opened 2017-08-23
[daint-gpu] Pygelf WIP

victorusu:
This is a python binding to GELF, which we use to send data to graylog servers.


lucamar:
Do you really need a new module for this package? Can't this package be included in our `Python` modulefile as an extension?

victorusu:
guys, sorry but.. This is the python package that OPS uses and the one they told me to use. I  have already implemented the ReFrame and XALT proposals using this python package. I am not in favor of changing the python programs that I have written so far, just to check whether `graypy` is better or not.

victorusu:
I am not against adding this package in the default python module. But the influx (similar package but for another database) package is also an additional python module...

lucamar:
@victorusu We try to keep `Python` packages within our main `Python` modulefile rather than installing separate modules (see for instance `mpi4py`)

jgphpc:
xalt is not for users hence keep it outside of the main Python please.

lucamar:
@jgphpc Then there's no need to add the package it to our production list

jgphpc:
I don't get your problem. This will be hidden and we will need it to run daily, sounds production to me.

lucamar:
There's no `--hidden` flag in the proposed production file enclosed to this pull request. 
Anyway I think that `gelf` and also `influxdb` could simply be in ` common-SLES12-2016.10` instead, both with `--hidden`.

jenkins-builds/common-SLES12-2016.10
```diff
@@ -17,6 +17,7 @@
  mc-4.8.18.eb
  nano-2.7.0.eb
  numactl-2.0.11.eb
+ pygel-0.3.1-CrayGNU-2016.11-Python-2.7.12.eb
```
lucamar: There's a typo here, `f` is missing: `pygelf-0.3.1-CrayGNU-2016.11-Python-2.7.12.eb`

gppezzi:
+1 for hidding this from the users

However I don't like the idea of adding a recipe on `common` which depends on something that is not generic (python cray gnu). 

I suggest that you either use python bare or move this to the main software stack (if python cray gnu is really needed)

jgphpc:
Can you also add `pygelf-0.3.1-CrayGNU-17.08-cray-python-17.06.1.eb` or do you need another PR ?

victorusu:
I think this will be irrelevant in two weeks. So we should not merge this one. Right?

PR #271, opened 2017-08-24
[daint-gpu] New recipes for VASP 5.4.4

lucamar:
This is the latest release of VASP and some adjustments were necessary to build using Intel MKL libraries. Although the toolchain `2016.11` will be replaced by the end of September by `17.08`, there is still the chance that users might want to build this new release in the next month.


lucamar:
retest this please

PR #274, opened 2017-08-25
[daint-gpu] Reframe 2.6 easyconfig

vkarak:


vkarak:
Any idea why the CI is stuck with that?

lucamar:
I have seen several builds stuck in the queue: it might also be that `keschTDS` disabled has created troubles to the projects, it might be better to remove the node from the build beforehand.

PR #262, opened 2017-08-12
[dom-gpu] WIP craype/17.08

jgphpc:
eb for cle/604

PR #278, opened 2017-08-30
[dom] Toolchains for Cray CDT 17.08

gppezzi:


jgphpc:
I would compare with the toolchains that we already use in  https://github.com/eth-cscs/production/pull/262

gppezzi:
I prefer to breakdown #262 into smaller PRs, so that it can be properly reviewed and pass the CI tests. Let's discuss the details tomorrow.
 
I took these files from your branch in #262, as it should be a basic requirement for all the new software stack (I've just removed the lines which were commented out)

PR #280, opened 2017-08-31
[dom] linux utilities

jgphpc:
After some thinking, i will not add these to the jenkins list of builds but still want to push so i can use them:

```
git/git-2.14.1.eb
htop/htop-2.0.1.eb
hwloc/hwloc-1.11.7.eb
likwid/likwid-4.3.0.eb
nano/nano-2.8.6.eb
ncurses/ncurses-6.0.eb
Ruby/Ruby-2.4.1.eb
tmux/tmux-2.5.eb
```


lucamar:
retest this please

PR #284, opened 2017-08-31
[dom] Update cray_external_modules_metadata.cfg

lucamar:
New Cray modules available:
- `cray-fftw`
- `cray-python`
- `cray-R`

New versions of existing modules:
- `cudatoolkit`
- `fftw`


easybuild/cray_external_modules_metadata.cfg
```diff
@@ -101,6 +106,16 @@ name = PETSc
...
 
+[cray-python/17.06.1]
+name = Python
+version = 17.06.1
+prefix = PYTHONPATH
+
+[cray-R/3.3.3]
+name = R
+version = 3.3.3
+prefix =
```
gppezzi: why not using:
```
prefix = PATH/..
```

PR #293, opened 2017-09-01
WIP [dom-gpu] LAMMPS PE 17.08

kraushm:


kraushm:
Will  be combined with the non-gpu one

PR #294, opened 2017-09-01
[dom-gpu] NAMD and CHARM easyconfig for PE 17.08

ChristopherBignamini:
Could you please have a look? Thanks!


gppezzi:
retest this please

lucamar:
retest this please

lucamar:
I will use this pull request to test the new Jenkins project `TestingEB`, so please don't be worried if it fails

lucamar:
retest this please

lucamar:
`ERROR: Irresolvable dependencies encountered: Charm++/6.7.1-CrayIntel-17.08`

ChristopherBignamini:
Error should be resolved by the last commit

lucamar:
```
Current diff list by 'git diff origin/master..HEAD --name-only --oneline --no-merges --diff-filter=ACMRTUXB':
easybuild/cray_external_modules_metadata.cfg
easybuild/easyconfigs/c/Charm++/Charm++-6.7.1-CrayIntel-17.08.eb
easybuild/easyconfigs/c/Charm++/Charm++-6.8.0-CrayIntel-17.08.eb
easybuild/easyconfigs/n/NAMD/NAMD-2.12-CrayIntel-17.08-cuda-8.0.eb
easybuild/easyconfigs/n/NAMD/NAMD-2.12-CrayIntel-17.08.eb
easybuild/easyconfigs/n/NAMD/NAMD-2.12.patch
easybuild/easyconfigs/v/VMD/VMD-1.9.3-egl.eb
easybuild/easyconfigs/v/VMD/VMD-1.9.3-ogl.eb
easybuild/easyconfigs/v/Visit/Visit-2.12.0-CrayGNU-2016.11.eb
```

lucamar:
Can you please try if NAMD can be built using `Charm++-6.8.0-changa-CrayIntel-17.08-cuda-8.0.eb` and `Charm++-6.8.0-changa-CrayIntel-17.08.eb` that are already available in production in `17.08`?

ChristopherBignamini:
@lucamar I'm using Charm 6.7.1 because, according to the release notes of NAMD 2.12, it seems that this is the foreseen version. I added Charm 6.8.0 because @jgphpc asked it for another module.

jgphpc:
I will remove 6.8 as it is already in master

jgphpc:
:ok: https://jenkins.cscs.ch/job/TestingEB/label=dom/600/console

easybuild/easyconfigs/c/Charm++/Charm++-6.7.1-CrayIntel-17.08.eb
```diff
@@ -0,0 +1,30 @@
+# Contributed by gppezzi, Luca Marsella and Christopher Bignamini (CSCS)
+easyblock = 'CmdCp'
+
+nameletterlower = 'c'
```
lucamar: You don't need to define this variable, as it is already created by EasyBuild: 
```
$ eb --avail-easyconfig-templates | grep nameletterlower
    %(nameletterlower)s: lower case of value of nameletter
    PYPI_LOWER_SOURCE: pypi source url (lowercase name) (https://pypi.python.org/packages/source/%(nameletterlower)s/%(namelower)s)
```

easybuild/easyconfigs/n/NAMD/NAMD-2.12-CrayIntel-17.08-cuda-8.0.eb
```diff
@@ -0,0 +1,41 @@
+# Contributed by gppezzi, Luca Marsella and Christopher Bignamini (CSCS)
+easyblock = 'MakeCp'
+
+nameletterlower = 'n'
```
lucamar: You don't need to define this variable, as it is already created by EasyBuild:
```
$ eb --avail-easyconfig-templates | grep nameletterlower
    %(nameletterlower)s: lower case of value of nameletter
```

lucamar:
I think we should try to build using up to date versions of the dependencies, as far as possible. 

`NAMD` developers package NAMD give their own Charm release, in this case `6.7.1`, but we might give `6.8.0` a try anyway. 

Same holds for `Tcl`, which is not `8.6.7` while `NAMD` packages `8.5.9`.

I know that it takes some time to test these build, but that's the goal of the milestone `Summer update`.

jgphpc:
> give 6.8.0 a try

In that case, please try using the charm++ from master: 
https://github.com/eth-cscs/production/tree/master/easybuild/easyconfigs/c/Charm%2B%2B

lucamar:
I have modified the recipes using the latest modules available in production: I have used `Charm++` without cuda support, as the NAMD packaged CHamr version was not using it, however we can re-assess that if necessary.

I only miss a patch to `arch/CRAY-XC.tcl` in order to remove `-ltcl8.5` and use the current Tcl version instead.

lucamar:
I have updated the patch `NAMD-2.12.patch` in order to match Tcl release `8.6` when linking and the build works for both cuda and non cuda versions in my environment.

PR #295, opened 2017-09-01
[dom-mc] CPMD_4_1_CrayIntel_17_08

ajocksch:


easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-17.08-pat-645-nogpu.eb
```diff
@@ -0,0 +1,43 @@
...
+
+sources = ['/apps/common/UES/easybuild/sources/c/CPMD/cpmd-v%s.tar.gz' %version]
+
+patches = [('CRAY-XE6-INTEL-MKL', 'CPMD/configure')]
+
+dependencies = [('fftw/3.3.4.10', EXTERNAL_MODULE)]
+
+skipsteps = ['configure']
+
+common_prebuildopts = './configure.sh PREFIX=%(installdir)s CRAY-XE6-INTEL-MKL &&'
```
gppezzi: I don't see why you need to define this as a variable, couldn't you add this string directly to prebuildopts?

easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-17.08.eb
```diff
@@ -0,0 +1,37 @@
...
+prebuildopts = [
+                         common_prebuildopts,
+]
+#     'export omp=1 &&' + common_prebuildopts + ' sed -i \'s#cpmd\.x#cpmd_omp\.x#\' Makefile &&',
+
+files_to_copy = [(['bin/cpmd*.x'],'bin')]
+
+sanity_check_paths = {
+    'files': ['bin/cpmd.x'],
+#    'files': ['bin/cpmd.x', 'bin/cpmd_omp.x'],
```
gppezzi: do you need to keep this line?

easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-17.08-pat-645-nogpu.eb
ajocksch: If we want to keep the OpenMP option, which might work for a next release, the variable is used twice. Thus it is efficient.

easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-17.08.eb
ajocksch: Do we want to keep the OpenMP option?

gppezzi:
retest this please

jgphpc:
Let's try to not use `perftools*cscs` as explained in https://madra.cscs.ch/scs/doc/milestones/8

jgphpc:
Open another PR for cpmd+perftools please

https://jenkins.cscs.ch/job/TestingEB/587/label=dom/console

PR #301, opened 2017-09-04
[daint] Fixing CDO sources

lucamar:
The link given in the configuration file is not valid anymore, see the error message in the output below:
https://jenkins.cscs.ch/job/RegressionEB/label=daint/25/consoleFull

The current download link https://code.mpimet.mpg.de/projects/cdo/files does not provide release `1.8.1`, therefore we need to use the absolute path for this release.

### DIFF
--- a/easybuild/easyconfigs/c/CDO/CDO-1.8.1-CrayGNU-2016.11.eb
+++ b/easybuild/easyconfigs/c/CDO/CDO-1.8.1-CrayGNU-2016.11.eb
@@ -1,4 +1,4 @@
-# Updated by @gppezzi @jgphpc
+# contributed by Guilherme Peretti Pezzi, Jean-Guillaume Piccinali and Luca Marsella (CSCS)
 easyblock = 'ConfigureMake'
 
 name = 'CDO'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/c/CDO/CDO-1.8.1-CrayGNU-2016.11.eb
```diff
@@ -1,4 +1,4 @@
-# Updated by @gppezzi @jgphpc
+# contributed by Guilherme Peretti Pezzi, Jean-Guillaume Piccinali and Luca Marsella (CSCS)
```
jgphpc: i personally prefer @jgphpc instead of full name.

easybuild/easyconfigs/c/CDO/CDO-1.8.1-CrayGNU-2016.11.eb
```diff
@@ -10,8 +10,7 @@ description = """CDO is a collection of command line Operators to manipulate and
 toolchain = {'name': 'CrayGNU', 'version': '2016.11'}
 toolchainopts = {'opt': True, 'pic': True}
 
-sources = ['cdo-%(version)s.tar.gz']
-source_urls = ['https://code.zmaw.de/attachments/download/14271/']
+sources = ['/apps/common/UES/easybuild/sources/%(nameletterlower)s/%(name)s/' + SOURCELOWER_TAR_GZ]
```
jgphpc: Do you want to upgrade to :
* https://code.mpimet.mpg.de/attachments/download/15187/cdo-1.9.0.tar.gz
or even:
* https://code.mpimet.mpg.de/attachments/download/14686/cdo-1.8.2.tar.gz

lucamar: We should do that in `17.08`, it's not necessary for `2016.11`.

PR #302, opened 2017-09-04
[dom] boost/1.6x pe/1708

jgphpc:


jgphpc:
* I will try 1.65.

* I don't know how to use python3 from cray-python as there is only 1
modulefile...

jgphpc:
@victorusu  I am trying to write `Boost-1.65.0-CrayGNU-17.08-cray-python3.eb`
can you help me change the easyblock so that i can add `-with-python=python3` to bootstrap.sh ?

easybuild/easyconfigs/b/Boost/Boost-1.65.0-CrayGNU-17.08-cray-python2.eb
```diff
@@ -0,0 +1,28 @@
+# contributed by Luca Marsella (CSCS)
+name = 'Boost'
+version = '1.65.0'
+versionsuffix = '-cray-python2'
```
lucamar: What do you think of `-python2` only as versionsuffix? The toolchain `CrayGNU` already suggests the use of `cray-python` and the version `17.08` will hint at the use of version `17.06.1`.

jgphpc:
merge please

PR #313, opened 2017-09-05
[dom] XZ and dependencies

lucamar:
XZ with dependencies: `libxml2` might be an issue as in some cases it has `XZ` as a depedency in its turn.

### DIFF
--- a/easybuild/easyconfigs/x/XZ/XZ-5.2.3-CrayGNU-2016.11.eb
+++ b/easybuild/easyconfigs/x/XZ/XZ-5.2.3-CrayGNU-2016.11.eb
@@ -1,0 +1,31 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'ConfigureMake'
+
+name = 'XZ'
+version = "5.2.3"
+
+homepage = 'http://tukaani.org/xz/'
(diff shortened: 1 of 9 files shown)


jgphpc:
For reference:

```
autoconf-2.69-9.71.noarch
automake-1.13.4-6.2.noarch
libtool-2.4.2-16.1.x86_64
```

easybuild/easyconfigs/x/XZ/XZ-5.2.3-CrayGNU-2016.11.eb
```diff
@@ -0,0 +1,31 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'ConfigureMake'
+
+name = 'XZ'
+version = "5.2.3"
+
+homepage = 'http://tukaani.org/xz/'
+description = "xz: XZ utilities"
+
+toolchain = {'name': 'CrayGNU', 'version': '2016.11'}
```
jgphpc: 2016.11 still needed  ?

lucamar: Not really, it was an intermediate file before creating the final recipe.

PR #315, opened 2017-09-06
WIP [dom] Helper modules pointing to the new PE

vkarak:
I am proposing the introduction of the `dom-gpu` and `dom-mc` helper login packages as helpers to the `daint-gpu` and `daint-mc`. They will be practically the same except during the transition periods to a new PE, in which case the Dom scripts will be updated first to the new paths. This has the advantage that we can automate the testing of the new PE without having to memorise every time the correct module path to use and will also help the CI of Reframe while adapting or developing tests for the new PEs.

### DIFF
--- a/login/dom-gpu
+++ b/login/dom-gpu
@@ -1,0 +1,50 @@
+#%Module
+
+# This module is for testing purposes when upgrading the PE.
+
+# set environment
+set apps "$env(APPS)"
+set path "${apps}/UES/jenkins/6.0.UP04/ARCH/easybuild/modules/all"
+# set gpupath and mcpath
+set gpupath [string map { ARCH gpu } "${path}"]
+set mcpath [string map { ARCH mc } "${path}"]
+# list of modulefiles in gpupath
+set gpulist [exec ls --color=none "${gpupath}"]
+
+# checks for conflicts
+if { [is-loaded craype-broadwell] } {
+ module unload craype-broadwell
+ module unuse "${mcpath}"
+}
+# conflicts
+conflict craype-broadwell
+conflict daint-mc
+conflict dom-mc
+
+# module remove (unload) or switch (swap): unload dependencies
+if { [module-info mode remove] || [module-info mode switch] } {
+ foreach module ${gpulist} {
+#  if { [is-loaded ${module}] } {
+   module unload ${module}
+#  }
+ }
+}
+
+# load dependency and update MODULEPATH
+module load craype-haswell
+module use "${gpupath}"
+
+# load additional modules
+#module load xalt
+
+# print help screen
+if { [module-info mode help] } {
+proc ModulesHelp { } {
+ global apps gpupath
+ puts stderr "\t Set environment for gpu haswell compute nodes:
+  \t - module load craype-haswell
+  \t - module use ${gpupath}
+ "
+ return 0
+}
+}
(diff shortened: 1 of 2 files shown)


vkarak:
You mean prerequisite of the new PE Easybuild packages?

vkarak:
I see. I don't think we can do anything at the module level then. We should adapt the settings and the CI of reframe.

vkarak:
You can close this PR if you wish then, unless we find sth smarter that fixes this problem.

lucamar:
Let's mark it as `WIP` and let's think about a solution.

vkarak:
Is it possible to change manually the `daint-gpu` and `daint-mc` module files on Dom only? I've seen that they are different copies than those on Daint, aren't they? That would solve the problem with the CI. Currently, this is kind of a showstopper for adapting Reframe tests to the new PE, since there is no way to do `module use` depending on the partition a test will execute. This would require to manually test every new PR by manually using and unusing the corresponding module paths and manually selecting each time the GPU vs. the MC tests...

lucamar:
No, they are just symbolic links to the same modulefiles used on Piz Daint:
```
$ module avail daint-mc

-------------------------------------------------------- /apps/dom/UES/easybuild/modulefiles ---------------------------------------------------------
daint-mc
lucamar@dom101:~$ ls -l /apps/dom/UES/easybuild/modulefiles
lrwxrwxrwx 1 perettig csstaff 41 Jul 21 10:09 /apps/dom/UES/easybuild/modulefiles -> /apps/common/UES/jenkins/production/login
```

vkarak:
Would it make sense then `/apps/dom/UES/easybuild/modulefiles` to a real directory and copy and modify the modules accordingly on Dom. The `xalt` should remain a symlink to where it points now. Since the symlink is from Dom to Daint, unlinking and replacing the Dom should not affect the production setup.

lucamar:
Yes, as a workaround I have created the correct modulefiles for `dom` in `/apps/dom/UES/login/`.
I cannot modify myself the symbolic link `/apps/dom/UES/easybuild/modulefiles` as I don't have write permission in /apps/dom/UES/easybuild`, so I will ask the system administrators to do that.

victorusu:
`/apps/dom/UES/login` does not point to a valid easy build path.

vkarak:
I don't understand your comment. What is the problem with that?

lucamar:
Please do the following:
```
module use $APPS/UES/login 
module load daint-gpu
echo $MODULEPATH 
/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/modules/all:/apps/dom/UES/login:/opt/cray/pe/perftools/6.5.1/modulefiles:/opt/cray/pe/craype/2.5.12/modulefiles:/opt/cray/pe/modulefiles::/opt/cray/modulefiles:/opt/modulefiles:/apps/dom/modulefiles:/apps/dom/UES/easybuild/modulefiles:/apps/common/UES/modulefiles:/apps/common/system/modulefiles:/apps/dom/system/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/ari/modulefiles
```

lucamar:
I have managed to change the symbolic link: now you will see the dom specific modulefiles `daint-gpu` and `daint-mc` loading the new software stack in the `MODULEPATH` directly upon login.

vkarak:
Thanks @lucamar! It seems fine now!

PR #324, opened 2017-09-07
[dom] QuantumESPRESSO 6.1.0 with updated production lists

lucamar:
EasyBuild configuration file for QuantumESPRESSO and updated production lists: version 6.1.0 does not have a GPU port, so the same module will be built on both software stacks `gpu` and `mc`.

### DIFF
--- a/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.1.0-CrayIntel-17.08.eb
+++ b/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.1.0-CrayIntel-17.08.eb
@@ -1,0 +1,39 @@
+# contributed by Guilherme Peretti Pezzi and Luca Marsella (CSCS)
+easyblock = "ConfigureMake"
+
+name = 'QuantumESPRESSO'
+version = '6.1.0'
+
+homepage = 'http://www.quantum-espresso.org/'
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.1.0-CrayIntel-17.08.eb
```diff
@@ -0,0 +1,39 @@
...
+source_urls = ['http://qe-forge.org/gf/download/frsrelease/240/1075/']
+
+builddependencies = [
+    ('cray-fftw/3.3.6.2', EXTERNAL_MODULE),
+]
+
+preconfigopts = ' CC=cc LDFLAGS+="$LDFLAGS -qopenmp" && ' 
+preconfigopts += ' BLAS_LIBS="$MKLROOT/lib/intel64/libmkl_blas95_lp64.a" && ' 
+preconfigopts += ' LAPACK_LIBS="$MKLROOT/lib/intel64/libmkl_lapack95_lp64.a" && ' 
+preconfigopts += ' SCALAPACK_LIBS="-L$MKLROOT/lib/intel64 -lmkl_scalapack_lp64 -lmkl_intel_lp64 -lmkl_sequential -lmkl_core -lmkl_blacs_intelmpi_lp64" && '
```
jgphpc: line is a bit too long but there is probably no other way to do it.

PR #326, opened 2017-09-07
[dom-gpu] Adding AMBER CUDA pat version

victorusu:
This build should be done in the compute node. It does not work in the login nodes.

PR #328, opened 2017-09-08
Removing module purge from daint and dom

victorusu:
I do not understand why we have this module purge. 
It hurts on the builds when attempting to build with craypat support. 
See:
https://github.com/eth-cscs/production/pull/326
https://jenkins.cscs.ch/view/ProductionEB/job/TestingEB/551/label=dom/console

IMHO, there are two solutions, one is to remove the `module purge` and the other is to recover the needed folder by adding `module use ...`. I am choosing the first one if there is no other problem, that I cannot foresee.

### DIFF
--- a/jenkins-builds/production.sh
+++ b/jenkins-builds/production.sh
@@ -96,7 +96,7 @@ if [[ "$system" =~ "daint" || "$system" =~ "dom" ]]; then
         echo -e "\n No architecture defined. Please use the option -a,--arch to define the architecture \n"
         usage
     else
-        module purge
+        #module purge
         module load craype craype-network-aries modules ugni
         module load daint-${ARCH}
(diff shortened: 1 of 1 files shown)


victorusu:
@lucamar, what about this change?

lucamar:
@victorusu Please check the output below: the `MODULEPATH` is updated correctly
```
module purge
module load craype craype-network-aries modules perftools-base ugni
echo $MODULEPATH 
/opt/cray/pe/perftools/6.5.1/modulefiles:/opt/cray/pe/craype/2.5.12/modulefiles:/opt/cray/pe/modulefiles::/opt/cray/modulefiles:/opt/modulefiles:/apps/dom/modulefiles:/apps/dom/UES/easybuild/modulefiles:/apps/common/UES/modulefiles:/apps/common/system/modulefiles:/apps/dom/system/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/ari/modulefiles
```

victorusu:
Ok! then, `lgtm`

PR #330, opened 2017-09-08
[dom] Python extensions for cray-python

lucamar:
I propose the name `PyExtensions-%(version)s`, where `%(version)s` is the two digit Python version:
- I don't mention python in the name, the name itself suggests that it is based on Python
- the toolchain name `CrayGNU` clarifies that we are on a Cray system, hinting at `cray-python`
- the toolchain version `17.08` identifies uniquely the version of `cray-python` used as dependency

The extensions built in this recipe are the ones missing from `cray-python` with respect to Python modules built on the Cray so far: the list can be extended if needed (we already added new modules).

### DIFF
--- a/easybuild/easyconfigs/p/PyExtensions/PyExtensions-2.7-CrayGNU-17.08.eb
+++ b/easybuild/easyconfigs/p/PyExtensions/PyExtensions-2.7-CrayGNU-17.08.eb
@@ -1,0 +1,55 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'Bundle'
+
+name = 'PyExtensions'
+# python version used from cray-python
+version = '2.7'
+
+homepage = 'https://pypi.python.org/pypi'
+description = """This module is a bundle of Python packages on Cray systems based on the module cray-python"""
+
+toolchain = {'name': 'CrayGNU', 'version': '17.08'}
+toolchainopts = {'pic': True, 'verbose' : False }
+
+dependencies = [
+    ('cray-python/17.06.1', EXTERNAL_MODULE),
+]
+
+# bundle of Python packages
+exts_defaultclass = 'PythonPackage'
+
+exts_list = [
+     ('Cython', '0.26.1', {
+         'source_urls': ['https://pypi.python.org/packages/source/c/cython/'],
+     }),
+     ('six', '1.10.0', {
+         'source_urls': ['https://pypi.python.org/packages/source/s/six/'],
+     }),
+     ('matplotlib', '2.0.2', {
+         'source_urls': ['https://pypi.python.org/packages/source/m/matplotlib/'],
+     }),
+     ('pandas', '0.20.3', {
+         'source_urls': ['https://pypi.python.org/packages/source/p/pandas/'],
+     }),
+     ('pygelf', '0.3.1', {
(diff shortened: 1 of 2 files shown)


jgphpc:
* virtualenv/15.1.0' is missing in cray-python => add it to this PR ?

For reference, cray-python/17.06.1 provides:
* pip list

```
cycler (0.10.0)
dask (0.13.0)
functools32 (3.2.3.post2)
h5py (2.7.1)
jsonpickle (0.9.5)
lockfile (0.12.2)
matplotlib (2.0.2)
mpi4py (2.0.0)
nose (1.3.7)
numpy (1.11.3)
pip (9.0.1)
py (1.4.32)
pytest (3.0.5)
scipy (0.18.1)
setuptools (36.2.7)
six (1.10.0)
subprocess32 (3.2.7)
toolz (0.8.2)
vsc-base (2.5.8)
vsc-install (0.10.26)
vsc-utils (1.8.5)
wheel (0.29.0)
```

pip3 list
```
dask (0.13.0)
mpi4py (2.0.0)
nose (1.3.7)
numpy (1.11.3)
pip (8.1.1)
py (1.4.32)
pytest (3.0.5)
scipy (0.18.1)
setuptools (20.10.1)
toolz (0.8.2)
```

Cray may improve this:
* Why pip/811 instead of pip/901 in python3
* Why setuptools/20.10.1 in python while setuptools/36.2.7 in python2.
* Why not a more recent scipy/0.19.1 ?

lucamar:
It's better if you could open a Cray Case directly, also mentioning the old `cray-libsci` dependency.
Virtualenv is there, but they have changed the executable name to `pyvenv`:
```
module load cray-python
pyvenv -h
usage: venv [-h] [--system-site-packages] [--symlinks | --copies] [--clear]
            [--upgrade] [--without-pip]
            ENV_DIR [ENV_DIR ...]

Creates virtual Python environments in one or more target directories.

positional arguments:
  ENV_DIR               A directory to create the environment in.

optional arguments:
  -h, --help            show this help message and exit
  --system-site-packages
                        Give the virtual environment access to the system
                        site-packages dir.
  --symlinks            Try to use symlinks rather than copies, when symlinks
                        are not the default for the platform.
  --copies              Try to use copies rather than symlinks, even when
                        symlinks are the default for the platform.
  --clear               Delete the contents of the environment directory if it
                        already exists, before environment creation.
  --upgrade             Upgrade the environment directory to use this version
                        of Python, assuming Python has been upgraded in-place.
  --without-pip         Skips installing or upgrading pip in the virtual
                        environment (pip is bootstrapped by default)

Once an environment has been created, you may wish to activate it, e.g. by
sourcing an activate script in its bin directory.
```

lucamar:
I might try to create a custom Python package called `pyextensions` with `2.7` and `3.5`:
- we will be able to build Python extensions using the easyblock `PythonPackage` directly instead of `Bundle`, which makes some aspects of the build simpler
- we would be able to create symbolic links with standard names to Python executables, e.g. `venv -> pyvenv` or `python -> python3.5` if the module loaded is version `3.5` (`cray-python` uses `2.7` by default)

jgphpc:
We will need Luca's custom Python package called pyextensions

PR #331, opened 2017-09-08
[dom] Python dependencies

lucamar:
The recipes included in this pull request are the dependencies of Python modules built on Cray systems using EasyBuild: at present they are not listed as dependencies in `PyExtension`, but we still need to verify that they are not needed with `cray-python`: they might also be useful for other Python builds.


jgphpc:
Are any of the supported application depending on these recipes included in this pull request ?

lucamar:
@jgphpc not to my knowledge, only Python packages depend on these modules.

PR #332, opened 2017-09-08
[dom-gpu] CP2K on CrayPE 17.08

lucamar:
New CP2K recipes to be tested.


lucamar:
I had to link `CP2K` to the non threaded `cray-libsci` to avoid `Segmentation Fault` at runtime when using OpenMP threads. It was not needed on Piz Daint, although CP2K developers recommend it: 
https://www.cp2k.org/howto:compile

PR #345, opened 2017-09-13
[kesch] Changing easybuild flag in MCH list

lucamar:
The `--prefix` flag in the MCH production list is ignored by EasyBuild, therefore I changed it to `--installpath`. I have also updated `CMake` version from `3.7.2` to the more recent `3.9.1`.
The change of flag is needed to install modules in the paths that we have agreed with MCH:
https://mchwiki.cscs.ch/index.php/ESCHA_KESCH_SW_Naming_Conventions


jgphpc:
Why not using the "user view" so that you will not have to update the jenkins list:

```
Installation path (user view, link)
- /apps/escha/UES/generic/modulefiles --> /apps/escha/UES/RH7.3_generic/eb/modules/all
```

lucamar:
The `EASYBUILD_PREFIX` that we use on Kesch is `/apps/escha/UES/jenkins/RH7.3-17.02/easybuild`, so we don't need to specify that one with `--installpath`, as it will be the default install location.

PR #352, opened 2017-09-15
[dom] Hdfview

jgphpc:


jfavre:
is this intended for use only on the loggin nodes? hdfview is an interactive application AFAIK

PR #355, opened 2017-09-18
[dom] cpmd

jgphpc:


ajocksch:
retest this please

jgphpc:
The names refer to `CrayIntel-17.08.eb` but we use `CrayIntel-17.08g.eb` = intel/17.0.1 because cpmd compiled with intel/17.0.4 crashes.

PR #356, opened 2017-09-18
[dom-gpu] ParaView-5.4.1-EGL + Python3

jfavre:
…want Python3


jgphpc:
retest this please

jfavre:
The compilation still is not correct, in that it does not pick up python3, but uses python2 instead.

This behavior is not seen when I use the same recipe on the login node. In fact, I have discovered a config tool created by ParaView, which enables to verify the build. The command:

paraview-config --python-lib
correctly reports:    -l/opt/python/17.06.1/lib/libpython3.5m.so
when using my private module compile, whereas it reports:    -l/opt/python/17.06.1/lib/libpython2.7.so
when using the compilation done by jenkins on the CN. It's a mystery to me.

PR #361, opened 2017-09-19
[kesch] Update of CMake version

omlins:


lucamar:
retest this please

lucamar:
The build in Jenkins project `TestingEB` will fail, since the build that makes use of `CrayCCE` on `Kesch` need the standard EasyBuild module naming scheme and the architecture `haswell`:
`--module-naming-scheme=EasyBuildMNS --optarch=haswell`

PR #362, opened 2017-09-19
[dom-gpu] QuantumESPRESSO with perftools-lite/6.5.1

lucamar:
Unset the following variable defined by `perftools-lite/6.5.1` is needed during configuration: `CRAYPAT_COMPILER_OPTIONS`

The build would fail otherwise with the following error on the login nodes only:
```
checking whether the Fortran compiler works... configure: error: in `/dev/shm/lucamar/easybuild/stage/build/QuantumESPRESSO/6.1.0/CrayIntel-17.08-pat-6.5.1/qe-6.1':
configure: error: cannot run Fortran compiled programs.
If you meant to cross compile, use `--host'.
```

An alternative option might be using `--host`, as suggested by the error message itself.
The variable is then exported with the value `1` before building.

The previous recipe was instead successful on the compute nodes.

### DIFF
--- a/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.1.0-CrayIntel-17.08-pat-6.5.1.eb
+++ b/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.1.0-CrayIntel-17.08-pat-6.5.1.eb
@@ -23,13 +23,14 @@ builddependencies = [
     ('perftools-lite/%s' % patversion, EXTERNAL_MODULE)
 ]
 
-preconfigopts = ' CC=cc LDFLAGS+="$LDFLAGS -qopenmp" && ' 
+preconfigopts = ' unset CRAYPAT_COMPILER_OPTIONS && '
+preconfigopts += ' CC=cc LDFLAGS+="$LDFLAGS -qopenmp" && ' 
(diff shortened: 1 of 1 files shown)


lucamar:
retest this please

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.1.0-CrayIntel-17.08-pat-6.5.1.eb
```diff
@@ -23,13 +23,14 @@ builddependencies = [
     ('perftools-lite/%s' % patversion, EXTERNAL_MODULE)
 ]
 
-preconfigopts = ' CC=cc LDFLAGS+="$LDFLAGS -qopenmp" && ' 
+preconfigopts = ' unset CRAYPAT_COMPILER_OPTIONS && '
```
jgphpc: You could also try with:

```
preconfigopts = 'unset CRAYPAT_LITE &&'                                         
buildopts = ' CRAYPAT_LITE=sample_profile '               
```

lucamar: If I set `CRAYPAT_LITE=sample_profile` when building, can I change profile later at runtime?

lucamar:
Thanks @jgphpc, I have update the EasyBuild configuration file according to your advice. I don't know why the download failed yesterday, I have tried on the compute nodes today the same link and it worked.

PR #364, opened 2017-09-19
[dom] Support IPython-5.5.0 for CrayGNU-17.08 Python2

teojgo:


easybuild/easyconfigs/i/IPython/IPython-5.5.0-CrayGNU-17.08-python2.eb
```diff
@@ -0,0 +1,216 @@
...
+    ('cray-python/17.06.1', EXTERNAL_MODULE),
+    ('PyExtensions', pyshortver),
+    ('mock', '2.0.0', versionsuffix),
+    ('PyZMQ', '17.0.0b1', '%s-zmq4' % versionsuffix),
+]
+
+exts_defaultclass = 'PythonPackage'
+
+exts_list = [
+    ('nose', '1.3.7', {
```
jgphpc: nose/1.3.7 is part of cray-python but as they both use the same version it should not be an issue.

teojgo:
@jgphpc it is now ready to be merged.

PR #365, opened 2017-09-20
[dom-gpu] Add DCA++/17.08 easyconfigs.

uhaehner:


gppezzi:
ok to test

gppezzi:
retest this please (I forgot the add `dom` to the subject)

gppezzi:
@uhaehner would you mind changing your recipe to use the generic version of CMake? This can be achieved by changing the dependency line with:
```
    ('CMake', '3.9.1', '', True),
```
For our builds we decided to provide generic versions of CMake (instead of using the full Cray toolchain), but if that causes a problem just let me know.

PR #370, opened 2017-09-20
[dom-gpu] CMake in 6.0.UP04-17.08-gpu

lucamar:
Inserting `CMake-3.9.1.eb` (dummy) to make it visible in the software stack


lucamar:
I will merge this request as discussed with @gppezzi

PR #372, opened 2017-09-20
[dom-mc] CMake in 6.0.UP04-17.08-mc

lucamar:
Build `CMake` release `3.9.1` (dummy) to have it listed as a visible modulefile in the `mc` software stack as well.

### DIFF
--- a/jenkins-builds/6.0.UP04-17.08-mc
+++ b/jenkins-builds/6.0.UP04-17.08-mc
@@ -5,6 +5,7 @@
  Boost-1.65.0-CrayGNU-17.08.eb
  Charm++-6.8.0-CrayGNU-17.08.eb                     --set-default-module --hidden
  Charm++-6.8.0-CrayIntel-17.08.eb                   --hidden
+ CMake-3.9.1.eb
  CP2K-5.0r18043-CrayGNU-17.08.eb                    --set-default-module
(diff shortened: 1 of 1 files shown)


jenkins-builds/6.0.UP04-17.08-mc
```diff
@@ -5,6 +5,7 @@
  Boost-1.65.0-CrayGNU-17.08.eb
  Charm++-6.8.0-CrayGNU-17.08.eb                     --set-default-module --hidden
  Charm++-6.8.0-CrayIntel-17.08.eb                   --hidden
+ CMake-3.9.1.eb
```
jgphpc: and 6.0.UP04-17.08-gpu ?

jgphpc: => https://github.com/eth-cscs/production/pull/370

jgphpc: --set-default-module ?

jgphpc: --set-default-module ?

lucamar: I think it makes sense to use the flag `--set-default-module` if you have more than one release installed. So I believe that we can use the flag as soon as we decide to build a new release.

PR #373, opened 2017-09-20
[dom] extrae pprocessing tool

jgphpc:


jgphpc:
Yes i will hide it. I will add it as a dependency to extrae (next time).

PR #375, opened 2017-09-20
[daint,dom] Paraview with python3 include and libs

jfavre:


jfavre:
Thanks @lucamar for teh suggestion to force re-build Python 3 is now picked up correctly

I have tested in interactive client-server mode and things are now python3-compliant

lucamar:
@jfavre Can we close this pull request, since the force build of Paraview worked?

PR #379, opened 2017-09-20
[kesch] Python extensions for MCH

lucamar:


lucamar:
I have changed the module name according to the proposal that we have discussed. 
I build the python packages withing the extension list, I only kept `cairo` as a dependency as it involves more python packages to be built and `IPython`, as it is a bigger package. 
Please let me know if you prefer to include these packages within the `Python 3.6.2` module.

lucamar:
retest this please

kraushm:
In order to further facilitate the whole thing, the extensions provided inside this module will be added to the module Python 3.6.2 gmvolf 17.02. This way, only one single module has to be considered by the user and by us. (IPython will become a module apart in any case)

lucamar:
Thanks @kraushm, I will close this pull request.

PR #384, opened 2017-09-26
[dom] Reframe 2.6.1 patch

vkarak:
This provides an updated package for 2.6.1 that applies a patch for removing the maintenance tag from the HPL and HPCG tests.


vkarak:
retest this please

vkarak:
Since Dom is down and may not come back until the next maintenance, I manually patched the installed tests.

PR #393, opened 2017-10-02
[daint,dom] Fixing Boost suffixversion

victorusu:


victorusu:
retest this please

victorusu:
None, in the production list... But moose does...

jgphpc:
Do you have a tool to find out which recipes depend on Boost ?

victorusu:
None depends... If any depended the build would fail...

PR #395, opened 2017-10-02
[kesch] Python 3.6.2/no py ext file

kraushm:
Python 3.6.2 recipe including the extensions for MCH, in order to reduce module loading time.


kraushm:
Still need to find a way to get the cairo working in the extension list, which requires extra builddependencies and configopts.

kraushm:
retest this please

PR #397, opened 2017-10-03
[dom] Update VTK-8.0.1-EGL-CrayGNU-17.08-python2.eb

victorusu:


jfavre:
failing because the SOURCE_TAR_GZ is now including "-EGL" and failing to find the tar file

victorusu:
retest this please

PR #409, opened 2017-10-04
[daint-gpu,daint-mc] NetCDF 4 python

victorusu:
This fixes netcdf-python easyblock and updates netcdf-python to use the last toolchain and modules.


victorusu:
retest this please

PR #411, opened 2017-10-05
[daint] PLUMED GNU in production lists

lucamar:
As discussed at the last SCS meeting, I'm inserting PLUMED GNU in both production lists.


victorusu:
`latm`

PR #413, opened 2017-10-05
[ela] python package for eb dependency graph

gppezzi:
To be used on ela, it can be tested here:
```
module use /apps/ela/UES/SLES11/easybuild/modules/all
```

### DIFF
--- a/easybuild/easyconfigs/p/python-graph-dot/python-graph-dot-1.8.2-Python-2.7.13.eb
+++ b/easybuild/easyconfigs/p/python-graph-dot/python-graph-dot-1.8.2-Python-2.7.13.eb
@@ -1,0 +1,26 @@
+easyblock = 'PythonPackage'
+
+name = 'python-graph-dot'
+version = '1.8.2'
+versionsuffix = '-Python-2.7.13'
+
+homepage = 'https://github.com/pmatiello/python-graph'
+description = "python-graph-dot"
+
+toolchain = {'name': 'dummy', 'version': ''}
+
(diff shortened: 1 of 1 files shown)


lucamar:
I get this error message on Ela:
```
module use /apps/ela/UES/SLES11/easybuild/modules/all
module load python-graph-dot/1.8.2-Python-2.7.13
module load EasyBuild-custom
eb netcdf4-python-1.2.9-gmvolf-17.02-Python-3.6.0.eb --dep-graph=deps.txt --detect-loaded-modules=ignore
== temporary log file in case of crash /dev/shm/lucamar/easybuild/stage/tmp/eb-juT_l5/easybuild-0Jp6O9.log
ERROR: ImportError: None of the specified modules gv is available; required module 'gv' is not available (provided by Python package graphviz, available from https://pypi.python.org/pypi/graphviz)
```

victorusu:
@lucamar, can you change your `--dep-graph=deps.txt` to `--dep-graph=deps.dot`?
I works for me when I use the `.dot` extension...

lucamar:
Thanks @victorusu, it worked following your advice: it seems that only the extension `.dot` is allowed for the graph file...
```
eb netcdf4-python-1.2.9-gmvolf-17.02-Python-3.6.0.eb --dep-graph=deps.dot --detect-loaded-modules=ignore
== temporary log file in case of crash /dev/shm/lucamar/easybuild/stage/tmp/eb-WAfaE5/easybuild-dxPE61.log
Wrote dependency graph for 1 easyconfigs to deps.dot
```

PR #431, opened 2017-10-16
Add a new directory `jenkins` for Jenkinsfiles

teojgo:
* Create a groovy file `Machines.groovy` containing
the configurations of the machines available at CSCS.

* Create a groovy file `util.groovy` where various
utility methods used in jenkins pipelines are defined.

* Create the jenkins pipeline file `JenkinsfileRegressionEB`
which is used periodically to run a regression suite for
Easybuild.


teojgo:
The checks are going to fail since no machine is defined because no easybuild file is added.

PR #433, opened 2017-10-16
[dom] und/40913

jgphpc:
fix for https://madra.cscs.ch/scs/doc/issues/165
It will fail when a new version is released which is ok.


jgphpc:
https://jenkins.cscs.ch/job/TestingEB/838/label=dom/console :ok: 
merge please

PR #448, opened 2017-10-20
[daint dom] Added cxx flag to bazel build command to make a distributed ML plugin…

mschoengens:
… work with TF 1.3.0.


gppezzi:
retest this please

gppezzi:
@mschoengens do you want to re-deploy TF with this new recipe or do you want to have an alternative module with this new option?

mschoengens:
I would just redeploy it, the option is required if you want to add your own tf operations and does no harm otherwise.

PR #456, opened 2017-10-25
[kesch] Replace deprecated tuple syntax with dictionary in sources

teojgo:


gppezzi:
@teojgo does this fix an error or just a warning?

In case of an error, can you please provide the error log (path) of the problem that this is supposed to fix?

teojgo:
It corrects the deprecation warning but the error seemed to be the download in the case of Szip which now seems to work.

teojgo:
retest this please

PR #459, opened 2017-10-26
[kesch] IDL 8.5.1

kraushm:
Installation recipe for IDL on kesch-tds making use of the "silent" mode of the installation script, i. e. answers to the questions asked during the installation process are provided in a text file.


kraushm:
Unfortunately, the script fails. The reason seems to be that the installation command is executed more than once, because during the installation process  the question for accepting the license terms reoccurs. This restart has to lead to an error, since at the end of the first installation, the 'install.sh' apparently removes the installation files.
At the same time, the installdir contains the complete and working installation of idl in easybuild/software/idl..., only the module files are not created.

easybuild/easyconfigs/i/IDL/IDL.eb
```diff
@@ -0,0 +1,28 @@
...
+
+homepage = 'http://www.harrisgeospatial.com/SoftwareandTechnology/IDL.aspx'
+
+description = """
+IDL, short for Interactive Data Language, is a programming language 
+used for data analysis. It is popular in particular areas of science, 
+such as astronomy, atmospheric physics and medical imaging. 
+"""
+
+toolchain = {'name': 'gmvolf', 'version': '17.02'}
```
gppezzi: for binary packages I suggest to use the dummy toolchain:
```
```

easybuild/easyconfigs/i/IDL/IDL.eb
```diff
@@ -0,0 +1,28 @@
...
+
+# File was provided
+sources = [{
+    'filename' : 'idl85envi53sp1linux.x86_64.tar.gz',
```
gppezzi: if the download url is not available, please use the absolute file path here

lucamar:
I did not manage to work around the issue encountered by @kraushm, using the simple command: 
```
echo -e "y \n%(installdir)s \ny \nn \ny \nn \ny \nn \ny" | ./install.sh
```
Therefore I have adapted the easyblock and easyconfig files provided by JSC for IDL:
 https://github.com/easybuilders/JSC/blob/master/Custom_EasyBlocks/2016a/idl.py
https://github.com/easybuilders/JSC/blob/master/Golden_Repo/2017b/i/IDL/IDL-8.5.1-GCCcore-7.2.0.eb

I kept the `dummy` toolchain with no dependency: we might consider using GCCcore as well as the X11 dependency,

easybuild/easyconfigs/i/IDL/IDL-8.5.1.eb
```diff
@@ -0,0 +1,34 @@
+# This file is part of JSC's public easybuild repository (https://github.com/easybuilders/jsc)
+# adapted by Matthias Kraushaar and Luca Marsella (CSCS)
+name = "IDL"
+version = "8.5.1"
+shortversion = '85'
+
+homepage = 'http://www.harrisgeospatial.com/SoftwareandTechnology/IDL.aspx'
+
+description = """IDL (Interactive Data Language) is a scientific programming language used across disciplines to extract meaningful visualizations from complex numerical data """
```
gppezzi: please split the long line

kraushm:
I removed the dependency on X11 from the build. I tried with a small script based on an example from the internet, whether graphs will be displayed, and so far it worked. 
Regarding the LD_LIBRARY_PATH, I just commented that part for the moment, since I was not able to test it myself.

gppezzi:
Can you please also add this file to the production list?

PR #473, opened 2017-11-06
[leone] Yade 2017.01a and  Py 2.7.12

gppezzi:
I've adapted @ChristopherBignamini's recipes to re-use the installed libraries and to avoid duplicates.


gppezzi:
retest this please

gppezzi:
@lucamar most of the python modules have specific dependencies so I don't think it is worth to create a bundle here, but we can reconsider that if this helps to speed up module loading (which is pretty slow on leone).

PR #480, opened 2017-11-07
[daint dom] Use 6.0.UP04 instead of 6.0.UP02 in xalt table of production.sh

teojgo:
* Use 6.0.UP04 when updating the xalt table of modulefiles


lucamar:
retest this please

PR #484, opened 2017-11-09
[daint dom] Adios with CrayIntel 17.08

lucamar:
Adios configuration file as requested by CSCS #29730.


lucamar:
@jfavre Thanks for the comment, I have adapted the configuration file previously used by the user. Please add to the file in the branch the libraries that you think might be useful in production.

jfavre:
Does that mean the user is not using any parallel I/O options? My favorite type of users.
We ought to discuss this in a group meeting. Should we be pro-active at promoting parallel IO API such as ADIOS, or should we be reactive to users requests?

lucamar:
I think it is a good idea to improve the build with parallel I/O: the user can always choose to keep the bare minimum configuration that he initially created, if this is his informed choice. 
I have an old configuration file for `adios` that used the following flags:
```
--with-mxml=/apps/daint/mxml/2.6/gcc-472 --with-lustre=/usr/lib64 --with-cray-pmi-incdir=/opt/cray/pmi/default/include --with-cray-pmi-libdir=/opt/cray/pmi/default/lib64 --with-cray-ugni-incdir=/opt/cray/gni-headers/default/include --with-cray-ugni-libdir=/opt/cray/ugni/default/lib64
```
I can easily insert this in the 'configureopts', as we have `mxml-2.10-CrayGNU-17.08.eb` as a hidden dependency already in the software stack on Piz Daint. insert as well your configuration lines.

PR #485, opened 2017-11-13
[dom] extrae/352

jgphpc:
Will fix https://madra.cscs.ch/scs/doc/issues/172


jgphpc:
Extrae-3.5.2-gmvapich2-17.02.eb will not build but that's ok.

PR #497, opened 2017-11-20
[kesch leone monch] CMake 3.10.0 rc5

lucamar:
Building CMake now requires `-std=c++11`, that is not provided by `GCC 4.4.7` available on the operating system of Kesch, Leone and Monch: therefore on these systems we are forced to build `CMake` using a toolchain providing a more recent version of `GCC` with respect to the `dummy` toolchain.


teojgo:
retest this please

lucamar:
retest this please

lucamar:
retest this please

lucamar:
retest this please

PR #498, opened 2017-11-22
[daint-gpu] Extend the list of default packages LAMMPS

victorusu:
As requested by an user.
I am extending the list of programs supplied by LAMMPS-cuda.


victorusu:
The user has confirmed that the build works for him ticket 29877.

victorusu:
retest this please

victorusu:
@lucamar, I also checked the documentation and I ran the Regression and the CrayPAT tests. Both passed. So far, I did not find any issues.

PR #503, opened 2017-11-28
[kesch] Ruby mail 2.7.0 -> 2.6.3

kraushm:
revert mail to 2.6.3 due to issues experienced by MCH while testing


kraushm:
mail 2.7.0 got installed by another module (probably pony itself), therefore the solution at the end of installing and uninstalling, instead of simply directly installing the required version.

kraushm:
User was able to adapt his software to the new environment.

PR #505, opened 2017-11-30
[dom daint kesch monch leone] Check for empty unuse path

teojgo:
* If the unusePath is empty then do not use the corresponding option.

* If the unusePath is not empty then use the unusePath option making
  the necessary replacements if architectures are defined.


lucamar:
Thanks @teojgo : anyway, if the variable `unuse_path` is empty, the use of an empty `module unuse ...` command does not produce any effect on the variable `MODULEPATH`. 

It would be useful though to print in the output the variable `unuse_path` if not empty, as this as an effect on the build.

gppezzi:
@lucamar the problem is that module use/unuse is not bullet proof, so it is better to avoid calling it if it is not strictly necessary.

PR #508, opened 2017-12-04
[daint dom kesch monch leone] Revert removing of unuse paths for kesch and leone

teojgo:


gppezzi:
@teojgo can you also revert the path for leone? 

Is it written somewhere the descriptions of these variables? If not, how about adding a comment in the header of this `Machines` file?

For example, what is `toolkitVersions` used for?

PR #518, opened 2017-12-18
[daint dom kesch] New CMake version

lucamar:
Operating system gcc compilers on `Leone` and `Monch` are version `4.4.7`, too old to have `--std=c++11` which is now required to build `CMake`. Therefore we cannot use `dummy` on these systems for `CMake`.


lucamar:
retest this please

PR #523, opened 2017-12-19
[leone] LIGGGHTS (used by Hilti)

kraushm:
Based on https://webrt.cscs.ch/Ticket/Display.html?id=30237

Module file has been tested. (see ticket)


easybuild/easyconfigs/l/LIGGGHTS/LIGGGHTS-3.8.eb
```diff
@@ -0,0 +1,25 @@
+easyblock = 'Bundle'
```
gppezzi: Please add as comment the name of the person in charge of this manual installation and the ticket number

PR #530, opened 2018-01-12
[dom daint leone kesch monch] Add easyconfig for GC3Pie version 2.4.2

teojgo:


easybuild/easyconfigs/g/GC3Pie/GC3Pie-2.4.2.eb
```diff
@@ -0,0 +1,110 @@
+easyblock = 'Bundle'
+
+name = 'GC3Pie'
+version = '2.4.2'
+
+homepage = 'https://gc3pie.readthedocs.org'
+description = """GC3Pie is a Python package for running large job campaigns on diverse batch-oriented execution
+ environments."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
lucamar: @teojgo You can as well leave the field `version` as an empty string, i.e. ''

PR #535, opened 2018-01-24
[daint-gpu] Build scripts for version 1.4.1 of TensorFlow.

mschoengens:


easybuild/easyconfigs/c/cuDNN/cuDNN-7.0-cuda-8.0.eb
```diff
@@ -0,0 +1,35 @@
...
+cudaminor = '61_2.4.3-6.0.4.0_3.1__gb475d12'
+cudaver = "%s.%s" % (cudamajor, cudaminor)
+
+versionsuffix = '-cuda-%s' % cudamajor
+
+dependencies = [
+    ('cudatoolkit/%s' % cudaver, EXTERNAL_MODULE),
+]
+
+sources = ['/apps/daint/UES/6.0.UP02/sandbox-ds/soures/cudnn-8.0-linux-x64-v7.tgz']
```
gppezzi: We're trying to gather all the sources (which cannot be automatically downloaded) in the same folder:
`/apps/common/UES/easybuild/sources/c/cuDNN/`. Could you please copy the new tgz here and update the path?

PR #537, opened 2018-01-27
[daint-gpu] otf2

jgphpc:
C++ and Python3 api

### DIFF
--- a/jenkins-builds/6.0.UP04-17.08-mc
+++ b/jenkins-builds/6.0.UP04-17.08-mc
@@ -35,6 +35,7 @@
  ncview-2.1.7-CrayGNU-17.08.eb                      --set-default-module
  netcdf-python-1.2.9-CrayGNU-17.08-python2.eb
  netcdf-python-1.2.9-CrayGNU-17.08-python3.eb       --set-default-module
+ otfprofile-1.20.eb                                 --set-default-module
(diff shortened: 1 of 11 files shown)


jgphpc:
retest this please

jenkins-builds/6.0.UP04-17.08-mc
```diff
@@ -35,6 +35,7 @@
  ncview-2.1.7-CrayGNU-17.08.eb                      --set-default-module
  netcdf-python-1.2.9-CrayGNU-17.08-python2.eb
  netcdf-python-1.2.9-CrayGNU-17.08-python3.eb       --set-default-module
+ otfprofile-1.20.eb                                 --set-default-module
```
gppezzi: Why this is visible to everyone? Was this a request from (a group of) users? 

If not, this should go with `--hidden`

jgphpc: It is not a dependency, it is a tool that should be part of scorep, as it will help users to quickly check there traces before launching Vampir. I will add it to the recipe of scorep next time we build it but for now it should be visible.

gppezzi: Our policy is that only 'supported software' should be visible. 

If you want it visible then please add it to the documentation on the user portal.

jgphpc:
https://madra.cscs.ch/ues/userportal/merge_requests/199

PR #539, opened 2018-02-01
[daint dom leone kesch monch] ReFrame 2.9 easyconfig.

vkarak:


vkarak:
retest this please

vkarak:
Should I make sure to pass on Monch and Kesch? Ideally, I would like to, but I haven't looked yet into the problem there.

vkarak:
@lucamar It seems that `jenscscs` on Mönch cannot clone the repository. I tried it myself as `jenscscs` and it asks for a password.

vkarak:
It seems that monch has no access to madra (even from my account):

```
[13:13:06] karakasv@monch01 ~ $ wget https://madra.cscs.ch/scs/reframe/repository/v2.9/archive.tar.gz
--2018-02-02 13:13:09--  https://madra.cscs.ch/scs/reframe/repository/v2.9/archive.tar.gz
Resolving madra.cscs.ch (madra.cscs.ch)... 148.187.18.22
Connecting to madra.cscs.ch (madra.cscs.ch)|148.187.18.22|:443...
```

vkarak:
I will remove Monch from the list of environments, since it takes too long to enable access to madra.

vkarak:
retest this please

easybuild/easyconfigs/p/Python-bare/Python-bare-3.6.2.eb
```diff
@@ -20,7 +20,6 @@ sources = ['Python-%(version)s.tgz']
 dependencies = [
```
gppezzi: Since you're changing the dependencies here, you could also change the toolchain version to:
```
```
With the `dummy` toolchain version I think EB does not actually load the dependencies when building (not sure if this is a bug or a feature).

BTW, in the future we should consider stop using dummy and use a proper toolchain for building python-bare (`GCC`) for example.

vkarak: Sure, I'll do that.

vkarak:
retest this please

vkarak:
I think this is now ready to be merged. Works on all systems.

PR #542, opened 2018-02-08
[daint, dom] GPAW and ASE python packages for CrayGNU 17.08

lucamar:
I have prepared an updated recipe for `GPAW` and the dependency `ASE` using the toolchain `CrayGNU` version `17.08`, based on request number 30693: both recipes load `PyExtensions` as dependency.


lucamar:
retest this please

PR #552, opened 2018-02-27
WIP [daint dom leone kesch monch] Fix Jenkinsfiles for leone git cloning

teojgo:
* Load internal git module which provides a newer version.

* Do the git cloning manually instead of using `checkout scm`.


teojgo:
@lucamar I have already tried with the default git and the ssh version of the repo and it doesn't work.

teojgo:
The problem with leone was fixed without changing the Jenkinsfile.

PR #557, opened 2018-03-08
[daint dom monch leone kesch] ReFrame 2.11 easyconfig

vkarak:
- Does not fetch the private `reframe-tests` repo. This needs to be discussed and resolved.


lucamar:
Thanks @vkarak: we would also need to address how to circumvent `daint-gpu` and `daint-mc` modules in order to move from the `common` list.

PR #559, opened 2018-03-09
Tmux 2.6 easyconfig

vkarak:
We have several user requests for tmux, so I think it's good to have it in the production lists.


vkarak:
The `/apps/ela/system/bin/tmux` is too old (version 1.4 !). I always had to install my version on Ela for my personal use.

lucamar:
@vkarak GIven that a recent version of `tmux` has been provided by `root` on the front end, I propose to remove it from the production lists on the different systems (we might still merge the recipe to the master though).

vkarak:
@lucamar Sure, I'll remove it from the production lists.

PR #562, opened 2018-03-15
[daint leone monch kesch] ddt/1811

jgphpc:
https://developer.arm.com/products/software-development-tools/hpc/arm-forge/release-history


jgphpc:
Since when do you add
`conflict	 craype-haswell`
?

PR #568, opened 2018-03-30
[dom daint kesch monch] Python-bare: Replace nose with pytest and plugins

teojgo:


gppezzi:
Changing existing modules is not considered best practices. If you need additional extensions, I'd suggest you provide a new version of this module (we should only force re-builds if there's a good reason for it).

Remember to update the jenkins build list with the new eb.

PR #586, opened 2018-04-19
[daint dom kesch monch] Add reframe-2.12-dev1.eb and Python-bare-3.6.5.eb

teojgo:


easybuild/easyconfigs/p/Python-bare/Python-bare-3.6.5.eb
```diff
@@ -0,0 +1,69 @@
+# @author: manitart
+# This version should be used only for replacing the system python, not for building scientific apps
+
+easyblock = 'EB_Python'
+
+name = 'Python-bare'
+version = "3.6.5"
```
vkarak: Just for consistency, you could use single-quotes here.

easybuild/easyconfigs/p/Python-bare/Python-bare-3.6.5.eb
```diff
@@ -0,0 +1,69 @@
...
+    ('ncurses', '6.0', '', True),
+]
+
+# workaround for being able to change name to Python-bare
+modextravars = {'EBROOTPYTHON': '%(installdir)s'}
+
+# order is important!
+# package versions updated Aug/2017
+exts_list = [
+    ('setuptools', '36.4.0', {
```
gppezzi: How about updating versions of setuptools, pip and virtualenv as well?

PR #587, opened 2018-04-19
 Adding helper scripts for managing eb files

gppezzi:
* `dep-count.sh`
  * script for counting number of times each dep is used given a jenkins production build file
* `convert-all-tc.sh` 
  *  script for in-place bulk creating EB files for a new toolchain (as requested by @victorusu )


gppezzi:
(dep-count.sh output)[https://madra.cscs.ch/snippets/6]

PR #590, opened 2018-04-24
[daint, dom] New release of QuantumESPRESSO

lucamar:
I have provided a recipe for `QuantumESPRESSO 6.2.1`, based on the request 31464. 

We still need to test the build, are there any known issues with this release?

### DIFF
--- a/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.2.1-CrayIntel-17.08.eb
+++ b/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.2.1-CrayIntel-17.08.eb
@@ -13,8 +13,9 @@ description = """Quantum ESPRESSO  is an integrated suite of computer codes
 toolchain = {'name': 'CrayIntel', 'version': '17.08'}
 toolchainopts = {'usempi': True, 'openmp': True}
 
-sources = ['qe-%(version)s.tar.gz']
-source_urls = ['http://qe-forge.org/gf/download/frsrelease/240/1075/']
+sources = ['/apps/common/UES/easybuild/sources/%(nameletterlower)s/%(name)s/' + 'qe-%(version)s.tar.gz']
+#sources = ['qe-%(version)s.tar.gz']
+#source_urls = ['http://qe-forge.org/gf/download/frsrelease/240/1075/']
 
 builddependencies = [
     ('cray-fftw/3.3.6.2', EXTERNAL_MODULE),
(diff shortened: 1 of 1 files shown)


lucamar:
Downloading the package did not work with `jenscscs`, while it worked for me:
```
ls -ltrh /apps/common/easybuild/sources/q/QuantumESPRESSO/qe-6.2.1.tar.gz
-rw-rw-r--+ 1 lucamar csstaff 25M Dec 21 16:31 /apps/common/easybuild/sources/q/QuantumESPRESSO/qe-6.2.1.tar.gz

ls -ltrh /scratch/snx3000/jenscscs/TestingEB/daint-mc/2466c8a/sources/q/QuantumESPRESSO/qe-6.2.1.tar.gz
-rw-r--r-- 1 jenscscs jenscscs 230K Apr 24 15:23 /scratch/snx3000/jenscscs/TestingEB/daint-mc/2466c8a/sources/q/QuantumESPRESSO/qe-6.2.1.tar.gz
```

lucamar:
retest this please

toxa81:
QE-6.2.1 runs.
The only issue is that it doesn't generate the xml file (-D__OLDXML must be set during compilation), so, for example, the AIiiDA parser fails.

lucamar:
Thanks for your feedback @toxa81, however I this option was never included in our builds, in fact the log file of the module in production on Piz Daint has only the following:
```
grep "^DFLAGS  " /apps/daint/UES/sandbox/lucamar/easybuild/software/QuantumESPRESSO/6.1.0-CrayIntel-17.08/easybuild/easybuild-QuantumESPRESSO-6.1.0-20171215.105807.log
DFLAGS         =  -D__OPENMP -D__FFTW3 -D__MPI
```
I believe that I can fix it by just adding `DFLAGS += -D__OLDXML` to the configure options in the EasyBuild recipe. By the way, checking the log I have found that `make.inc` now does not define `-D__OPENMP`, although I explicitly set `--enable-openmp` in the configure options: 
```
$ grep "^DFLAGS " /apps/daint/UES/sandbox/lucamar/easybuild/software/QuantumESPRESSO/6.2.1-CrayIntel-17.08/easybuild/easybuild-QuantumESPRESSO-6.2.1-20180424.145023.log
DFLAGS         =  -D__FFTW3 -D__MPI
```
Although the `OpenMP` parallelisation is not particularly efficient, I should probably add it manually to the `make.inc` file.

toxa81:
-D__OPENMP is obsolete. All OpenMP-enabled compilers defined _OPENMP macro. QE is testing for it now.

lucamar:
Thanks! The `release-notes` say that I can just use the configure option `--disable-xml` to get the flag that you suggested in the build; however this format will be incompatible with the new one: 
```
  grep -A 3 -B 1 OLDXML Doc/release-notes
  * The new XML format with schema is now the default. Use configure option
    "--disable-xml", or add -D__OLDXML to MANUAL_FLAGS in make.inc, to revert
    to the old xml format. IMPORTANT NOTICE: the new format is incompatibile
    both with the "old" format and with the previous "new" one: files may be
    in different locations with different names and contain different data.
```

PR #592, opened 2018-04-25
[daint dom] QuantumESPRESSO 6.2.1 with --disable-xml

lucamar:
I have inserted the flag `--disable-xml` in the `configure` step to allow compatibility with the AiiDA parser, as suggested by @toxa81


lucamar:
Thanks @toxa81 !

PR #593, opened 2018-04-26
[dom-gpu] Example COSMO with easybuild (please do not merge)

havogt:
I tried to build COSMO some time ago. I think the following was the status when I stopped.

Let me explain some of my "design" decisions (and please forgive if I mess up with the terminology or don't recall how it works, it was less than one day that I invested in easybuild in total).

The COSMO build (with the C++ dycore) has the following dependency (upper depends on lower)
- COSMO
- C++ dycore
- STELLA

Unfortunately STELLA has some compile time switches (like KFLAT, or single/double etc). When invoke "build COSMO I would like to specify these flags, i.e. on the upper most level, and propagate them down to STELLA.
As a hack I used the versionsuffix to encode this information. Therefore (at least that was my understanding) I needed to use EasyBlocks.

If I remember correctly, the version below works, except building COSMO. I think my conclusion was the the `MakeCp` easyblock is broken. The `files_to_copy` did not work and I couldn't fix it. (Of course the nice solution would be that COSMO comes with a proper install target in its `Makefile`; but you know, it's COSMO...)

Would be nice if you keep me in the loop on how you implement it, I assume it will be very useful for MeteoSwiss as well, but I didn't follow up on this, because of lack of time...

### DIFF
--- a/easybuild/easyblocks/cosmo_dycore.py
+++ b/easybuild/easyblocks/cosmo_dycore.py
@@ -1,0 +1,50 @@
+"""
+EasyBuild support for building and installing the COSMO C++ dynamical core
+
+@author: Hannes Vogt (CSCS)
+"""
+import glob
+import os
+import re
+import shutil
+from distutils.version import LooseVersion
+
+import easybuild.tools.environment as env
+import easybuild.tools.toolchain as toolchain
+from easybuild.easyblocks.generic.configuremake import ConfigureMake
+from easybuild.easyblocks.generic.cmakemake import CMakeMake
+from easybuild.framework.easyconfig import CUSTOM
+from easybuild.tools.build_log import EasyBuildError
+from easybuild.tools.filetools import download_file, extract_file, which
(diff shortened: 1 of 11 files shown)


lucamar:
test this please

lucamar:
Thanks @havogt: please note that the title of the pull request should match a production system (see https://github.com/eth-cscs/production/blob/master/README.md for a list), otherwise the test build will not run

lucamar:
test this please

havogt:
I think it will not work anyway. I got a request from Guilherme to share what I did with COSMO and I thought the easiest way would be to open a PR. It is not meant to be merged... If it is bothering your workflow, I can open the PR just in my fork. Let me know...

lucamar:
This is the right place to share your recipes and test them on our systems, provided that you have already managed to build them locally.

havogt:
Ok, moved to havogt/production#1.

gppezzi:
@havogt thanks for that, we start working on your files and we'll keep you posted. We'll either do a PR to your fork or a new PR here (and let you know).

PR #598, opened 2018-05-07
[daint dom] CP2K 5.1. with perftools-lite

lucamar:
Recipes to build `CP2K` release 5.1. instrumented with `perftools-lite`


lucamar:
I have tested the executables and they produce the report correctly: `/scratch/snx3000/lucamar//regression/production/stage/`

PR #599, opened 2018-05-08
[dom daint monch kesch] Remove leone from the list of machines

teojgo:
* Remove custom module definition for leone.


gppezzi:
@teojgo Only the user lab partition of leone has been decommissioned. We still might get requests for building software for the other partitions, so I suggest we close this PR for the moment.

PR #602, opened 2018-05-09
[daint-gpu] Add dask to Tensorfow 1.7.0 cray gnu 17.12

rsarm:


rsarm:
retest this please

gppezzi:
@rsarm a force re-build of TF is needed since the module is already installed.

PR #603, opened 2018-05-14
[daint, dom] Updated JenkinsfileTestingEB to use cscsci

lucamar:
I have tested the waiting times using the partition `cscsci`: in the worst case I waited for `~ 3` minutes, but the average waiting time is less than a minute on average. Therefore I suggest to use the partition in the project `TestingEB`.

### DIFF
--- a/jenkins/JenkinsfileTestingEB
+++ b/jenkins/JenkinsfileTestingEB
@@ -64,7 +64,7 @@ stage('Build Stage') {
                                                 echo \$SCRATCH""").trim()
                     def buildPath = "/dev/shm/jenscscs/$gitCommit/"
                     def prefix = "$scratch/$projectName/$machineLabel/$gitCommit"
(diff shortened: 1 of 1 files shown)


jenkins/JenkinsfileTestingEB
```diff
@@ -64,7 +64,7 @@ stage('Build Stage') {
                                                 echo \$SCRATCH""").trim()
                     def buildPath = "/dev/shm/jenscscs/$gitCommit/"
                     def prefix = "$scratch/$projectName/$machineLabel/$gitCommit"
-                    def command = machineName == 'dom' ? "srun -u --constraint=$arch --job-name=${projectName} --time=06:00:00" : ''
+                    def command = machineName == 'daint' || 'dom'? "srun -u --constraint=$arch --job-name=${projectName} --partition=cscsci --time=06:00:00" : ''
```
teojgo: Change to 
```
def command = machineName in ['daint', 'dom'] ? "srun -u --constraint=$arch --job-name=${projectName} --partition=cscsci --time=06:00:00" : ''
```

PR #604, opened 2018-05-16
[daint-gpu, dom-gpu]

toxa81:
A small change in the MAGMA recipe to copy ./control/*h header files, which are required by the end user applications.


toxa81:
@victorusu Victor, can you please approve it?

PR #622, opened 2018-05-28
[dom daint kesch monch] Bugfix/jenkins production pipeline

teojgo:


teojgo:
@lucamar don't merge after review, the command has to be enabled

PR #624, opened 2018-05-29
[daint-gpu dom-gpu] Updating version of cray-python for TensorFlow-1.3.0 and 1.4.1

### LINKED ISSUE #614: TensorFlow builds FAILED on daint and dom
The following `RegressionEB` build of `TensorFlow` failed on `daint` and `dom`: 
```
TensorFlow-1.3.0-CrayGNU-17.08-cuda-8.0-python3.eb
TensorFlow-1.4.1-CrayGNU-17.08-cuda-8.0-python3.eb
```

Link to the logs:
https://jenkins.cscs.ch/blue/organizations/jenkins/RegressionEB/detail/RegressionEB/771/pipeline/38

rsarm:
Fixes #614 

Some dependencies of TensorFlow-1.3.0 and 1.4.1 would conflict with other easyconfigs if the `cray-python` version was replaced. To avoid breaking recipes, new versions of the dependencies were created which use `cray-python/3.6.1.1` and were labeled `-python3.6`, opposed to the ones labeled `-python3`, which still use `cray-python/2017.0.6.1`

### DIFF
--- a/easybuild/easyconfigs/t/TensorFlow/tensorflow-v1.3.0-cuda-python3-cscs.patch
+++ b/easybuild/easyconfigs/t/TensorFlow/tensorflow-v1.3.0-cuda-python3-cscs.patch
@@ -27,8 +27,8 @@ index 0000000..11f0357
 +
 +CONF_IN="configure.in"
 +
-+echo $EBROOTPYTHON"/bin/python"$PYMAJVER > $CONF_IN
-+echo $EBROOTPYTHON"/lib/python"$PYMAJVER"."$PYMINVER"/site-packages" >> $CONF_IN
++echo "/opt/python/$PYMAJVER.$PYMINVER.$PYREVVER/bin/python"$PYMAJVER > $CONF_IN
++echo "/opt/python/$PYMAJVER.$PYMINVER.$PYREVVER/lib/python"$PYMAJVER"."$PYMINVER"/site-packages" >> $CONF_IN
 +echo "n" >> $CONF_IN # Do you wish to build TensorFlow with MKL support? [y/N] n
 +echo "-march=native" >> $CONF_IN # Please specify optimization flags "--config=opt" :
 +echo "y" >> $CONF_IN # Do you wish to use jemalloc as the malloc implementation? [Y/n] y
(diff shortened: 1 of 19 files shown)


easybuild/easyconfigs/t/TensorFlow/tensorflow-v1.3.0-cuda-python3-cscs.patch
```diff
@@ -27,8 +27,8 @@ index 0000000..11f0357
 +
 +CONF_IN="configure.in"
 +
-+echo $EBROOTPYTHON"/bin/python"$PYMAJVER > $CONF_IN
-+echo $EBROOTPYTHON"/lib/python"$PYMAJVER"."$PYMINVER"/site-packages" >> $CONF_IN
++echo "/opt/python/$PYMAJVER.$PYMINVER.$PYREVVER/bin/python"$PYMAJVER > $CONF_IN
```
gppezzi: I would suggest to use the EB variable: `$EBROOTCRAYPYTHON` 

Here and in the other patches as well.

rsarm: `cray-python/3.6.1.1` module doesn't define `$PYTHON_PATH` so the `$EBROOTPYTHON` variable is not created automatically by EasyBuild. It stopped working when moving from `cray-python/17.06.1` which do define `$PYTHON_PATH`.

The SWIG-3.0.10-CrayGNU-17.08-python3 and -python3.6 easyconfigs also use
`$EBROOTCRAYPYTHON` for a `configopts`, which seems to be ignored. @lucamar and @gppezzi should we replace there also `$EBROOTCRAYPYTHON` by `/opt/python/$PYMAJVER.$PYMINVER.$PYREVVER/`? Or is there a better solution?

gppezzi: Just for the record: we should continue to use `$EBROOTPYTHON`, since `Python` is the package name taken from the metadata file (`$EBROOTCRAYPYTHON` should not be used).

We're going to ask Cray to add a variable pointing to "python root" on their `cray-python` module file, in order to avoid hardcoding the path in the metadata file in the future.

PR #626, opened 2018-05-30
[dom-gpu] aprilPE

jgphpc:
april PE toolchains for testing cuda9 on dom


gppezzi:
LGTM

I'm just wondering how the cuda8 toolchains will co-exist with these new TC: we would need to add a new version suffix, such as `CrayGNU-18.04-gcc-5.3.0`

jgphpc:
> CrayGNU-18.04-gcc-5.3.0

I would compare `CrayGNU/.17.12` (=gcc/5) and `CrayGNU/.18.04` (=gcc/6) instead.

PR #630, opened 2018-06-04
[dom-gpu] Elavip

jgphpc:


easybuild/easyconfigs/v/Vim/Vim-8.1.eb
```diff
@@ -0,0 +1,28 @@
+easyblock = 'ConfigureMake'
+
+name = 'Vim'
+version = '8.1'
+
+homepage = 'http://www.vim.org'
+description = """ Vim is an advanced text editor that seeks to provide the power 
+ of the de-facto Unix editor 'Vi', with a more complete feature set. """
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
gppezzi: tc version should be empty like the other two files:
`toolchain = {'name': 'dummy', 'version': ''}`

(or else the 'dummy' dependencies are not loaded)

PR #632, opened 2018-06-05
[dom-gpu] h5py

lucamar:
Removed dependency `PyExtensions`


lucamar:
The build failed on `dom-gpu` for temporary issues, therefore I will retest only there.

PR #638, opened 2018-06-07
[daint] fix h5py-2.7.x 

gppezzi:
- EB 2.7.0 recipe was copied from latest version 2.7.1
- Added pyextensions (missing 'six' package)
- Enabled sanity checking for the serial versions (only the parallel versions have the `MPI_Init` issue and doing the check on the serial versions will ensure that importing h5py works )


gppezzi:
I was unsure, but now that Jenkins tested it, why not 👍

Issue #644, opened 2018-06-11
Build of Ruby-2.2.2-gmvolf-17.02.eb failed on Kesch


lucamar:
The build of `Ruby-2.2.2-gmvolf-17.02.eb` failed on `Kesch` while building an extension: the log file is `/scratch/jenscscs/RegressionEB/kesch/sources/eb-ePXReS/easybuild-Ruby-2.2.2-20180610.200407.wzGgN.log`

lucamar:
The build fails manually as well: however the more recent version `Ruby-2.4.0-gmvolf-17.02.eb` is already installed on `Kesch` and the build succeeds.

lucamar:
The old version of `Ruby` will soon be EOL (see https://www.ruby-lang.org/en/downloads).

lucamar:
The software folder and the modulefile of `Ruby/2.2.2-gmvolf-17.02` are copied by hand to the new installation path.

PR #647, opened 2018-06-12
WIP [kesch] Update MCH-RH7.3-17.02

lucamar:
Update version of Ruby on Kesch in production: version `2.4.0` was originally requested in RT #29008 and the previous version `2.2.2` is currently failing to build on the system, as reported in #644. The former modulefile will not be deleted and will still be available on the system.


gppezzi:
I'm not sure if MCH is ready to move to Ruby 2.4.0, please take a look at #29859 and sync with Matthias.

Maybe we can start by providing the an updated Ruby version with the required Gems and highlight that the outdated version is not building anymore. However, for the moment we cannot simply remove 2.2.2 as it is used in production.

lucamar:
I have contacted MCH in #32058

gppezzi:
I'm closing until we get feedback from MCH

PR #651, opened 2018-06-20
[daint,dom] Add HPX and jemalloc to production lists

msimberg:


victorusu:
We need to check if the `jemalloc` should be selected as a hidden dependency...

gppezzi:
dependencies should be hidden by default

victorusu:
I have approved the PR, but I have also made changes to it... It would be nice if we had another review... Or I should not have approved in the first place 🤔...

PR #668, opened 2018-07-10
[leone] LIGGGHTS 4.0 build recipe requested by Hilti

kraushm:
(see WebRT https://webrt.cscs.ch/Ticket/Display.html?id=32347)

According to the information, only the version number of LIGGGHTS was adapted.

Version 4.0 of LIGGGHTS was set as default version.


kraushm:
retest this please

PR #673, opened 2018-07-12
[dom-gpu]: ipm/cle6u06

jgphpc:
Will fix https://madra.cscs.ch/scs/doc/issues/318
waiting for CrayPGI toolchain

### DIFF
--- a/easybuild/easyconfigs/c/CrayPGI/CrayPGI-18.07.eb
+++ b/easybuild/easyconfigs/c/CrayPGI/CrayPGI-18.07.eb
@@ -15,6 +15,7 @@ dependencies = [
     ('PrgEnv-pgi', EXTERNAL_MODULE),
     ('atp/2.1.2', EXTERNAL_MODULE),
     ('pgi/18.3.0', EXTERNAL_MODULE),
+# libsci is not part of the default PrgEnv-pgi, do not load it
 #    ('cray-libsci/18.07.1', EXTERNAL_MODULE),
     ('cray-mpich/7.7.1', EXTERNAL_MODULE),
     ('craype/2.5.15', EXTERNAL_MODULE),
--- a/easybuild/easyconfigs/c/CrayPGI/CrayPGI-18.07.eb
+++ b/easybuild/easyconfigs/c/CrayPGI/CrayPGI-18.07.eb
@@ -15,7 +15,7 @@ dependencies = [
     ('PrgEnv-pgi', EXTERNAL_MODULE),
     ('atp/2.1.2', EXTERNAL_MODULE),
     ('pgi/18.3.0', EXTERNAL_MODULE),
-    ('cray-libsci/18.07.1', EXTERNAL_MODULE),
+#    ('cray-libsci/18.07.1', EXTERNAL_MODULE),
     ('cray-mpich/7.7.1', EXTERNAL_MODULE),
     ('craype/2.5.15', EXTERNAL_MODULE),
     ('pmi/5.0.14', EXTERNAL_MODULE),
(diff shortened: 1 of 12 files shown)


jgphpc:
retest this please

easybuild/easyconfigs/c/CrayPGI/CrayPGI-18.07.eb
```diff
@@ -15,7 +15,7 @@ dependencies = [
     ('PrgEnv-pgi', EXTERNAL_MODULE),
     ('atp/2.1.2', EXTERNAL_MODULE),
     ('pgi/18.3.0', EXTERNAL_MODULE),
-    ('cray-libsci/18.07.1', EXTERNAL_MODULE),
+#    ('cray-libsci/18.07.1', EXTERNAL_MODULE),
```
gppezzi: Can you please delete this line instead of commenting it out?

jgphpc: I would rather keep it as a reminder that we should not have libsci. 
One recently felt in that trap.

lucamar: I suggest to explain in a comment why the line is commented out

gppezzi: Since libsci is not part of the default PrgEnv-pgi, please just remove it. 

The only reason why one felt in the trap, is because the was a commented out entry without any explanation.

If you wish to add reminders on the files they should be understandable to everyone and not only for you.

We always create new recipes based on the latest available, if libsci is not listed originally then we're safe.

easybuild/easyconfigs/i/IPM/IPM-2.0.6-CrayGNU-18.07.eb
```diff
@@ -0,0 +1,67 @@
...
+
+modextrapaths = { 'IPM_KEYFILE': 'etc/ipm_key_mpi' }
+modextravars = { 
+    'IPM' : '-L%(installdir)s/lib -Wl,--whole-archive -lipmf -lipm -Wl,--no-whole-archive',
+    'IPM_HPM' : 'PAPI_TOT_CYC',
+}
+
+moduleclass = 'perf'
+
+#set             host       [lindex [split $env(APPS) "/"] 2]
```
gppezzi: delete or document the following comments

easybuild/easyconfigs/i/IPM/IPM-2.0.6-CrayIntel-18.07.eb
```diff
@@ -0,0 +1,67 @@
...
+
+modextrapaths = { 'IPM_KEYFILE': 'etc/ipm_key_mpi' }
+modextravars = { 
+    'IPM' : '-L%(installdir)s/lib -Wl,--whole-archive -lipmf -lipm -Wl,--no-whole-archive',
+    'IPM_HPM' : 'PAPI_TOT_CYC',
+}
+
+moduleclass = 'perf'
+
+#set             host       [lindex [split $env(APPS) "/"] 2]
```
gppezzi: delete or document the following comments

easybuild/easyconfigs/i/IPM/IPM-2.0.6-CrayPGI-18.07.eb
```diff
@@ -0,0 +1,67 @@
...
+
+modextrapaths = { 'IPM_KEYFILE': 'etc/ipm_key_mpi' }
+modextravars = { 
+    'IPM' : '-L%(installdir)s/lib -Wl,--whole-archive -lipmf -lipm -Wl,--no-whole-archive',
+    'IPM_HPM' : 'PAPI_TOT_CYC',
+}
+
+moduleclass = 'perf'
+
+#set             host       [lindex [split $env(APPS) "/"] 2]
```
gppezzi: delete or document the following comments

PR #674, opened 2018-07-12
[dom] Custom EasyBuild buildpath for each machine

lucamar:
Inserting a custom `EASYBUILD_BUILDPATH` for each machine, to allow defining the build path as  `/tmp/$USER/easybuild` on `Dom`.


jenkins/JenkinsfileProductionEB
```diff
@@ -3,6 +3,7 @@
 def methods
 def machinesList
 def loginBash = '#!/bin/bash -l'
+def buildPath = params.buildPath
```
teojgo: This one here is wrong, `buildPath` has to be defined in the machines loop.

PR #698, opened 2018-07-17
[dom] vampir/950

jgphpc:
license issue
will fix https://madra.cscs.ch/scs/doc/issues/346

### DIFF
--- a/easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
+++ b/easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
@@ -1,38 +1,69 @@
-# contributed by jg (CSCS)
-# easyblock = 'Tarball'
-easyblock = 'Binary'
-
+easyblock = 'Bundle'
 name = 'vampir'
 version = '9.5.0'
-shortversion = '95'
-
 homepage = 'http://www.vampir.eu'
 description = """ Vampir provides a manageable framework for analysis, which
 enables developers to quickly display program behavior at any level of detail.
 Detailed performance data obtained from a parallel program execution can be
-analyzed with a collection of different performance views: http://user.cscs.ch"""
+analyzed with a collection of different performance views.""" 
 
 toolchain = {'name': 'dummy', 'version': ''}
-sources = ['/apps/common/UES/easybuild/sources/v/Vampir/{}/{}-{}-linux-x86_64-setup.bin'.format(shortversion,name,version)]
-checksums = ['ebf62c854f1e9ad4a459f84f0737dd15']
-#sources = [{
-#  'filename': '/apps/common/UES/easybuild/sources/v/Vampir/%(shortversion)s/%(name)s-%(version)s-linux-x86_64-setup.bin', \
-#  'extract_cmd': "bash ./%s",
-#}]
-#postinstallcmds = ['echo "Server: velan 00000000 9000" > ' \
-#'%(installdir)s/conf/license/locallicense.dat']
-install_cmd = "bash " + sources[0] + " --silent --instdir=%(installdir)s "
-install_cmd += "--no-icon --no-menu-items --no-trace-file-association"
 
-sanity_check_paths = {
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
```diff
@@ -0,0 +1,69 @@
...
+description = """ Vampir provides a manageable framework for analysis, which
+enables developers to quickly display program behavior at any level of detail.
+Detailed performance data obtained from a parallel program execution can be
+analyzed with a collection of different performance views.""" 
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+modtclfooter = """
+# HOSTNAME ------------------------------------------------------
+set H        "[exec uname -n]"
```
victorusu: I do not see the use of these variables in the module.
And if you plan to use them, can you please use more descriptive variable name?

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
```diff
@@ -0,0 +1,69 @@
...
+# ---------------------------------------------------------------
+set     url                        "http://user.cscs.ch/compiling_and_optimizing/performance_tools/vi_hps/"
+set     url2                       "http://www.vampir.eu"
+set     packageroot                "/apps/common/UES/vampir/9.5.0"
+
+# ============================================================================
+# module-info mode ==> load, remove, display, help, whatis, switch, switch1, switch2, or switch3.
+# load
+# ============================================================================
+# if { [ module-info mode load ] } {
```
victorusu: You should not wrap the commands in a `[ module-info mode load ]`, because when unloading the modules, the variables won't be unset and the `$PATH` won't be updated.

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
```diff
@@ -0,0 +1,69 @@
...
+
+setenv          CSCS_VAMPIR     "$packageroot"
+prepend-path    PATH            "$packageroot/bin"
+# }
+
+# ============================================================================
+# Help message called by : module help ...
+# ============================================================================
+if { [ module-info mode help ] } {
```
victorusu: no need to wrap in a  `if { [ module-info mode help ] } {`. Modules will call the function only in the `help` mode.

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
```diff
@@ -0,0 +1,69 @@
...
+         \t      $packageroot/doc/"
+ 
+  return 0
+}
+}
+
+# ============================================================================
+# Short help message called by : module whatis ...
+# ============================================================================
+if { [ module-info mode whatis ] } {
```
victorusu: no need to wrap in a  `if { [ module-info mode whatis ] } {`. Modules will call the function only in the `whatis` mode.

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
jgphpc: Good point, thanks. We want to get rid of this modulefile.

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
jgphpc: Good point, thanks. We want to get rid of this modulefile.

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
jgphpc: Good point, thanks. We want to get rid of this modulefile.

easybuild/easyconfigs/v/vampir/vampir-9.5.0.eb
jgphpc: Good point, thanks. We want to get rid of this modulefile.

PR #713, opened 2018-07-20
[daint dom] ReFrame 2.14-dev0 easyconfig

vkarak:
We need this so as to freeze production runs to this version as we upgrade to the new PE.


lucamar:
@vkarak Do you really want to have in production on every system a `dev` release of `ReFrame`?

vkarak:
@lucamar Not necessarily in every system, but certainly for Daint and Dom. In any case, it's not the default. If you want, I can remove it from other systems.

vkarak:
@lucamar I have removed the dev release from all systems not required.

PR #714, opened 2018-07-23
[openpower] openpower

jgphpc:
```
ssh greina
ssh openpower01
```

* no os update since 03/2017 but nice to have


easybuild/easyconfigs/c/CMake/CMake-3.7.1.eb
```diff
@@ -0,0 +1,24 @@
...
+easyblock = 'ConfigureMake'
+
+name = 'CMake'
+version = "3.7.1"
+
+homepage = 'http://www.cmake.org'
+description = """CMake, the cross-platform, open-source build system.
+ CMake is a family of tools designed to build, test and package software."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
gppezzi: If you use version `dummy`, your dependencies might not actually be loaded. (it soulde be `'version': 'dummy'`)

There are other files that have the same issue in this PR.

jgphpc:
kept only xlcxlf recipes

PR #724, opened 2018-07-24
[daint-gpu] qe-6.3 

toxa81:
This EB script should work, but it doesn't (the similar build from the Daint shell works). The problem seems to come from the export variables like CPPFLAGS, CFLAGS, F90FLAGS that EB sets and they conflict with the way how QE configures itself.


jgphpc:
retest this please

PR #728, opened 2018-07-25
[dom-gpu] Tensorflow with CrayGNU-18.07

rsarm:
According to https://github.com/tensorflow/tensorflow/issues/20656, the newest versions of TensorFlow 1.8.0 and 1.9.0 can not be build with GCC-6.2.0.

I guess from this PR https://github.com/tensorflow/tensorflow/pull/20711 that the issue will be fixed soon.  In the meanwhile I update TF-1.7.0 to CrayGNU-18.07 with cuda-9.1.


gppezzi:
I'm fine with merging this and re-try building TF 1.8 (or later) when we're closer to the upgrade.

PR #729, opened 2018-07-25
[dom] Removing PyMySQL from PyExtensions

lucamar:
As agreed, I'm removing the extension `PyMySQL` from the list. For more details, see https://git.cscs.ch/scs/doc/issues/303


gppezzi:
How about removing pygelf from here and add it to a separate eb (together with pymysql?)

jgphpc:
@gppezzi where is pymysql.eb ?

gppezzi:
@jgphpc I don't know if we have it already, you should create a pull request for that if you don't find it. You could create a bundle of python extensions with pygelf and pymsql, similar to pyextensions, and install it as a hidden module.

We should avoiding mixing up software used internally at CSCS with scientific user modules.

PR #733, opened 2018-07-26
[dom] Addition of CDO recipe for CrayGNU-18.07 and CrayIntel-18.07

omlins:


omlins:
With CrayIntel, CDO v.1.9.4 failed to build; so I chose 1.9.3.

PR #748, opened 2018-07-31
[Leone] Changing MPI in LIGGGHTS/4.0

kraushm:
Test by Hilti showed, that the OpenMPI/1.10.2-GCC-5.3.0-2.26 gives better
performance than mvapich2/2.1-gcc-5.3.0. EB-file was adapted accordingly.

Also, LIGGGHTS 3.8 was removed from module list, since the installation
is no longer present

see  webRT ticket number 32590 (https://webrt.cscs.ch/Ticket/Display.html?id=32590)


kraushm:
retest this please

PR #755, opened 2018-08-09
[dom-gpu] Update cuda 9.1 module

victorusu:


gppezzi:
@victorusu could you please update the entry for the new `cudatoolkit` on the `cray_external_modules_metadata.cfg`

PR #756, opened 2018-08-09
[dom] rename dom files to match with OS version (UP07)

gppezzi:


lucamar:
I have created the temporary folder `/apps/dom/UES/6.0.UP07/login-dom` with the modulefiles, headers and footers. In order to use it in production, I just need to update the symbolic link below:
```
$ ll /apps/dom/UES/easybuild/modulefiles
lrwxrwxrwx 1 lucamar csstaff 32 Jul 11 11:53 /apps/dom/UES/easybuild/modulefiles -> /apps/dom/UES/6.0.UP06/login-dom/
```

PR #762, opened 2018-08-17
[daint dom leone] Update Machines.groovy

lucamar:
Commenting out `Fulen` until the system will be available.


teojgo:
@lucamar don't comment it since during the return statement it won't find fulen. Just remove fulen from the return list.

lucamar:
retest this please

PR #766, opened 2018-08-21
[daint-mc] New CPMD version

ajocksch:
newer intel compiler version
newer cray-mpich version

fixes a bug for one user but might introduce new bugs


easybuild/easyconfigs/c/CrayIntel/CrayIntel-17.12.eb
```diff
@@ -0,0 +1,22 @@
+easyblock = 'CrayToolchain'
+
+name = 'CrayIntel'
+version = '17.12'
+
+homepage = 'https://pubs.cray.com/discover'
+description = """Toolchain using Cray compiler wrapper, using PrgEnv-intel module (PE release: August 2017).\n"""
```
gppezzi: could you please update the description (`PE release: December 2017`)

gppezzi:
@lucamar We decided to create a new standard toolchain for intel 17.12, therefore it should be clear now what is the difference wrt the other installed version of CPMD.

easybuild/easyconfigs/c/CrayIntel/CrayIntel-17.12.eb
```diff
@@ -0,0 +1,22 @@
+easyblock = 'CrayToolchain'
+
+name = 'CrayIntel'
+version = '17.12'
+
+homepage = 'https://pubs.cray.com/discover'
+description = """Toolchain using Cray compiler wrapper, using PrgEnv-intel module (PE release: December 2017).\n"""
+
+toolchain = {'name': 'dummy', 'version': ''}
```
gppezzi: Apparently, the toolchain version needs to be `dummy` when defining toolchains.

PR #769, opened 2018-08-23
[dom] Update of CDO recipe for CrayGNU-18.0{7,8} and CrayIntel-18.0{7,8}

omlins:


omlins:
Hopefully the issues we have seen with the previous versions compiled with 18.07 will not appear anymore...

gppezzi:
@omlins please put on hold this PR until we have CrayGNU 18.08 available and then use 18.08 instead of 18.07.

omlins:
@gppezzi : I added therefore WIP to avoid merging...

easybuild/easyconfigs/c/CDO/CDO-1.9.5-CrayGNU-18.07.eb
```diff
@@ -0,0 +1,36 @@
...
+checksums = [ '0c60f2c94dc5c76421ecf363153a5043' ]
+
+dependencies = [
+    ('cray-netcdf/4.6.1.2', EXTERNAL_MODULE),
+    ('cray-hdf5/1.10.2.0', EXTERNAL_MODULE),
+    # CDO: OpenMP support for compute intensive operators: https://code.zmaw.de/projects/cdo/wiki/OpenMP_support
+    # No MPI support
+]
+
+preconfigopts = 'export NC_CONFIG=$EBROOTNETCDF/bin/nc-config && '
```
teojgo: Change to `preconfigopts = 'export NC_CONFIG=`which nc-config` && '` since it nc-config is not found during the configuration. That's why we had the problems with netcdf4 support.

easybuild/easyconfigs/c/CDO/CDO-1.9.5-CrayIntel-18.07.eb
```diff
@@ -0,0 +1,36 @@
...
+checksums = [ '0c60f2c94dc5c76421ecf363153a5043' ]
+
+dependencies = [
+    ('cray-netcdf/4.6.1.2', EXTERNAL_MODULE),
+    ('cray-hdf5/1.10.2.0', EXTERNAL_MODULE),
+    # CDO: OpenMP support for compute intensive operators: https://code.zmaw.de/projects/cdo/wiki/OpenMP_support
+    # No MPI support
+]
+
+preconfigopts = 'export NC_CONFIG=$EBROOTNETCDF/bin/nc-config && '
```
teojgo: Sane as above

omlins:
retest this please

easybuild/easyconfigs/c/CDO/CDO-1.9.5-CrayGNU-18.07.eb
gppezzi: If you look at the [metadata file](https://github.com/eth-cscs/production/blob/master/easybuild/cray_external_modules_metadata.cfg), you can see that we're defining the root variable to `NETCDF_DIR`, which contains only `lib` and `include` directories.

I would prefer that we use explicit paths when writing easyconfigs whenever possible, rather than using `which`. In this case you could use `$CRAY_NETCDF_DIR`

gppezzi:
@omlins the PE 18.08 is ready and merged, you can already use it instead of 18.07.

PR #781, opened 2018-08-27
WIP: [dom] Suggestion for standard way of ref. python version

kraushm:
Different ways of referencing the python version in a build occur in our EB-recipes:

1. hard-coded
2. different ways of variables
    - only define python-version
    - define py_maj/min/rev_ver and use them later directly in the recipe
    - define py_maj/min/rev_ver as well as pyver, pyshortver
    - ...

This is a proposition to standardize this, in order to facilitate efforts when a change has to be applied (e. g. from 18.07 -> 18.08)


gppezzi:
@kraushm how about creating a new 'CSCS EB best practices' wiki page [here](https://github.com/eth-cscs/production/wiki) and include this info?

kraushm:
@gppezzi A wiki page has been created [here](https://github.com/eth-cscs/production/wiki/UES-standards-(CUDA,-Python,-...)-for-EB-recipes).

gppezzi:
Thanks @kraushm !

PR #792, opened 2018-08-28
[dom-mc] Add GROMACS 2018.2 without CUDA

victorusu:


gppezzi:
retest this please

gppezzi:
are you planning to add this (and the GPU version) to the production list?

victorusu:
Yes, I am. But only after they are merged. I was planning to have a single PR for that. So, I do not have the production list out-dated if someone gets approved before.

PR #807, opened 2018-09-03
[fulen] Fix GDB 7.10.1 and add GBD 8.1.1

victorusu:
There is a missing patch for GDB 7.10.1

### DIFF
--- a/easybuild/easyconfigs/g/GDB/GDB-8.1.1.eb
+++ b/easybuild/easyconfigs/g/GDB/GDB-8.1.1.eb
@@ -1,0 +1,27 @@
+# jgp@cscs
+easyblock = 'ConfigureMake'
+
+name = 'GDB'
+version = '8.1.1'
+
+homepage = 'http://www.gnu.org/software/gdb/gdb.html'
+description = "The GNU Project Debugger"
+
+# http://mirror.switch.ch/ftp/mirror/gnu/gdb/gdb-7.10.1.tar.gz
+source_urls = ['http://mirror.switch.ch/ftp/mirror/gnu/gdb/']
+sources = [SOURCELOWER_TAR_XZ]
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+dependencies = [('ncurses', '6.0')]
+
+patches = ['GDB-%(version)s-missing-makeinfo.patch']
+
+parallel = 1
+
+sanity_check_paths = {
+    'files': ['bin/gdb', 'bin/gdbserver'],
+    'dirs': [],
+}
+
+moduleclass = 'debugger'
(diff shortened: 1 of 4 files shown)


victorusu:
retest this please

easybuild/easyconfigs/g/GDB/GDB-8.1.1.eb
```diff
@@ -0,0 +1,27 @@
...
+version = '8.1.1'
+
+homepage = 'http://www.gnu.org/software/gdb/gdb.html'
+description = "The GNU Project Debugger"
+
+# http://mirror.switch.ch/ftp/mirror/gnu/gdb/gdb-7.10.1.tar.gz
+source_urls = ['http://mirror.switch.ch/ftp/mirror/gnu/gdb/']
+sources = [SOURCELOWER_TAR_XZ]
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
gppezzi: version should be empty or dependencies are not loaded

victorusu: I made it `dummy`, because I think GDB should not be a dependency ever.
But I can change it to ''.

gppezzi: Let me rephrase that: the tc version should be empty (i'm not referring to the name) or the dependencies (inside this eb file) are not loaded (unless eb has recently changed this behaviour recently, this recipe is using ncurses from the system and not from eb).

victorusu: I think that setting `'version': 'dummy'` affects the use of the `GBD` as a dependency, but not the use of dependencies within `GDB`. Because I can see the `module load ncurses` command inside the generated `GDB` module
```
module show GDB/8.1.1 
-------------------------------------------------------------------
/apps/.../GDB/8.1.1:

module-whatis	 Description: The GNU Project Debugger 
module-whatis	 Homepage: http://www.gnu.org/software/gdb/gdb.html 
conflict	 GDB 
module		 load ncurses/.6.0 
...
```

gppezzi: Did you check also during the load occurs during the build process? 

EB not loading the dependencies when the toolchain version is defined is a known issue and we had to chang many recipes because of this. 

If this behaviour changed, we should confirm (ideally asking @boegel and identifying the PR that touched that) and then inform all the team.

boegel: Using `dummy` as version means that the dependencies won't be loaded during the build, but the will be included in the generated module file, see also https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#loading-of-modules-for-dependencies-with-a-dummy-toolchain .

Long story short: always use `''` as version when using `dummy` toolchain. The behavior with the `dummy` version is a historical, which we intend to rectify with the `system` toolchain concept which is supposed to replace `dummy`, see https://github.com/easybuilders/easybuild-framework/issues/1181 (it'll happen some day, will put this back on the agenda for EB v4.0 (see https://github.com/easybuilders/easybuild/issues/447)).

PR #809, opened 2018-09-03
[kesch] netcdf+pgi

jgphpc:
https://madra.cscs.ch/scs/doc/issues/372

PR #819, opened 2018-09-12
[kesch] Kesch+pgi

jgphpc:
for https://github.com/eth-cscs/production/pull/809

### DIFF
--- a/easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
+++ b/easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
@@ -12,7 +12,7 @@ description = "C, C++ and Fortran compilers from The Portland Group - PGI"
 toolchain = {'name': 'dummy', 'version': ''}
 
 dependencies = [
-    ('gcc', '5.4.0-2.26'),
+    ('GCC', '5.4.0-2.26'),
     ('pgi/17.10', EXTERNAL_MODULE),
 ]   
 
--- a/easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
+++ b/easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
@@ -12,7 +12,7 @@ description = "C, C++ and Fortran compilers from The Portland Group - PGI"
 toolchain = {'name': 'dummy', 'version': ''}
 
 dependencies = [
-    ('gcc/5.4.0-2.26', EXTERNAL_MODULE),
+    ('gcc', '5.4.0-2.26'),
     ('pgi/17.10', EXTERNAL_MODULE),
 ]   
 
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
```diff
@@ -0,0 +1,19 @@
...
+binutilsver = '2.26'
+versionsuffix = '-GCC-%s-%s' % (gccver, binutilsver)
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+dependencies = [
+    ('gcc/5.4.0-2.26', EXTERNAL_MODULE),
```
kraushm: Would it make it more easy to maintain this, if the variables defined above were also used for the dependencies?

gppezzi: this module is provided by EB, why are you using as external module?

gppezzi:
I suggest to include this file in the same PR as netcdf. 

If this is only used by netcdf (as dependency), you don't need to add it to the production list, only adding netcdf should be enough.

easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
jgphpc: this is what i get if i don't put it as external:
```
[kesch] ERROR: Irresolvable dependencies encountered: gcc/5.4.0-2.26
```

`/apps/escha/UES/PrgEnv-gnu-17.02/modulefiles/gcc/5.4.0-2.26`

gppezzi: GCC is capital in EB.

MCH asked to have a lowercase naming scheme so the module names are misleading, we need always to look at the eb files.

easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
```diff
@@ -0,0 +1,19 @@
...
+binutilsver = '2.26'
+versionsuffix = '-GCC-%s-%s' % (gccver, binutilsver)
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+dependencies = [
+    ('GCC', '5.4.0-2.26'),
```
kraushm: I would still vote for using the variables `gccver` and `binutilsver` also for the gcc dependency.

easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
```diff
@@ -0,0 +1,19 @@
...
+    ('pgi/17.10', EXTERNAL_MODULE),
```
kraushm: Same argument, only for the variable `version`.

easybuild/easyconfigs/p/pgi/pgi-17.10-gcc-5.4.0-2.26.eb
```diff
@@ -0,0 +1,19 @@
+easyblock = 'Bundle'
+
+name = 'PGI'
```
gppezzi: The eb file name here does not match the case of the package name , so you might have problems when using this eb as dependency (one more reason for having this file in the same pr as netcdf).

PR #820, opened 2018-09-13
[dom] Addition of julia 1.0.0 recipe for dummy-toolchain

omlins:


easybuild/easyconfigs/j/julia/julia-1.0.eb
```diff
@@ -0,0 +1,20 @@
+# Recipe for linux, x86_64 created by Samuel Omlin (CSCS), September 12, 2018
+easyblock = 'PackedBinary'
+
+name = 'julia'
+version = '1.0.0'
```
gppezzi: the file name does not match the version number

omlins: good catch!

PR #822, opened 2018-09-13
WIP [leone] trying to update Leone to PV5.5

jfavre:
The Qt part is OK.

The ParaView compilation, actually the **linking** fails, with the error

collect2: fatal error: cannot find 'ld'


jfavre:
We have v5.6 successfully installed, so let's forget about v5.5.

PR #824, opened 2018-09-14
[monch] dismissed

jgphpc:


jgphpc:
I would remove too:

common-RH6.7-2016.10 
common-SLES12-2016.10

gppezzi:
if you want to do some cleaning, why not removing all the old files out of the way?

also `6.0.UP02-2016.11-mc`, `6.0.UP04-17.08-gpu` and `MCH-RH6.7-15.11`?

PR #834, opened 2018-09-20
[daint-gpu] Add GROMACS 2018.3

victorusu:
The reason why I am also changing the default is because the current
GROMACS 2018 has 5 major bugs fixed using version 2018.3

Especially the is a Replica Exchange bug that hinders users from using
RE with GROMACS 2018.


victorusu:
retest this please

PR #846, opened 2018-09-25
[daint] ParaView-5.6.0

jfavre:
Release Candidate 2.

Progress at:
https://gitlab.kitware.com/paraview/paraview/milestones/8


jfavre:
both EGL and OSMesa build logs:

/scratch/snx1600tds/jfavre/daint/software/ParaView/5.6.0-CrayGNU-18.08-*/easybuild/easybuild-ParaView-5.6.0*log

jfavre:
retest this please

jfavre:
@victorusu To summarize our discussion about you asking and saying "we do not know how these libraries are built". These are not actually libraries, they are reader plugins for specific CSCS projects. Reproducibility is fully insured because the options provided to EB and cmake are directory names where the source code resides. ParaView is build by including that source code into the main code. It is optional, and non-invasive. Being provided as VTK source code, it is simple to deploy and maintain.

https://www.paraview.org/Wiki/ParaView/Plugin_HowTo

documents how plugins are build. See paragraph "Adding plugins to ParaView source". Following that documentation, indeed, each of the Directory names we supply to ParaView contains a file called plugin.cmake which triggers the source code inclusion and compilation.

HIH

PR #848, opened 2018-09-25
[dom-gpu] clang700

jgphpc:


jgphpc:
test never started, retest this please

easybuild/easyconfigs/l/LLVM/LLVM-7.0.0-CrayGNU-18.08.eb
```diff
@@ -0,0 +1,48 @@
...
+is particularly easy to invent your own language (or port an existing compiler)
+to use LLVM as an optimizer and code generator."""
+
+toolchain = {'name': 'CrayGNU', 'version': '18.08'}
+toolchainopts = {'dynamic': True}
+
+source_urls = ["http://llvm.org/releases/%(version)s"]
+sources = ["llvm-%(version)s.src.tar.xz"]
+
+#python = 'python'
```
victorusu: Do you need to keep all these comments?

easybuild/easyconfigs/l/LLVM/LLVM-7.0.0-CrayGNU-18.08.eb
```diff
@@ -0,0 +1,48 @@
...
+popular CPUs (as well as some less common ones!) These libraries are built
+around a well specified code representation known as the LLVM intermediate
+representation ("LLVM IR"). The LLVM Core libraries are well documented, and it
+is particularly easy to invent your own language (or port an existing compiler)
+to use LLVM as an optimizer and code generator."""
+
+toolchain = {'name': 'CrayGNU', 'version': '18.08'}
+toolchainopts = {'dynamic': True}
+
+source_urls = ["http://llvm.org/releases/%(version)s"]
```
victorusu: Can you share the list of installed packages. The documentation says:
> These binaries include Clang, LLD, compiler-rt, various LLVM tools, etc. varying slightly between platforms.

Do you know the complete list?

For instance does it have OpenMP support?
Does it use the gold linker?
Does it use Polly?

jgphpc: I don't have such a list but the binaries come from:
http://releases.llvm.org/7.0.0/clang+llvm-7.0.0-x86_64-linux-sles12.3.tar.xz

In the installed dir, there is lib/libomp.so and LLVMPolly.so.

> ld.lld -V
LLD 7.0.0 (compatible with GNU linkers)

For the rest, i don't know. I would suggest to start with what is available and make another PR if we need more.

easybuild/easyconfigs/l/LLVM/LLVM-7.0.0-CrayGNU-18.08.eb
jgphpc: 97d69172156481cffecfaa0476aab0053222e7f0

PR #857, opened 2018-10-03
[dom-gpu] googletest

jgphpc:


jgphpc:
retest this please

jgphpc:
retest this please

easybuild/easyconfigs/g/googletest/googletest-1.8.1-CrayGNU-18.08.eb
```diff
@@ -0,0 +1,33 @@
...
+sources = ['release-%(version)s.tar.gz']
+checksums = ['2e6fbeb6a91310a16efe181886c59596']
+
+builddependencies = [
+    ('CMake', '3.12.0', '', True),
+]
+
+separate_build_dir = True
+
+configopts  = '-DCMAKE_CXX_COMPILER=`which g++` -DCMAKE_C_COMPILER=`which gcc` '
```
victorusu: What about using the cray compiler wrappers?

jgphpc: in this special case, i also want to be able to run on the login nodes too but if i compile with the cray wrappers, i won't be able to.

victorusu: Ok, then! 👍

PR #860, opened 2018-10-03
[dom-gpu] QuantumESPRESSO 6.3 and libxc with CrayIntel

lucamar:
Recipe for `QuantumESPRESSO 6.3` and `libxc 4.2.3` built using the toolchain `CrayIntel-18.08` (linear algebra routines are linked directly with `MKL`).


toxa81:
why I don't see make.inc in the output?

lucamar:
@toxa81 
```
grep -A 189 "# make.inc." /scratch/snx1600tds/jenscscs/TestingEB/dom-gpu/37113a6-1029/software/QuantumESPRESSO/6.3-CrayIntel-18.08/easybuild/easybuild-QuantumESPRESSO-6.3-20181004.094551.log
```

toxa81:
Try another configuration; the idea is to have -D__DFTI in the make.inc and -O3 in the FFLAGS

toxa81:
-D__DFTI is there, but scalapack and FFLAGS are not

toxa81:
This looks fine now. All f90 files are compiled with:

ftn -O3 -g -assume byterecl -traceback -qopenmp -fpp -D__DFTI -D__MPI -D__SCALAPACK

lucamar:
Thanks @toxa81 : I have cleaned up a bit the recipe, in order to remove the lines that you have commented out and also delete unnecessary lines. I have modified the following variables:
- `FFT_LIBS= -mkl` is needed only to let `configure` define `-D__DFTI` in the `DFLAGS`
- `LDFLAGS` does not seem to need `$LIBS`, as the link lines in the output are already taking `$LIBS`, which includes the libraries previously defined in `make.inc`

PR #863, opened 2018-10-05
[dom-gpu] Improve the compilation flags setup

victorusu:
This PR attempts to decrease the confusion created by adding the 'cojnfigopts' in order to identify which compilation flags are used by GROMACS. Now, we let EB set the variables.

There should be a small improvement of the CPU side of the code.


victorusu:
retest this please

PR #871, opened 2018-10-08
[dom-gpu] Adjusting EASYBUILD_BUILDPATH

lucamar:
We set a new buildpath by default under `XDG_RUNTIME_DIR`.

### DIFF
--- a/easybuild/module/Easybuild
+++ b/easybuild/module/Easybuild
@@ -87,13 +87,13 @@ if { ! [ info exists ::env(EASYBUILD_PREFIX) ] } {
 }
 setenv EASYBUILD_INSTALLPATH                    $::env(EASYBUILD_PREFIX)
 #
-# If not set, EASYBUILD_BUILDPATH and EASYBUILD_TMPDIR will point to /dev/shm/$USER/easbuild/stage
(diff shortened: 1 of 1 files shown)


easybuild/module/Easybuild
```diff
@@ -87,13 +87,13 @@ if { ! [ info exists ::env(EASYBUILD_PREFIX) ] } {
 }
 setenv EASYBUILD_INSTALLPATH                    $::env(EASYBUILD_PREFIX)
 #
-# If not set, EASYBUILD_BUILDPATH and EASYBUILD_TMPDIR will point to /dev/shm/$USER/easbuild/stage
+# If not set, EASYBUILD_BUILDPATH and EASYBUILD_TMPDIR will point to $XDG_RUNTIME_DIR/easybuild
```
victorusu: Wouldn't it be better to point to `/tmp`?
Because currently the `$XDG_RUNTIME_DIR` point to a subfolder of `/apps` which is readonly on the compute nodes.

victorusu: Sorry @lucamar... I have confused `$XDG_RUNTIME_DIR` with `$XDG_CONFIG_DIRS`...
All looks great! Thanks @lucamar!

PR #872, opened 2018-10-08
[dom-gpu] Generic download link compatible with previous releases

lucamar:
I'm setting in the EasyBuild recipe a different download link from `QuantumESPRESSO` GitHub project that is compatible with older releases: in this way, the EasyBuild flag `--try-software-version=...` would also download automatically with previous releases.


lucamar:
retest this please

PR #883, opened 2018-10-12
[daint-gpu] Horovod

rsarm:


mschoengens:
Shall we make Horovod also available on the MC partition and Dom?

mschoengens:
Aha I see. I would actually remove the dependency to TF since
 - Horovod might also be used with TF without Cuda support on the MC partition, and 
 - it also works with other frameworks such as pytorch

rsarm:
I tried to do that but I think Horovod needs some info from TensorFlow (and from PyTorch if `$HOROVOD_WITHOUT_PYTORCH=0`). For instance, during building, Horovod's [setup.py](https://github.com/uber/horovod/blob/master/setup.py) file imports TensorFlow to get info like  TensorFlow's version, libs, includes, etc.

Maybe we can split the module on versions, one for TensorFlow and one for PyTorch, if needed. What do you think?

mschoengens:
And this isn't just a build dependency? I just wonder since the example for [pytorch](https://github.com/uber/horovod/blob/master/examples/pytorch_mnist.py) does not import TF.

rsarm:
Ah right! It's better as a build dependency. Changing it.

PR #893, opened 2018-10-19
[daint-gpu dom-gpu] reverting magma and CP2K to use only cudatoolkit 

gppezzi:
(instead of craype-acel-nvidia60)

Builds were failing (most probably because of the variable `CRAY_TCMALLOC_MEMFS_FORCE=1`, but the issue needs to be confirmed)

### DIFF
--- a/easybuild/easyconfigs/c/CP2K/CP2K-6.1-CrayGNU-18.08-cuda-9.1.eb
+++ b/easybuild/easyconfigs/c/CP2K/CP2K-6.1-CrayGNU-18.08-cuda-9.1.eb
@@ -26,6 +26,7 @@ sources = [SOURCELOWER_TAR_BZ2]
 source_urls = [SOURCEFORGE_SOURCE]
 
 builddependencies = [
+    ('cudatoolkit/%s.85_3.18-6.0.7.0_5.1__g2eb7c52' % cudaversion, EXTERNAL_MODULE),
     ('cray-fftw/3.3.6.5', EXTERNAL_MODULE),
     ('cray-python/%s' % pyver, EXTERNAL_MODULE),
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/c/CP2K/CP2K-6.1-CrayGNU-18.08-cuda-9.1.eb
```diff
@@ -6,12 +6,6 @@ version = '6.1'
 cudaversion = '9.1'
 versionsuffix = '-cuda-%s' % cudaversion
 
-py_maj_ver = '2'
```
lucamar: We need to restore the `Python` variables in order to comply with the standard template described in the Wiki: I'll push a new version of the recipe for `CP2K`

lucamar:
Thanks @gppezzi: the variable `CRAY_TCMALLOC_MEMFS_FORCE=1` itself does not create problems during the build and at runtime. However `craype-accel-nvidia60` loads as dependency `cray-libsci_acc`: if I unload the latter during the build, then at runtime `CP2K` runs correctly.
See a short summary here: `/scratch/snx3000/lucamar/cp2k/benchmark/legend.txt`
My tests show that having `cray-libsci_acc` only is not enough to reproduce the error, though.

Therefore I agree that for the time being it is safer to remove `craype-accel-nvidia60` from the dependencies, untile the issue gets clarified with the help of Cray experts (I will open a case).

gppezzi:
I'm pasting the [benchmark results](https://gist.github.com/gppezzi/7c9fdf02e2f90c3b0580461cd502a654)  in a gist just for completeness

lucamar:
retest this please

PR #905, opened 2018-10-30
[daint] Adding Dycore and STELLA to production list

kraushm:
COSMO recipes include Dycore and STELLA as EXTERNAL_MODULE, since
(based on the recommendation by MCH) both dependencies are built
for the CrayGNU toolchain, whereas COSMO is built for CrayCCE.


kraushm:
After internal discussions this approach has been rejected.

PR #908, opened 2018-10-31
[daint]  ADIOS recipe with dependencies for CrayIntel-18.08

omlins:
This ADIOS recipe features support for many different compression algorithms including those used by the PRACE group of HZDR (blosc and zlib). Moreover, it is optimized for Cray and Lustre.


omlins:
retest this please

PR #909, opened 2018-11-05
[dom-gpu daint-gpu] added module cudatoolkit to magma recipe

ajocksch:
@toxa81 why do we use magma 2.2 with GNU and magma 2.4 with Intel ? Is this inconsistency wanted?

The pull request will avoid a separate loading of cudatoolkit.


toxa81:
From personal experience: MAGMA works better with Intal+MKL stack. I didn't try to test GNU-compiled MAGMA with LibSci

PR #913, opened 2018-11-06
[leone] zero length field name in format

jgphpc:
On leone, python default/system version is 2.6.6, which will give:
`ValueError: zero length field name in format`
Using python>=2.7 fixes the issue.


jgphpc:
retest this please

kraushm:
Using the Python module as a dependency was not possible?

jgphpc:
next time

PR #916, opened 2018-11-07
[daint] Add CMake 3.12.4

havogt:
This patch version 3.12.4 fixes a bug concerning MPI and CUDA


lucamar:
ok to test

lucamar:
@havogt Do you need the new build of `CMake` on Piz Daint only?

havogt:
We will use it on daint and kesch (and tave, which we anyway need to manage ourselves, right?). But we have also jenkins plan on the MCH to install it for us. So I'd say it's up to you where you want to deploy it officially.

PR #929, opened 2018-11-26
[daint-gpu] Lammps with Plumed

lucamar:
The recipe will build `LAMMPS` patched with `PLUMED`.


lucamar:
Started a new pull request from a clean branch.

PR #933, opened 2018-11-27
[daint] Update of ADIOS and dependencies

omlins:
The update enhances the ADIOS module with proper integration to the CRAY driver, such that the wrappers cc, CC and ftn will automatically use the correct linking commands. 
To this purpose the easybuild files sets the CRAY specific environment variables and notes the linking dependencies in the pkgconfig files of ADIOS and its dependencies.

### DIFF
--- a/easybuild/easyconfigs/a/adios/adios-1.13.1-CrayIntel-18.08-nompi.eb
+++ b/easybuild/easyconfigs/a/adios/adios-1.13.1-CrayIntel-18.08-nompi.eb
@@ -3,6 +3,7 @@ easyblock = 'ConfigureMake'
 
 name = "adios"
 version = '1.13.1'
+versionsuffix = '-nompi'
 
 homepage = 'https://www.olcf.ornl.gov/center-projects/adios/'
 description = """The Adaptable IO System (ADIOS) provides a simple, flexible way for scientists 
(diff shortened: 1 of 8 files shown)


omlins:
I noted that there is a problem with the module. When I load it, then it works all fine; however, when I unload it and load it again, then I get an error:
```
omlins@daint107:~/10_io_perf/0_adios_helloworld/c> module load adios/1.13.1-CrayIntel-18.08 
adios/1.13.1-CrayIntel-18.08(71):ERROR:102: Tcl command execution failed: setenv	PE_PKGCONFIG_PRODUCTS		"ADIOS:$::env(PE_PKGCONFIG_PRODUCTS)
```
I suppose the module unsets all the environment variables created with setenv, when it is unloaded. Thus, PE_PKGCONFIG_PRODUCTS becomes unset. Therefore I should be able to use prepend path as e.g. in the netcdf module. I will probably have to call tcl directly...

omlins:
The solution with modtclfooter works. The modules can now be properly unloaded and reloaded.

omlins:
retest this please

omlins:
After discussion with @gppezzi and @lucamar, we decided not to do any changes to the dependencies of ADIOS, as they are already in production (in order to avoid retesting of many software). I restored therefore for all dependencies the version on the master. The original branch with all the properly modified dependencies is saved here: 
    https://github.com/eth-cscs/production/tree/ADIOS_Daint_Upgrade_2019
It will have to be integrated at the next Daint upgrade in 2019.

omlins:
Now, I will modify the ADIOS recipes - hardcoding the linking  to avoid that dependencies need to be adapted.

omlins:
The test applications (c and fortran) could be built with the Cray wrappers (cc / ftn) and produced the correct output when run. I think it it ready to be merged.

PR #937, opened 2018-12-04
[daint-gpu,dom-gpu]  Update 6.0.UP07-18.08-gpu

twrobinson:
Add dask (with dask.distributed) for eventual integration in jupyterhub


twrobinson:
retest this please

gppezzi:
As discussed with @rsarm, we can merge this but we could do a symlink to the existing hidden module in order to avoid re-building the existing software.

PR #942, opened 2018-12-05
[dom,daint] Create ipyparallel-6.2.3-CrayGNU-18.08-python3.eb

twrobinson:
ipyparallel when a user requests MPI to be set up automatically in their notebook from the jupyterhub spawner page (development).


twrobinson:
ready for merging

PR #960, opened 2019-01-14
[daint-gpu] update QE+SIRIUS to the latest version

toxa81:
I'm changing a name convention for the QE eb script because I sync with QE's develop branch which is always v6.3


gppezzi:
@toxa81 can you please update the subject to match the target system and the software name?

toxa81:
test this please

PR #962, opened 2019-01-15
[daint-gpu] ParaView enable NVIDIA IndeX for all users

jfavre:
All users are enabled to use NVIDIA IndeX. We provide the license file in /apps/common/UES/easybuild/sources/p/ParaView. It must be called "nvindex_config.xml" and we provide a default lib path in the same directory.

we updated $LD_LIBRARY_PATH
we defined $NVINDEX_PVPLUGIN_HOME


jfavre:
== COMPLETED: Installation ended successfully
== Results of the build can be found in the log file(s) /scratch/snx3000/jfavre/daint/software/ParaView/5.6.0-CrayGNU-18.08-EGL/easybuild/easybuild-ParaView-5.6.0-20190115.163140.log


tested in client-server mode with two nodes. If anyone asks....

Users should always limit themselves to a single pvserver task per node. I.e. the GPU should not be shared amongst multiple pvserver tasks on a given node

jfavre:
retest this please

PR #966, opened 2019-01-28
[kesch] NCL 6.5.0 recipe for kesch

omlins:


omlins:
This is asked for in the following ticket: [34747](https://webrt.cscs.ch/Ticket/Display.html?id=34747).

PR #975, opened 2019-02-05
[dom-gpu] syntax_tools

jgphpc:


easybuild/easyconfigs/c/cpplint/cpplint-6271f3f.eb
```diff
@@ -0,0 +1,21 @@
...
+version = '6271f3f'
+
+homepage = 'https://github.com/google/styleguide/tree/gh-pages/cpplint'
+description = """This is automated checker to make sure a C++ file follows
+Google's C++ style guide (https://google.github.io/styleguide/cppguide.html)"""
+
+toolchain = {'name': 'dummy', 'version': ''}
+source_urls = ['https://github.com/google/styleguide/archive']
+sources = ['gh-pages.zip']
+install_cmd = 'mkdir -p %(installdir)s/bin && ' + \
```
kraushm: If I understand correctly, this install_cmd could be omitted when using the easyblock `BinariesTarball`.

(see https://easybuild.readthedocs.io/en/latest/version-specific/generic_easyblocks.html#binariestarball)

jgphpc: I get your point but in this particular case, i want only 1 file from the repo (`styleguide-gh-pages/cpplint/cpplint.py`), nothing else,
see `unzip -l /apps/common/UES/easybuild/sources/c/cpplint/gh-pages.zip `

kraushm: Check...didn't see that there is, in fact, no bin-folder to copy!?

PR #976, opened 2019-02-07
[dom daint] Make installpaths use 'jenkins'

teojgo:
* Replace `jenscscs` with `jenkins` at the installpaths
  of the production lists for dom/daint.


jgphpc:
This link:
```
lrwxrwxrwx 1 karakasv csstaff 8 Jul 21  2017 /apps/dom/UES/jenscscs -> jenkins/
```
has been in place since the beginning of [daint-gpu](https://github.com/eth-cscs/production/commits/master/login/daint-gpu).

Can you give more details to why this is needed (for jenkins) ?

jgphpc:
It would make sense to replace `/apps/daint` with `$APPS`

teojgo:
> It would make sense to replace `/apps/daint` with `$APPS`

Should it make sense in the `installpath` or in the jenkins production script?

jgphpc:
This pr is justified because `/apps/*/jenkins` ==  `/apps/*/jenscscs` but more importantly `/apps/daint` != `/apps/dom`

PR #986, opened 2019-02-18
[dom] Update production.sh

victorusu:
Updated the `export XALTPROD=/apps/daint/UES/xalt/production` path, but I think the issue is related to the permissions on `export XALTJENKINS=/apps/daint/UES/xalt/JENSCSCS`. They should be owned by the Jenkins' user in our systems.


lucamar:
retest this please

PR #991, opened 2019-02-26
[daint-gpu] Update the GROMACS install

victorusu:
There is no need for the gromacs.py file, since it is added on the EB
3.8.1.

Moreover, we are adding a newer version of GROMACS 2018.4 which contains
some bug fixed and is using a newer CMake version


easybuild/easyconfigs/g/GROMACS/GROMACS-2018.4-CrayGNU-18.08-cuda-9.1.eb
```diff
@@ -0,0 +1,46 @@
...
+patches = ['gromacs-2018.2.patch']
+
+configopts = '-DGMX_CYCLE_SUBCOUNTERS=ON -DGMX_BUILD_OWN_FFTW=ON'
+
+skipsteps = ['test']
+
+# CMake dependency with dummy toolchain
+builddependencies = [
+    ('CMake', '3.13.3', '', True),
+    #('cray-fftw/3.3.6.5', EXTERNAL_MODULE),
```
gppezzi: why do you keep fftw and libxml2 commented out?

lucamar: Thanks @victorusu

PR #992, opened 2019-03-01
[daint dom] Cosmo_pompa

kraushm:
for MeteoSwissAPN (branch: baroclinic)


omlins:
@gppezzi : The Dycore of this branch is not c++11/c++14 standard compatible; this is why it failed first with GCC 6.2.0, where the c++14 standard is the default, but not with GCC 5.3.0, where the c++98 standard (with GNU extensions) is the default... So, using `-std=gnu++98` made it compile with GCC 6.2.0 (`-std=c++98` does not work as GNU extensions are used).

kraushm:
The recipes for the different COSMO builds have been tested manually on Piz Daint. (jenkins is not able to resolve the "EXTERNAL_MODULE" dependencies)
The outputs can be found here:

https://gist.github.com/kraushm/12f42ddff722f4ef421e19f2796a5763 (Double precision)
https://gist.github.com/kraushm/addf9c751e868dca59d8833be00639d1 (Single precision)
https://gist.github.com/kraushm/c08ddb178e4d3210657fae1db808d14e (Single - CUDA)
https://gist.github.com/kraushm/41a8568bced3b4527cbc451a3074e429 (Double - CUDA)

PR #995, opened 2019-03-07
[daint dom fulen kesch leone] Update EasyBuild-custom modulefile and remove easyconfig

lucamar:
Adding `$::env(USER)` in the buildpath and tmpdir of systems that do not define `XDG_RUNTIME_DIR`. 

I have also removed the recipe `EasyBuild-custom/EasyBuild-custom-cscs.eb` as it is outdated and it should not be used to create the `EasyBuild-custom/cscs` modulefile anymore, since `production.sh`  links the hard-coded `EasyBuild-custom/cscs` module.


lucamar:
For `fulen` I had to set `XDG_RUNTIME_DIR` manually, as it is not defined on the system:
```
    setenv XDG_RUNTIME_DIR                      /dev/shm/$::env(USER)
```
In principle we could use this approach on `leone` as well, avoiding to define a separate `if` later in the modulefile. I could not set it to `/tmp/$::env(USER)`, since `/tmp` has limited space available on `fulen`; I could not use `/run/user` either, since users are not allowed to create folders there:
```
$ mkdir /run/user/$USER
mkdir: cannot create directory ‘/run/user/lucamar’: Permission denied
```

lucamar:
@victorusu Do we still need to define `XDG_CONFIG_DIRS`?
```
setenv XDG_CONFIG_DIRS                          $::env(EB_CUSTOM_REPOSITORY)
```

lucamar:
retest this please

victorusu:
> @victorusu Do we still need to define `XDG_CONFIG_DIRS`?
> 
> ```
> setenv XDG_CONFIG_DIRS                          $::env(EB_CUSTOM_REPOSITORY)
> ```

@lucamar, I do not think so.

victorusu:
> > @victorusu Do we still need to define `XDG_CONFIG_DIRS`?
> > ```
> > setenv XDG_CONFIG_DIRS                          $::env(EB_CUSTOM_REPOSITORY)
> > ```
> 
> @lucamar, I do not think so.

@lucamar, Yes! We still need to define it!

PR #996, opened 2019-03-07
[dom] Add CrayGNU 19.03

victorusu:


easybuild/easyconfigs/c/CrayGNU/CrayGNU-19.03.eb
```diff
@@ -0,0 +1,31 @@
...
+(PE release: March 2019).\n"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+dependencies = [
+    # PrgEnv version is not pinned, as Cray recommends to use the latest (default) version
+    ('PrgEnv-gnu', EXTERNAL_MODULE),
+    ('atp/2.1.3', EXTERNAL_MODULE),
+# gcc versions later than 6 are not supported by cuda9
+    ('gcc/6.1.0', EXTERNAL_MODULE), # there is not gcc 6.2.0 anymore
```
gppezzi: isn't gcc 7.3 compatible with cuda10? (see https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html)

easybuild/easyconfigs/c/CrayGNU/CrayGNU-19.03.eb
```diff
@@ -0,0 +1,31 @@
...
+    ('cce/8.7.9', EXTERNAL_MODULE),
```
gppezzi: why `cce` should be part of CrayGNU?

easybuild/easyconfigs/c/CrayGNU/CrayGNU-19.03.eb
victorusu: 👍

easybuild/easyconfigs/c/CrayGNU/CrayGNU-19.03.eb
victorusu: Mistake!

victorusu:
> Thanks @victorusu: should we leave the request as `WIP` for the time being?

@lucamar, I would like to start installing stuff using this CDT...

PR #997, opened 2019-03-12
[daint] Add GREASY version 19.03-cscs

victorusu:
This is a very disruptive PR for two reasons:

1. GREASY is currently broken, so the module makes no sense to exist
2. The new GREASY has a different usage syntax that is similar to the original GREASY, but on steroids.

So, I am proposing to remove the current (old/broken) GREASY from the production list and update the documentation accordingly.

### DIFF
--- a/easybuild/easyconfigs/g/GREASY/GREASY-CSCS-19.03-CrayGNU-18.08.eb
+++ b/easybuild/easyconfigs/g/GREASY/GREASY-CSCS-19.03-CrayGNU-18.08.eb
@@ -1,0 +1,28 @@
+easyblock = 'ConfigureMake'
+
+name = 'GREASY'
+version = 'CSCS-19.03'
+
+homepage = 'https://github.com/jonarbo/GREASY'
+description = """Greasy is a tool designed to make easier the deployment of
+embarrassingly parallel simulations in any environment. It is able to run in
+parallel a list of different tasks."""
+
+toolchain = {'name': 'CrayGNU', 'version': '18.08'}
+toolchainopts = {'opt': True, 'pic': True}
+
+#
+# GREASY source code modified by us in order to properly use SLURM scheduler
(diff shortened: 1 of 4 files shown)


easybuild/easyconfigs/g/GREASY/GREASY-CSCS-19.03-CrayGNU-18.08.eb
```diff
@@ -0,0 +1,28 @@
+easyblock = 'ConfigureMake'
+
+name = 'GREASY'
+version = 'CSCS-19.03'
```
vkarak: Wouldn't it make sense for version to be `19.03_cscs`? I'm not an expert in EB though.

vkarak:
I think the procedure for doing this change is the following:

1. Merge this PR and install the new module.
2. Update documentation in the user portal.
3. Change the current module to print a big deprecation warning and point the users to use the new one. The message should explicitly state that the command-line syntax has changed and should point the users to the documentation page.

easybuild/easyconfigs/g/GREASY/GREASY-CSCS-19.03-CrayGNU-18.08.eb
victorusu: This is the release version I did on the repo...

gppezzi: This is not an EB thing, the version here is matching the release version in the github repo: https://github.com/eth-cscs/GREASY/releases

If you are not happy with the convention, I think we should first fix that with a new release?

victorusu: 👍

PR #998, opened 2019-03-14
[daint] Adapt naming scheme for Arolla/Tsa

kraushm:
Here, the machine daint is specified as a placeholder, since the intended one, Tsa, is not yet in the machines file. Note, however, that the modifications only change the module of EasyBuild.


kraushm:
retest this please

PR #999, opened 2019-03-15
[dom] Update Machines.groovy

lucamar:
I propose to remove Dom from production builds until we have a new software list, since the build will keep failing in the meanwhile. 
E.g.: The last failure occurred when building `GREASY` with the error `Unable to locate a modulefile for 'atp/2.1.2'`.
https://jenkins.cscs.ch/blue/rest/organizations/jenkins/pipelines/ProductionEB/runs/580/nodes/37/steps/74/log/?start=0

### DIFF
--- a/jenkins/Machines.groovy
+++ b/jenkins/Machines.groovy
@@ -38,4 +38,4 @@ def leone = [name: 'leone',
              modulesUnuseProduction: '',
              prefixProduction: '/apps/leone/UES/jenkins/RHEL6.10/easybuild']
 
-return [daint, dom, kesch, leone, fulen]
+return [daint, kesch, leone, fulen]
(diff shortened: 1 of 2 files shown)


teojgo:
@lucamar Don't remove from `Machines.groovy` since this is going to remove it from testing and regression as well. Its better to add:
```
machinesList.remove('dom')
```
after line 13 in `JenkinsfileProductionEB`

jenkins/Machines.groovy
```diff
@@ -38,4 +38,4 @@ def leone = [name: 'leone',
              modulesUnuseProduction: '',
              prefixProduction: '/apps/leone/UES/jenkins/RHEL6.10/easybuild']
 
-return [daint, dom, kesch, leone, fulen]
+return [daint, kesch, leone, fulen]
```
teojgo: Return also dom here since you now remove it only in production.

lucamar:
Thanks for the advice @teojgo, I have modified the pull request as you suggested.

PR #1012, opened 2019-03-28
[daint] lfs

jgphpc:
git-lfs needed for reframe

### DIFF
--- a/easybuild/easyconfigs/g/git-lfs/git-lfs-2.7.1.eb
+++ b/easybuild/easyconfigs/g/git-lfs/git-lfs-2.7.1.eb
@@ -1,0 +1,31 @@
+# @author: jg (cscs)
+easyblock = "Tarball"
+
+name = 'git-lfs'
+version = '2.7.1'
+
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/g/git-lfs/git-lfs-2.7.1.eb
```diff
@@ -0,0 +1,31 @@
...
+sources = ['%(name)s-linux-amd64-v%(version)s.tar.gz']
+
+sanity_check_paths = {
+    'files': ["git-lfs"],
+    'dirs': [""]
+}
+
+# add the installation dir to PATH
+modextrapaths = {
+    'PATH': '',
```
kraushm: Isn't something missing here?

jgphpc: no because there is no bin in the install dir hence i just need `PATH=$installdir` instead of `PATH=$installdir/bin`

PR #1015, opened 2019-04-01
[daint-gpu] NAMD/PLUMED recipe

kraushm:
Request for NAMD/PLUMED based on webRT https://webrt.cscs.ch/Ticket/Display.html?id=35317. 

This is a rather crude version, forcing the patch doing `plumed_patch -f ...`. The module was provided to the user, who reported back that his test case recognizes PLUMED. 
The idea is to move the `plumed_patch` to the easyblock.


lucamar:
I have removed the `WIP` from the title since @ChristopherBignamini approved the request. I have set the title to build only on `daint-gpu` in order to spare time, although both `NAMD` recipes are addressed in this pull request.

lucamar:
The build failed since it reached the wall time: I'm removing `guile-2.0.14-CrayIntel-17.08.eb` from the pull request.

PR #1021, opened 2019-04-03
[DOM_cle7] Rename to DOM_cle7-* due to Jenkins setup

kraushm:
Jenkins was looking for dom*, hence triggering also dom-gpu/mc.


teojgo:
retest this please

PR #1023, opened 2019-04-03
[daint-gpu] Add CNTK-2.6

rsarm:


easybuild/easyconfigs/l/libzip/libzip-1.5.1-CrayGNU-18.08-cmake3.12.0.eb
```diff
@@ -0,0 +1,25 @@
+# contributed by Luca Marsella and Guilherme Peretti Pezzi (CSCS)
+easyblock = "CMakeMake"
+
+name = 'libzip'
+version = '1.5.1'
+
+versionsuffix = '-cmake3.12.0'
```
lucamar: Thanks @rsarm: since `libzip` is used as a dependency by `CNTK` only, I suggest to remove the `versionsuffix` and change the existing recipe to use the most recent `CMake`

PR #1024, opened 2019-04-03
[DOM_cle7] Correct unuse patch in machine file

kraushm:
Correct copy-paste error. Unuse path now points to currently installed modules.


jgphpc:
Will now use:

- /apps/dom/UES/jenkins/7.0.UP00/gpu/easybuild/modules/all/daint-gpu
- /apps/dom/UES/jenkins/7.0.UP00/mc/easybuild/modules/all/daint-mc

instead of  /apps/dom/UES/easybuild/modulefiles/

PR #1026, opened 2019-04-03
[dom] Revert settings made for Jenkins

kraushm:
Parallel development to existing production stack is not needed and,
therefore, the settings for Jenkins are reverted to the original naming
of the agents with updated prefix.


jenkins/Machines.groovy
```diff
@@ -12,7 +12,7 @@ def dom = [name: 'dom',
            unusePath: '/apps/dom/UES/jenkins/6.0.UP07/ARCH/easybuild/tools/modules/all:/apps/dom/UES/jenkins/6.0.UP07/ARCH/easybuild/modules/all',
```
lucamar: Thanks @kraushm, I would also remove remove or update the `unusePath`

PR #1034, opened 2019-04-04
[dom-mc] CP2K with TC 19.03 (mc)

kraushm:


vkarak:
@kraushm Perhaps you have to open a Jira case

lucamar:
@vkarak I have created https://jira.cscs.ch/browse/MAINT-35

kraushm:
It just seems to miss some deps. This should be fixed once another PR is merged with these deps, e. g. Boost.

lucamar:
@kraushm Also the patch file was missing, since it takes the `name` and `version` of the recipe. I'm fixing it in my branch and I'll pull the changes.

PR #1035, opened 2019-04-04
[dom] CPMD with TC 19.03

kraushm:


easybuild/easyconfigs/c/CPMD/CPMD-4.1-CrayIntel-19.03.eb
```diff
@@ -0,0 +1,40 @@
...
+    CPMD The CPMD code is a parallelized plane wave / pseudopotential
+    implementation of Density Functional Theory, particularly designed
+    for ab-initio molecular dynamics.
+"""
+
+toolchain = {'name': 'CrayIntel', 'version': '19.03'}
+toolchainopts = {'usempi': True}
+
+sources = [
+    '/apps/common/UES/easybuild/sources/c/CPMD/cpmd-v%s.tar.gz' %version
```
vkarak: So `/apps/common` remains. Let's discuss tomorrow with @lucamar what we are going to do and update the Jira issue.

kraushm: As discussed this morning, we would keep `/apps/common` for sources.

PR #1038, opened 2019-04-04
[dom] pprof for PE 19.03

jgphpc:
```
pprof-9daae4f.eb                     --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/gpu/easybuild/tools
```

```
pprof-9daae4f.eb                     --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
```


jgphpc:
I need to push 1 more commit

PR #1050, opened 2019-04-05
[dom-mc] Update NAMD (mc) to TC 19.03

kraushm:

PR #1053, opened 2019-04-05
[dom-mc] Update NAMD_mc to TC 19.03

ChristopherBignamini:


lucamar:
@ChristopherBignamini There is already pull request #1050 open for this recipe (please use `[dom-gpu]` or `[dom-mc]` only in the title of the pull request if you don't need both, since `[dom]` will build on both of them and it will take more time)

vkarak:
@ChristopherBignamini Isn't this a duplicate of #1050 ?

ChristopherBignamini:
@lucamar @vkarak yes, it can be deleted. The NAMD easyconfig in #1050 is identical to the one of this PR and should work.

vkarak:
@lucamar I am closing this PR then.

PR #1054, opened 2019-04-06
[dom] Update production list 19.03

kraushm:


kraushm:
The lists are based on the modules already merged into master. The plan is to also add the merge-ready PRs, but for that I will wait a bit. In the end, this list should indicate, which applications still need to be build.

lucamar:
I have corrected the version of `PyExtensions` to match the ones available with the toolchain `19.03`.

PR #1062, opened 2019-04-10
[dom-gpu] VASP recipes for cle7

lucamar:
I have inserted a symbolic link to the `makefile.include`, since it is the same as before.


easybuild/easyconfigs/v/VASP/VASP-5.4.4-CrayIntel-19.03-cuda-10.0.makefile.include
```diff
@@ -0,0 +1 @@
+VASP-5.4.4-CrayIntel-18.08-cuda-9.1.makefile.include
```
vkarak: Shouldn't we change this as well to 19.03?

lucamar: I have set a symbolic link to that file with the new name, since they are the same.

PR #1065, opened 2019-04-10
[dom] Port NCO to PE 19.03

vkarak:


easybuild/easyconfigs/j/JasPer/JasPer-2.0.14-CrayGNU-19.03.eb
```diff
@@ -0,0 +1,28 @@
...
+description = """The JasPer Project is an open-source initiative to provide a free
+ software-based reference implementation of the codec specified in the JPEG-2000 Part-1 standard."""
+
+toolchain = {'name': 'CrayGNU', 'version': '19.03'}
+toolchainopts = {'pic': True}
+
+source_urls = ['http://www.ece.uvic.ca/~frodo/jasper/software/']
+sources = [SOURCELOWER_TAR_GZ]
+
+builddependencies = [('CMake', '3.13.3', '', True)]
```
vkarak: Which version of CMake are we using?

vkarak: Found it, 3.14.1.

kraushm:
As Miguel Gila pointed out, java is installed on the elogin and not on the PE image, as it was on CLE6. For the build, changing to the osdependency seems to work. To be seen how things are during runtime.

PR #1067, opened 2019-04-10
[dom] Port Spark to PE 19.03

vkarak:


easybuild/easyconfigs/s/Spark/Spark-2.3.1-CrayGNU-19.03-Hadoop-2.7.eb
```diff
@@ -0,0 +1,86 @@
...
+source_urls = [
+    'http://apache.belnet.be/%(namelower)s/%(namelower)s-%(version)s/',
+    'http://www.eu.apache.org/dist/%(namelower)s/%(namelower)s-%(version)s/',
+    'http://www.us.apache.org/dist/%(namelower)s/%(namelower)s-%(version)s/',
+]
+
+checksums = ['e87499e5417a64341cbda25e087632dd9f6ce7ad249dfeba47d9d02a51305fc2']
+
+dependencies = [
+    ('java/%s' % javaver, EXTERNAL_MODULE),
```
kraushm: It fails for the missing dependency. Do we ask ops to provide it or are we responsible for this one?

easybuild/easyconfigs/s/Spark/Spark-2.3.1-CrayGNU-19.03-Hadoop-2.7.eb
```diff
@@ -0,0 +1,89 @@
+easyblock = 'Tarball'
+
+name = 'Spark'
+version = '2.3.1'
+
+javaver = 'jdk1.8.0_51'
```
vkarak: You need to comment this out.

PR #1072, opened 2019-04-10
[dom-gpu] Port MAGMA recipes to PE 19.03

vkarak:


ajocksch:
I believe one intel version was a requirement of Anton.

toxa81:
Magma works best with Intel/MKL as they develop for this combination.

toxa81:
CrayIntel version is a must

PR #1073, opened 2019-04-10
[dom] VTK update to TC 19.03

kraushm:


vkarak:
Do we need both versions?

jfavre:
I would keep only 8.2 actually

PR #1081, opened 2019-04-11
[dom-gpu] Updating production list for CLE7

lucamar:
The pull request provides an updated production list for CLE7: it will be merged after few new items will be inserted.


jenkins-builds/7.0.UP00-19.03-mc
```diff
@@ -34,6 +36,7 @@
  PyExtensions-2.7.15.6-CrayGNU-19.03.eb
  PyExtensions-3.6.5.6-CrayGNU-19.03.eb
  Python-bare-3.6.2.eb                               --hidden
+ Spark-2.3.1-CrayGNU-19.03-Hadoop-2.7.eb            --set-default-module
  VASP-5.4.4-CrayIntel-19.03.eb                      --set-default-module
  gperftools-2.7.eb                                  --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
```
vkarak: Should we put a comment here that we purposely have the tools at the end? Because everything else is alphabetically ordered.

jenkins-builds/7.0.UP00-19.03-mc
```diff
@@ -24,7 +24,9 @@
  ipyparallel-6.2.3-CrayGNU-19.03-python3.eb
  jupyter-1.0.0-CrayGNU-19.03.eb
  LAMMPS-22Aug2018-CrayGNU-19.03.eb                  --set-default-module
+ LLVM-8.0.0-CrayGNU-19.03.eb
```
vkarak: Shouldn't we include this in the GPU list?

lucamar: It's not present in the current GPU list.

jenkins-builds/7.0.UP00-19.03-mc
lucamar: Yes, that is the silent agreement, However we cannot insert comments in the production list.

jenkins-builds/7.0.UP00-19.03-mc
jgphpc: i am preparing a recipe for clang, it will replace llvm.

jfavre:
before removing it completely, be aware that I am currently testing a compilation of Mesa (to be used by ParaView), which depends on LLVM. It currently uses LLVM 7.0.1, but I wanted to test it with your new LLVM 8.

reference: https://jira.cscs.ch/browse/UES-236

N.B. I only use LLVM on the MC partition.

lucamar:
@jfavre If you insert `LLVM-8.0.0-CrayGNU-19.03.eb` as a dependency in the recipe of `Mesa`, then we can remove it from the production list: in fact the recipe is present in the repository and it will be built as a dependency, so it does not need to appear explicitly. The same holds for `Mesa` itself, if it is a dependency of `ParaView`.

lucamar:
Please wait that #1098 is merged and the Intel toolchain is forced before merging this pull request.

vkarak:
@lucamar Could we merge a first of this today, so that by tomorrow we have regression results for the applications already built?

jenkins-builds/7.0.UP00-19.03-mc
lucamar: I'm removing version `8.0.0` from the list, after discussing with @jfavre: the updated recipe will still be in the repository and it will be available for next releases of `Mesa` and `ParaView`, if applicable.

lucamar:
I'm merging the pull request without `ParaView` since #1102 has not yet completed.

PR #1084, opened 2019-04-11
[dom-gpu] Add Horovod and Tensorflow

rsarm:


jenkins-builds/7.0.UP00-19.03-gpu
```diff
@@ -15,6 +15,7 @@
  GSL-2.5-CrayCCE-19.03.eb
  GSL-2.5-CrayIntel-19.03.eb
  HDFView-2.14.eb                                    --set-default-module
+ Horovod-0.16.0-CrayGNU-19.03-tf-1.12.0.eb          --set-default-module
```
lucamar: @rsarm Please remove the file `7.0.UP00-19.03-gpu` from your pull request: we will update the production list separately once the recipe will be merged in the repository (see #1081)

easybuild/easyconfigs/c/cuDNN/cuDNN-7.5.0-cuda-10.0.eb
```diff
@@ -0,0 +1,28 @@
...
+description = """Nvidias deep neural network library."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+cudamajor = '10.0'
+
+versionsuffix = '-cuda-%s' % cudamajor
+
+dependencies = [
+    ('craype-accel-nvidia60', EXTERNAL_MODULE),
```
lucamar: @rsarm Do you really need `craype-accel-nvidia60` as dependency? Or is it enough `cudatoolkit` as a builddependency?

rsarm: cuDNN is only available as binaries. I assumed that when the module is loaded, it would need `cudatoolkit` but now that you mention it, it I do `ldd` to the cuDNN binaries, I see that it doesn't refer to `cudatoolkit`. Should I remove `craype-accel-nvidia60`?

lucamar: Yes, please remove it if it is not necessary

easybuild/easyconfigs/n/NCCL/NCCL-2.4.2-cuda-10.0.eb
```diff
@@ -0,0 +1,28 @@
...
+description = """Nvidias deep neural network library."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+cudamajor = '10.0'
+
+versionsuffix = '-cuda-%s' % cudamajor
+
+dependencies = [
+    ('craype-accel-nvidia60', EXTERNAL_MODULE),
```
lucamar: @rsarm Please check also if this is necessary

lucamar:
@rsarm I have checked the log file and found this error in the build of `TensorFlow`:
```
Please specify the location where CUDA 10.0 toolkit is installed. Refer to README.md for more details. [Default is /u
sr/local/cuda]:

Invalid path to CUDA 10.0 toolkit. ['/usr/local/cuda/lib64/libcudart.so.10.0', '/usr/local/cuda/lib/powerpc64le-linux
-gnu/libcudart.so.10.0', '/usr/local/cuda/lib/x86_64-linux-gnu/libcudart.so.10.0'] cannot be found
Please specify the CUDA SDK version you want to use. [Leave empty to default to CUDA 9.0]:
```
`/scratch/snx1600tds/jenscscs/TestingEB/dom-gpu/3321349-1853/tmp/eb-UAneJd/easybuild-TensorFlow-1.12.0-20190412.111026.oGSnf.log` referenced from 
https://jenkins.cscs.ch/blue/organizations/jenkins/TestingEB/detail/TestingEB/1853/pipeline

PR #1093, opened 2019-04-12
[dom] ScoreP for 19.03

jgphpc:
https://jira.cscs.ch/browse/MAINT-62


vkarak:
@lucamar Can we merge this, or should we wait for the updated `CrayIntel` toolchain?

lucamar:
@vkarak `CrayIntel` has already been updated, therefore we can proceed.

PR #1095, opened 2019-04-12
[dom] Update otfprofile to TC 19.03

kraushm:


jgphpc:
`cube-config` missing, wip

kraushm:
The dependencies, i. e. Cube, have changed and it has to be investigated, whether the developers will update the tool or whether work on the tool discontinued.

PR #1096, opened 2019-04-12
[dom-gpu] Plumed 2.5.1 with CrayIntel 19.03

lucamar:
`Plumed` does not use anymore `libmatheval` since release `2.5`, therefore I have removed it from the dependencies: see https://plumed.github.io/doc-v2.5/user-doc/html/_c_h_a_n_g_e_s-2-5.html

I keep anyway the recipe of `libmatheval` and its dependencies as they were already tested on `19.03`.


lucamar:
retest this please

PR #1123, opened 2019-05-08
CrayPE in container

mmxcscs:
This is the same information as on CSCS gitlab. We make it public based on interest of other HPC sites who asked for the code during CUG.


mmxcscs:
I don't think that we can use Jenkins to build the CPE, but once CPE are built, integrating them into Jenkins for building the applications should be possible.

PR #1131, opened 2019-05-20
WIP [keschTDS] Add gcc/7.4.0-2.32 to experimental stack

kraushm:
Adding gcc/7.4.0 for MCH to test with FE.

This way was preferred to providing gcccore only, since no change in the workflow is required regarding the modules loaded before compilation.


kraushm:
Uses gcc/7.4.0 recipe with the same name as default EB repository, yet changing dependency versions. Furthermore, based on old branch.

PR #1143, opened 2019-06-03
Add Spack configuration options on Piz Daint

victorusu:
Adding basic configurations to allow Spack to be used on Piz Daint. The intention is to tell users to copy this file and use it as a base for their own configuration. These files are intended to speed up the process for the users to start using Spack if they want to. A user portal PR follows in our internal repo.

The files should be copied to `~/.spack/cray/`. This way it does not affect other systems configurations. But it could also be copied to `~/.spack`, but I do not suggest it, since these files are Piz Daint specific.

Since we are neither providing a module nor a Spack installation on the system, like we do for EasyBuild, the `config.yaml` file have hard-coded paths. But the users can easily modify the file.

The `modules.yaml` file contains a list of suggestion of packages to be hidden and several modifications to module names. This allows several installations of the same packages (that use different configuration options) to live side-by-side on a flat modules installation like TMod. The list is not complete and it is not the intention.

The `packages.yaml` file is based on the one provided by Spack (https://github.com/spack/spack-configs/blob/master/NERSC/cori/packages.yaml) for NERSC Cori. I have adapted the versions and added other packages in order to better fit our needs. 

I have tested this configuration using
```
mkdir -o $HOME/bin
cd $HOME/bin
git clone https://github.com/spack/spack.git
```

And I added the following lines to the end of `~/.bashrc` file
```
export SPACK_ROOT=$HOME/bin/spack
export PATH=$SPACK_ROOT/bin:$PATH
. $SPACK_ROOT/share/spack/setup-env.sh
```

You can test it, for instance, using 
```
spack compiler find
spack install quantum-espresso hdf5=parallel %intel target=broadwell
spack install gromacs %gcc target=broadwell
spack install quantum-espresso hdf5=parallel %intel target=haswell
spack install gromacs %gcc target=haswell
```

Then you can add the modules from spack to your environment and test
```
module use $HOME/daint/spack/share/modules/*
```

Using the current Spack master and the current Piz Daint, it should generate the following
```
----------------------------------------- $HOME/daint/spack/share/modules/cray-cnl6-haswell ------------------------------------------
cmake/3.14.4-gcc-6.2.0-cray-cnl6-haswell                            quantum-espresso/6.4-intel-18.0.2.199-cray-cnl6-haswell-mpi-omp-mkl
gromacs/2019.2-gcc-6.2.0-cray-cnl6-haswell-release-cuda-mpi-shared

---------------------------------------- $HOME/daint/spack/share/modules/cray-cnl6-broadwell -----------------------------------------
cmake/3.14.4-gcc-6.2.0-cray-cnl6-broadwell                           gromacs/2019.2-gcc-6.2.0-cray-cnl6-broadwell-release-cuda-mpi-shared

```

Please read the user portal PR in our internal repo for further info.


victorusu:
I have added a very long list of people to review, because you either have experience with Spack on the system or you are somehow gonna be highly affected by this PR. 😄

PR #1146, opened 2019-06-04
[dom] Create jupyterhub-0.9.6-CrayGNU-19.03.eb

twrobinson:


twrobinson:
retest this please

twrobinson:
The versions of JupyterHub should match between JupyterHub itself (on the VM), and the single-user server that it spawns (on Dom/Daint). i am updating to 0.9.6 on the VMs so we need the same version on Dom/Daint.

PR #1157, opened 2019-06-06
[daint] Use only first column of the productions lists

teojgo:


lucamar:
@teojgo I think you want to remove the flag `--installpath=...` only, since other flags are harmless (e.g.: `--module-naming-scheme=... on Kesch).

teojgo:
@lucamar Are they harmless or they are needed?

lucamar:
I would keep the `--module-naming-scheme=...` flag on Kesch for testing in `RegressionEB`, since we need it in production. Furthermore, in the future system we will have other necessary flags that will not be on the first column.

PR #1160, opened 2019-06-06
[daint-gpu] Suggest EB module fix w.r.t lmod

kraushm:
This is a suggestion for a fix, which became necessary for the lmod installed on Tsa (tsa-cn017/cn018).

### DIFF
--- a/easybuild/module/EasyBuild-custom/cscs
+++ b/easybuild/module/EasyBuild-custom/cscs
@@ -146,6 +146,6 @@ if { ! [ info exists ::env(EASYBUILD_SOURCEPATH) ] } {
 prepend-path MODULEPATH                         $::env(EASYBUILD_INSTALLPATH)/modules/all
 # Loading easybuild module
 module use /apps/common/UES/easybuild/modules/all
-module load /apps/common/UES/easybuild/modules/all/EasyBuild/latest
+module load EasyBuild
 
 # Built with EasyBuild version 3.6.2
--- a/easybuild/module/EasyBuild-custom/cscs
+++ b/easybuild/module/EasyBuild-custom/cscs
@@ -145,6 +145,7 @@ if { ! [ info exists ::env(EASYBUILD_SOURCEPATH) ] } {
 # Adding easybuild installed software to the list of modules
 prepend-path MODULEPATH                         $::env(EASYBUILD_INSTALLPATH)/modules/all
 # Loading easybuild module
+module use /apps/common/UES/easybuild/modules/all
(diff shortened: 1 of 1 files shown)


easybuild/module/EasyBuild-custom/cscs
```diff
@@ -145,6 +145,7 @@ if { ! [ info exists ::env(EASYBUILD_SOURCEPATH) ] } {
 # Adding easybuild installed software to the list of modules
 prepend-path MODULEPATH                         $::env(EASYBUILD_INSTALLPATH)/modules/all
 # Loading easybuild module
+module use /apps/common/UES/easybuild/modules/all
 module load /apps/common/UES/easybuild/modules/all/EasyBuild/latest
```
gppezzi:
```suggestion
module load EasyBuild/latest
```

gppezzi:
Just for the record: the reason behind this PR is that Lmod in tsa doesn't like to load a module using an absolute path as we were doing so far for EasyBuild.

gppezzi:
If we add this new `module use /apps/common/UES/easybuild/modules/all` statement, we could eliminate the 'latest' symlink (which sometimes is confusing), define a default for eb (the modules way) and simply do:
```
module load EasyBuild
``` 

How about that @lucamar @victorusu @kraushm ?

lucamar:
@gppezzi The modules under `/apps/common/UES/easybuild/modules/all/EasyBuild` are created by EasyBuild `bootstrap.py`, e.g. `/apps/common/UES/easybuild/bootstrap_eb.py-3.9.1`: is there any option to set a module as default running `bootstrap`? I only see the following options
```
usage: bootstrap_eb.py-3.9.1 [-h] prefix

positional arguments:
  prefix      Installation prefix directory

optional arguments:
  -h, --help  show this help message and exit
```

gppezzi:
By "module way" I meant creating a `.version` file. I've just set the default to 3.9.1 for testing: 
```
EasyBuild/3.9.1(default)
```

lucamar:
@gppezzi That's fine for me, we just need to manually update the `.version` file after each EasyBuild installation instead of updating the symlink

kraushm:
> If we add this new `module use /apps/common/UES/easybuild/modules/all` statement, we could eliminate the 'latest' symlink (which sometimes is confusing), define a default for eb (the modules way) and simply do:
> 
> ```
> module load EasyBuild
> ```
> 
> How about that @lucamar @victorusu @kraushm ?

Tests on Tsa and Daint worked fine, therefore I commited the proposed change.

PR #1163, opened 2019-06-11
[daint] Spack for Piz Daint

victorusu:
Add Spack EB recipe pre-configure for the current Piz Daint.

### DIFF
--- a/easybuild/easyconfigs/s/Spack/Spack-daint.eb
+++ b/easybuild/easyconfigs/s/Spack/Spack-daint.eb
@@ -1,0 +1,96 @@
+#@author: Victor Holanda
+easyblock = 'Tarball'
+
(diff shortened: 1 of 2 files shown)


victorusu:
retest this please

easybuild/easyconfigs/s/Spack/Spack-daint.eb
```diff
@@ -0,0 +1,96 @@
+#@author: Victor Holanda
+easyblock = 'Tarball'
+
+name = 'Spack'
+version = 'daint'
+
```
gppezzi: Maybe add a hint in the module name in order to tell users that this version comes from the git master branch?
Something like:
```
versionsuffix = '-git' or '-git-master' 
```

PR #1168, opened 2019-06-17
[dom] jupyterhub 1.0

twrobinson:
Add JupyterHub 1.0.0 release! Also made default so to be consistent with jupyter-dev.


twrobinson:
retest this please

PR #1172, opened 2019-06-22
[dom-gpu] nvprof

jgphpc:
Will fix https://jira.cscs.ch/browse/MAINT-87

### DIFF
--- a/jenkins-builds/7.0.UP00-19.03-mc
+++ b/jenkins-builds/7.0.UP00-19.03-mc
@@ -59,6 +59,7 @@
  mpiP-57fc864-CrayGNU-19.03.eb                      --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
  mpiP-57fc864-CrayIntel-19.03.eb                                         --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
  MUST-v1.6-rc1-CrayGNU-19.03.eb                     --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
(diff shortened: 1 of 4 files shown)


jenkins-builds/7.0.UP00-19.03-mc
```diff
@@ -59,6 +59,7 @@
  mpiP-57fc864-CrayGNU-19.03.eb                      --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
  mpiP-57fc864-CrayIntel-19.03.eb                                         --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
  MUST-v1.6-rc1-CrayGNU-19.03.eb                     --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
+ nvidia-nsight-2019.3.1.eb                          --set-default-module --installpath=$APPS/UES/jenkins/7.0.UP00/mc/easybuild/tools
```
kraushm: Does nsight also work on CPU only applications?

jgphpc: maybe but with https://developer.nvidia.com/nsight-systems

jgphpc: nvprof is on the $PATH when `cudatoolkit` is loaded.

jgphpc: ah nvtop

kraushm: Check for nvtop. Regarding `nsight`, this installation is not for `nsight-systems`, is it? Does it make sense then to add it to the `mc` production list?

jgphpc: let's focus on gpu first + i will do another PR for nsight-systems if needed

PR #1173, opened 2019-06-24
[daint] Fix Spack module name (develop)

gppezzi:
The default branch when performing git clone seems to be actually `develop` and not `master`.


gppezzi:
retest this please

PR #1180, opened 2019-07-01
[daint dom]  Update production list to include dask 1.2.2

twrobinson:
updating dask to 1.2.2


lucamar:
Thanks @twrobinson: you are not changing the default to `dask-1.0.0-CrayGNU-19.03-python3.eb`, even in the new software stack?

twrobinson:
Hi @lucamar yes, I'll make it default everywhere, thanks for checking

PR #1192, opened 2019-07-08
[dom] Add LLVM 8.0.0

victorusu:


jfavre:
Thanks @victorusu . I need to see if this version can be used instead of the one proposed by the ParaView developers as per my Jira issue https://jira.cscs.ch/browse/UES-236

PR #1193, opened 2019-07-09
[daint] Add toolchainless GREASY

victorusu:
Since GREASY uses c++11 and the current system gcc on diant is 4.8.5, we need to boost the gcc in order to compile the code. I am using the current default on daint gcc/6.2.0. 

I am changing the default because it does not bring any regression to the users.


victorusu:
@lucamar, I would say so, in case a user uses the full module name.

PR #1202, opened 2019-07-12
[dom-gpu] Add TensorFlow-1.14.0 for CrayGNU-19.06

rsarm:


rsarm:
retry this please

victorusu:
@gppezzi and @rsarm, I just have this problem with having a CUDA module visible. I think we should hide it in order to avoid confusion. What do you guys think?

Apart from that `lgtm`!

gppezzi:
This recipe is for the TDS. For the final deployment I hope that this cuda toolkit will be available from Cray. But hiding it is a good idea, so that we don't forget and add it to the public modules without realising.

PR #1212, opened 2019-07-18
[dom] Added Boost/1.70 with PE19.06

sebkelle1:


teojgo:
retest this please

teojgo:
I Think we need also the version which does not have the python suffix.

sebkelle1:
@jenkins-cscs retest dom

sebkelle1:
@jenkins-cscs retry dom

sebkelle1:
For 1.69.0, I just added one test version, I don't think we need the full set for 1.69.0, but it's of course
easy to add if you think we need it.

teojgo:
Then it's fine. In any case we can put only the latest one in the production lists.

easybuild/easyconfigs/b/Boost/Boost-1.70.0-CrayGNU-19.06.eb
```diff
@@ -0,0 +1,29 @@
+easyblock = 'boostcray'
```
lucamar: Is the custom easyblock `boostcray` still needed? 
The other recipes in your pull request use the standard easyblock `Boost`.

sebkelle1: No, it's possible without `boostcray`

gppezzi: in that case I think it is better to avoid using the custom easyblock if not really needed.

sebkelle1: I removed it

PR #1247, opened 2019-08-12
[dom] jupyterlab-1.0.4-CrayGNU-19.06.eb

twrobinson:
JLab 1.x. Major upgrade of JLab easyconfig file with inclusion of JHub dependecies directly and a host of extensions including jupyter-widgets, plotlywidgets, bokeh, topbar, bqplot, dask-labextension, jupyter_server_proxy. hub-extension itself is no longer required as the hub functionality it is built directly in JLab core.


twrobinson:
retest this please

PR #1248, opened 2019-08-12
[daint] Create IPython-7.7.0-CrayGNU-18.08-python3.eb

twrobinson:


twrobinson:
@victorusu @teojgo Any concerns? If not, I will request Luca to merge

PR #1255, opened 2019-08-13
[daint,dom] Update production lists to add dask 2.2.0

twrobinson:
add dask 2.2.0 and make it default.


twrobinson:
retest this please

PR #1258, opened 2019-08-14
[daint] Create R-3.4.2 for CrayGNU 18.08

teojgo:


teojgo:
@lucamar then we should also remove some of the added eb recipes not needed anymore.

PR #1262, opened 2019-08-15
[dom] jupyterlab-1.0.4-CrayGNU-19.06.eb

twrobinson:
Only build the web pack once by doing a --no-build for individual components followed by a jupyter lab build. Reduce the size of the JS bundle significantly by doing a production build only (jupyter lab build --dev-build=False) - speed up launching of JuptyerLab at run time (launch in ca. 10 sec vs 30 sec).


twrobinson:
retest this please

PR #1269, opened 2019-08-20
[dom] Update jupyterlab-1.0.4-CrayGNU-19.06.eb

twrobinson:


twrobinson:
retest this please

twrobinson:
@rsarm this test fails but I believe this is to do with the temporary install files. Can we merge and test this in a real build? It's very hard to debug otherwise. (the build script build fine locally with eb ./<recipe> -f)

twrobinson:
OK, all checks pass now, thanks @rsarm

PR #1273, opened 2019-08-22
[dom] Update 7.0.UP00-19.06-gpu

twrobinson:
Add julia 1.1.1 to dom. This is not targetted for final production on Daint but is a placeholder for testing purposes for creating an IJulia kernel for jupyterlab


omlins:
Maybe it would be better to do this with the long-term support (LTS) release: v1.0.4?

twrobinson:
Yes, sounds good to me.

PR #1278, opened 2019-08-25
[dom] Extrae/3.7.1

jgphpc:
Will fix UES-490

### DIFF
--- a/easybuild/easyconfigs/b/basicanalysis/basicanalysis-0.3.6.eb
+++ b/easybuild/easyconfigs/b/basicanalysis/basicanalysis-0.3.6.eb
@@ -12,14 +12,14 @@ sources = ['%(name)s-%(version)s-src.tar.bz2']
 checksums = ['6909417304f18f5573bb9bcfa2d19941']
 keepsymlinks = True
 postinstallcmds = [
-    # The following commands will allow greater flexibility in upgrading Extrae's dependencies
-    # that are binaries without having to rebuild Extrae:
-	#   ls -l easybuild/software/basicanalysis/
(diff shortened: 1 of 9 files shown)


easybuild/easyconfigs/b/basicanalysis/basicanalysis-0.3.6.eb
```diff
@@ -12,14 +12,14 @@ sources = ['%(name)s-%(version)s-src.tar.bz2']
...
-	#       latest -> 0.3.6
+    # The following commands will allow greater flexibility in upgrading
+    # Extrae's dependencies that are binaries without having to rebuild Extrae:
+    #   ls -l easybuild/software/basicanalysis/
+        #       0.3.6
+        #       latest -> 0.3.6/
+    #   ls -l easybuild/modules/all/basicanalysis/
+        #       0.3.6
+        #       latest -> 0.3.6
     'ln -fs %(version)s/ %(installdir)s/../latest && ' \
```
gppezzi: If we install an older version of this package, `latest` will point to the last version installed (which is not necessarily the latest)?

I would suggest anyway to add new easyconfigs in order to create wrapper modules, so that we can explicit control over the value of `latest`. 

You can see for example how this is done for Java, where the installation is split in two: [jdk installation](https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/j/Java/Java-1.8.0_212.eb) and [wrapper](https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/j/Java/Java-1.8.eb).

PR #1287, opened 2019-09-02
[dom] Create jupyterlab-1.1.1-CrayGNU-19.06.eb

twrobinson:
Adding 1.1.1 because 1.0.X is now unbuildable thanks to changes in the javascript package manager. Also added graphviz module and associated graphviz python module  for visualising dask task graphs.


twrobinson:
Note that the graphviz module (which provides the executables and libraries) and the python module are both required.

PR #1288, opened 2019-09-02
[dom] Update 7.0.UP00-19.06-gpu

twrobinson:
Test if there is a difference in the behaviour of the depot paths and load paths between julia 1.0.x and 1.1.1

### DIFF
--- a/jenkins-builds/7.0.UP00-19.06-gpu
+++ b/jenkins-builds/7.0.UP00-19.06-gpu
@@ -29,7 +29,8 @@
  IDL-8.5.1.eb                                       --set-default-module
  ipykernel-4.8.2-CrayGNU-19.06-python2.eb
  ipyparallel-6.2.3-CrayGNU-19.06-python3.eb
- julia-1.0.4.eb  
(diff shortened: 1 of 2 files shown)


jenkins-builds/7.0.UP00-19.06-gpu
```diff
@@ -29,7 +29,8 @@
  IDL-8.5.1.eb                                       --set-default-module
  ipykernel-4.8.2-CrayGNU-19.06-python2.eb
  ipyparallel-6.2.3-CrayGNU-19.06-python3.eb
- julia-1.0.4.eb  
+ julia-1.0.4.eb
```
lucamar: @twrobinson Do you really need to keep also `julia-1.0.4.eb`?

twrobinson: Yes, I want to compare the behaviour between 1.0 1.1 and 1.2, in preparation for IJulia and JupyterLab deployment. With local builds there is too much "it works on my machine", hence my prefernece for a production-like build (Dom only!).

PR #1289, opened 2019-09-03
[dom] Create julia-1.2.0.eb

twrobinson:
Comparison of the behavior of environment variables across julia versions.


twrobinson:
This pull request is to enable testing of system and user package management in Julia + JupyterLab

PR #1292, opened 2019-09-04
[daint] Create jupyterlab-1.1.1-CrayGNU-18.08.eb

twrobinson:
Will be announcing JLab 1.1.1 at the user lab day. Looks like Julia is not going to ready, but I need to test dask lab extension in production (via jupyter.cscs.ch) to ensure ports are open as necessary etc, so this cannot be delayed any further


twrobinson:
retest this please

PR #1295, opened 2019-09-06
[daint] Create jupyterhub-1.0.0-CrayGNU-18.08.eb

twrobinson:
Add JHub 1.0.0 to support legacy 0.35.2 jupyterlab (newer JLab has jupyterhub in-built)


victorusu:
`lgtm`

PR #1303, opened 2019-09-13
[dom] Add 19.09 cray metadata

ekouts:
This is the updated metadata for CDT 19.09

### DIFF
--- a/easybuild/cray_external_modules_metadata.cfg
+++ b/easybuild/cray_external_modules_metadata.cfg
@@ -1,16 +1,6 @@
 # metadata useful to EasyBuild (hpcugent.github.io/easybuild) for modules provided by Cray on Cray systems
 # authors: Guilherme Peretti-Pezzi (CSCS), Petar Forai (IMP/IMBA), Kenneth Hoste (HPC-UGent), Eirini Koutsaniti (CSCS)
 
-[cray-R/3.3.3]
-name = CrayR
-version = 3.3.3
-prefix = /opt/R/3.3.3
-
-[cray-R/3.4.2]
-name = CrayR
-version = 3.4.2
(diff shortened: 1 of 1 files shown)


teojgo:
Ok to test

easybuild/cray_external_modules_metadata.cfg
```diff
@@ -1,5 +1,15 @@
...
+
+[cray-R/3.3.3]
+name = CrayR
+version = 3.3.3
+prefix = /opt/R/3.3.3
+
+[cray-R/3.4.2]
+name = CrayR
+version = 3.4.2
+prefix = /opt/R/3.4.2
```
vkarak: Looks like you've ordered the entries by taking the case into account, as normally `sort` would do. This can be a bit confusing for human beings for mixed case inputs. I would suggest to move `cray-R` entries down where they were placed initially, as if you were ignoring the case.

easybuild/cray_external_modules_metadata.cfg
```diff
@@ -321,11 +391,46 @@ name = TPSL
...
+[cudatoolkit/10.1.105_3.23-7.0.0.1_2.1__g85b26e9]
+name = CUDA
+version = 10.1.105_3.23-7.0.0.1_2.1__g85b26e9
+prefix = CRAY_CUDATOOLKIT_DIR
+
+[cudatoolkit/10.1.105_3.27-7.0.1.0_8.1__ga311ce7]
+name = CUDA
+version = 10.1.105_3.27-7.0.1.0_8.1__ga311ce7
+prefix = CRAY_CUDATOOLKIT_DIR
+
```
vkarak: Same comment here. I'd sort the versions numerically. It's easier to follow.

PR #1305, opened 2019-09-13
[dom-gpu] First recipe for COSMO-POMPA on Dom

omlins:


easybuild/easyconfigs/b/Boost/Boost-1.67.0-CrayGNU-19.08.eb
```diff
@@ -0,0 +1,26 @@
+easyblock = 'boostcray'
```
gppezzi: I think we should deprecate this `boostcray` easyblock. If this builds work with the standard easyblock from EB.
(se PR #1212), it could be safely removed.

omlins: Done.

easybuild/easyconfigs/c/CrayCCE/CrayCCE-19.08.eb
```diff
@@ -0,0 +1,26 @@
...
+(PE release: June 2019).
+"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}  # (default) version
+
+dependencies = [
+    ('PrgEnv-cray', EXTERNAL_MODULE),
+    ('atp/2.1.3', EXTERNAL_MODULE),
+    ('cce/9.0.1', EXTERNAL_MODULE),
+    ('modules/3.2.11.2', EXTERNAL_MODULE),
```
gppezzi: We current have an issue with using non default `modules`, therefore I suggest that you remove this from the dependencies (for the moment we will rely on the default version)

easybuild/easyconfigs/c/CrayGNU/CrayGNU-19.08.eb
```diff
@@ -0,0 +1,33 @@
...
+    ('PrgEnv-gnu', EXTERNAL_MODULE),
+    ('atp/2.1.3', EXTERNAL_MODULE),
+    # gcc versions later than 8 are not supported by cuda/10.1:
+    # -----------------   gcc/4   gcc/5   gcc/6   gcc/7   gcc/8   gcc/9
+    # cuda/09.2 < gcc/8   Y       Y       Y       Y       -       -
+    # cuda/10.0 < gcc/8   Y       Y       Y       Y       -       -
+    # cuda/10.1 < gcc/9   Y       Y       Y       Y       Y       -
+    # -----------------
+    ('gcc/8.3.0', EXTERNAL_MODULE),
+    ('modules/3.2.11.2', EXTERNAL_MODULE),
```
gppezzi: Please remove this line (see comment above).

easybuild/easyconfigs/b/Boost/Boost-1.67.0-CrayGNU-19.08.eb
omlins: It failed when I removed `easyblock = 'boostcray'`. So, I added it again.

easybuild/easyconfigs/c/CrayCCE/CrayCCE-19.08.eb
omlins: Done.

easybuild/easyconfigs/c/CrayGNU/CrayGNU-19.08.eb
omlins: Done.

PR #1310, opened 2019-09-16
[dom] Fix Julia related package installation

victorusu:


omlins:
@twrobinson 
> The extensions don't appear to be pre-compiled. I think this means you will end up with duplicated 
packages for every user, which defeats the purpose of having centrally installed extensions.

The extensions are intentionally not pre-compiled as 1) some CUDA packages need to be pre-compiled on a compute node and 2) the result of pre-compilation is one single file. E.g. in my home, my folder `compiled` containes 46 files, making up a size of 22 MB, i.e. it is completely negligible:
```
omlins@daint106:~> du -sh .julia/1.0.4/daint-gpu/compiled/v1.0/
22M	.julia/1.0.4/daint-gpu/compiled/v1.0/
omlins@daint106:~> find .julia/1.0.4/daint-gpu/compiled/v1.0 -type f | wc -l
46
```
For each user the pre-compilation will only happen once...

PR #1319, opened 2019-09-18
WIP [dom] Update recipes to PE19.09

ekouts:
These recipes fail
- dask-labextension-1.0.3-CrayGNU-19.09-python3.eb
- ParaView-5.6.0-OSMesa-CrayGNU-19.09.eb


twrobinson:
dask-labextension-1.0.3-CrayGNU-19.09-python3.eb deleted, as it is no longer used (extension is built directly in jupyterlab easyconfig)

lucamar:
Thanks a lot @ekouts: I think that we can clean up a bit before building the software stack, since we have multiple releases of certain packages and if they are not strictly necessary, we should keep only the most recent release. For instance I see the following multiple builds:
1. `dask`, `jupyterhub` and `jupyterlab` @twrobinson 
1. `HPX` @msimberg 
1. `Cubelib` and `CubeW` @jgphpc
Please have a look and let me know, so that we can update the branch `toolchains-19.09`

ekouts:
ParaView-5.6.0-OSMesa-CrayGNU-19.09.eb can also be deleted according to Jean Favre

twrobinson:
> Thanks a lot @ekouts: I think that we can clean up a bit before building the software stack, since we have multiple releases of certain packages and if they are not strictly necessary, we should keep only the most recent release. I see at least the following multiple builds, but there might be more:
> 
>     1. `dask`, `jupyterhub` and `jupyterlab` @twrobinson
> 
>     2. `HPX` @msimberg
> 
>     3. `Cubelib` and `CubeW` @jgphpc
>        Please have a look and let me know, so that we can update the branch `toolchains-19.09`

As we still make the older JLab selectable from the GUI in production I prefer to keep multiple versions for now. I have removed JLab 1.0.4 as it is no longer buildable.

jenkins-builds/7.0.UP01-19.09-mc
```diff
@@ -0,0 +1,88 @@
+ abcpy-0.5.1-CrayGNU-19.09-python3.eb               --set-default-module
```
gppezzi: I suggest we remove abcpy from the jenkins list, unless someone is aware of production projects using it.

jenkins-builds/7.0.UP01-19.09-gpu
```diff
@@ -0,0 +1,98 @@
+ abcpy-0.5.1-CrayGNU-19.09-python3.eb               --set-default-module
```
gppezzi: I suggest we remove abcpy from the jenkins list, unless someone is aware of production projects using it.

easybuild/easyconfigs/o/OpenImageDenoise/oidn-0.9.0-CrayGNU-19.09.eb
```diff
@@ -0,0 +1,20 @@
+easyblock = "Tarball"
+
+name = 'oidn'
```
teojgo: The directory of the package has to be renamed to `oidn` as https://github.com/eth-cscs/production/pull/1318/files to be consistent with the module name. Otherwise it is not found by EasyBuild.

easybuild/easyconfigs/o/OptiX/NVIDIA-OptiX-6.0.0-CrayGNU-19.09.eb
```diff
@@ -0,0 +1,22 @@
+# contributed by Jean Favre (CSCS)
+easyblock = "Tarball"
+
+name = 'OptiX'
```
teojgo: The name of the dir might have to be renamed here as well.

lucamar: I would rename the recipe instead as just `OptiX-6.0.0-CrayGNU-19.09.eb`, since the `name` of the software appearing inside the recipe is just `Optix`: what do you think, @jfavre ?

jgphpc:
> https://github.com/eth-cscs/production/pull/1319#issuecomment-532575354

Score-P-6.0* will be the new default in the production list for 19.0x:

Score-P-6.0-CrayCCE-19.06-classic.eb
Score-P-6.0-CrayCCE-19.06.eb
Score-P-6.0-CrayGNU-19.06.eb
Score-P-6.0-CrayIntel-19.06.eb
Score-P-6.0-CrayPGI-19.06.eb
all depend on: (CubeLib/4.4.4, CubeW/4.4.3 and CubeGUI/4.4.4) <-- keep the .eb

I need to rebuild `Score-P-6.0-CrayGNU-19.06-cuda-10.1.eb` with the same deps.
Should i make a PR for that ?

gppezzi:
We have 3 versions of boost, do we really need to keep all of them?

vkarak:
@lucamar I guess we should change also the paths at `jenkins/Machines.groovy`, shouldn't we?

lucamar:
> @lucamar I guess we should change also the paths at `jenkins/Machines.groovy`, shouldn't we?
Yes, I will update `prefixProduction` and `unusePath` for Dom in `jenkins/Machines.groovy` before triggering the build of the new software stack on Dom for `7.0UP01`.

lucamar:
> We have 3 versions of boost, do we really need to keep all of them?
The version is only `1.70.0`, but we have `versionsuffix` for `python2`, `python3` and without `python`.

PR #1320, opened 2019-09-18
[daint-mc] Addition of mc recipe

omlins:


omlins:
This seems to have failed due to the network problems yesterday.
Retest this please.

PR #1321, opened 2019-09-18
[daint] Add easyconfigs for ReFrame 2.17.2, 2.18 and 2.19 for Daint

vkarak:
This time I am updating also the `gpu` production list so as to remove ReFrame from `/apps/common` and I am using a separate install path. I am planning to put a symlink to the actual modules inside the `login/` folder as we do for `xalt`, but I am not sure how to make it point it  differently for Daint and Dom now.

### DIFF
--- a/easybuild/easyconfigs/r/reframe/reframe-2.17.2.eb
+++ b/easybuild/easyconfigs/r/reframe/reframe-2.17.2.eb
@@ -1,0 +1,31 @@
+# @author: gppezzi, vkarak, manitart
+
+easyblock = 'Tarball'
+
+name = 'reframe'
+version = '2.17.2'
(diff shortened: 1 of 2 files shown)


lucamar:
@vkarak If I remember correctly, `ReFrame` is built separately by Jenkins since it cannot appear in the production list: any software build from this list will load as `prereq` the module `daint-gpu`.

vkarak:
Thanks @lucamar. I will close this PR and get it done separately through Jenkins. I didn't know about this `prereq`. I will do a separate PR with a symlink under `login` as we do for XALT.

vkarak:
Reopened. I still need the recipe :-) I will just remove it from the production list.

PR #1325, opened 2019-09-19
[dom-gpu] Changing Paraview dependency to Boost 1.70.0

lucamar:
We removed Boost 1.67.0 in PE 19.09


lucamar:
Thanks for your comment @jfavre: we can remove the recipe as soon as 5.7 is released and drop the Boost dependency.

PR #1326, opened 2019-09-19
[dom-gpu] Removing old HPX releases

lucamar:
We keep only the latest `HPX-1.3.0-CrayGNU-19.09-APEX.eb` in production


jenkins-builds/7.0.UP01-19.09-gpu
```diff
@@ -16,8 +16,6 @@
  GSL-2.5-CrayCCE-19.09.eb
  GSL-2.5-CrayIntel-19.09.eb
  HDFView-2.14.eb                                    --set-default-module
- HPX-1.2.0-CrayGNU-19.09.eb                         --set-default-module
- HPX-1.3.0-CrayGNU-19.09.eb
  HPX-1.3.0-CrayGNU-19.09-APEX.eb
```
msimberg: I'm a bit late, but it would be good to keep the module without APEX as well (the one with APEX is essentially a debug/profiling build with small runtime overheads).

lucamar: Thanks for your comment @msimberg: would it then make sense to include the APEX build within the same installation of `HPX`? EasyBuild can manage multiple builds in the same recipe.

msimberg: Yes, I suppose it would. I didn't realize that was an option. Could you point me to an example recipe where this has been done?

lucamar: Please have a look at https://github.com/eth-cscs/production/pull/1331 and modify the branch directly if needed.

PR #1330, opened 2019-09-20
[dom] Update CrayCCE-19.09-classic toolchain

ekouts:


ekouts:
Should I also change the patches? For example
`patches = [ 'IPM-2.0.6-Cray-19.06.patch' ]` to`patches = [ 'IPM-2.0.6-Cray-19.09.patch' ]`

gppezzi:
I don't think that duplicating the patches is necessary if their content is the same.

> Should I also change the patches? For example in the file IPM-2.0.6-CrayIntel-19.09.eb change
> `patches = [ 'IPM-2.0.6-Cray-19.06.patch' ]` to
> `patches = [ 'IPM-2.0.6-Cray-19.09.patch' ]`?

vkarak:
retest this please

PR #1331, opened 2019-09-20
[dom] HPX with APEX for PE 19.10

lucamar:
The build should address both standard and APEX versions.


lucamar:
We might need to rename the executable if the apex build step uses the same name off the standard build.

msimberg:
Thanks @lucamar. The options look correct. Both builds definitely produce libraries and executables (examples) with the same names. What does Easybuild do with the two configurations? Does it produce two different modules? What will the names of the modules be?

lucamar:
EasyBuild will create a single modulefile, it will be up to us to rename and keep the `APEX` release separate from the standard `HPX` build. We use to do that for builds where the difference is minimal (e.g.: few files), but I see that for `HPX` we would need to keep separate folders. Furthermore, I'm not sure that we can use the CMake build type `RelWithDebInfo` for the standard build as well.

lucamar:
The `APEX` recipe used to work on Dom, but now it fails with the error:
```
-- Could NOT find PAPI (missing: PAPI_INCLUDE_DIR)
CMake Error at cmake/HPX_Message.cmake:47 (message):
  apex Requested APEX_WITH_PAPI but could not find PAPI.  Please specify
  PAPI_ROOT.
```

lucamar:
The variable `$PAT_BUILD_PAPI_BASEDIR` is not defined by `perftools-base`, therefore I try with `$(dirname $PAT_BUILD_PAPI_LIBDIR)`

msimberg:
@lucamar good point about the build type. `RelWithDebInfo` implies `-O2` which may make a difference. We haven't benchmarked to see if there is a significant difference between `-O2` and `-O3`. Was the idea to set `CMAKE_INSTALL_PREFIX`? Is there a way to get the prefix that EasyBuild *would* have used and just append something for the two variants?

lucamar:
@msimberg I have updated the recipe to the latest Cray PE on Dom (19.10) and added `-DCMAKE_INSTALL_PREFIX=%(installdir)s/APEX` for the `APEX` build, as you also suggested. 
I have also inserted in the module the environment variable `APEX_PATH` to get the path to the `APEX` build.

lucamar:
The `APEX` build is failing with this error in my local build (`/scratch/snx3000tds/lucamar/easybuild-HPX-1.3.0-20191025.151200.xDMkh.log`):
```
Overriding APEX git tag v2.1.3 with v2.1.3
-- Found Git: /usr/bin/git (found version "2.16.4")
-- git clone https://github.com/khuck/xpress-apex.git /run/user/21827/easybuild/build/HPX/1.3.0/CrayGNU-19.10/hpx-1.3.0/apex
-- apex : git checkout -q v2.1.3
-- apex Will build APEX
-- apex APEX source dir is /run/user/21827/easybuild/build/HPX/1.3.0/CrayGNU-19.10/hpx-1.3.0/apex/src/apex
-- apex APEX binary dir is /run/user/21827/easybuild/build/HPX/1.3.0/CrayGNU-19.10/easybuild_obj/apex/src/apex
System architecture: x86_64
fatal: Not a git repository (or any parent up to mount point /run/user)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

msimberg:
Thanks @lucamar. Should we maybe call the path to the APEX build `HPX_APEX_PATH` since it's fundamentally HPX which is there, not just APEX?

For the build failure I see that daint has an environment variable `PAT_BUILD_PAPI_BASEDIR` whereas dom only has `PAT_BUILD_PAPI_LIBDIR`. We can derive `PAPI_ROOT` from that, but is `..._BASEDIR` just missing on dom or is it not going to be available there? Most importantly, is it going to be available when daint is upgraded?

lucamar:
> Thanks @lucamar. Should we maybe call the path to the APEX build `HPX_APEX_PATH` since it's fundamentally HPX which is there, not just APEX?
@msimberg I have changed the name of the variable as you suggested.

> For the build failure I see that daint has an environment variable `PAT_BUILD_PAPI_BASEDIR` whereas dom only has `PAT_BUILD_PAPI_LIBDIR`. We can derive `PAPI_ROOT` from that, but is `..._BASEDIR` just missing on dom or is it not going to be available there?
I'm pretty sure that `PAT_BUILD_PAPI_BASEDIR` was also available on Dom sometime ago, since the recipe was working. Now only `PAT_BUILD_PAPI_LIBDIR` is defined, therefore I have adapted the script.

> Most importantly, is it going to be available when daint is upgraded?
If the recipe works, yes.

lucamar:
@msimberg The build was successful, therefore I have also modified the list of builds that will be processed as soon as the pull request will be merged, in order to include the current recipe of HPX. Once you approve it, we'll merge it and the module will be available on Dom.

vkarak:
retest this please

vkarak:
I've re-triggered CI, since this is meant for the mc nodes as well.

PR #1333, opened 2019-09-20
[dom] removing modules from the cray TCs

gppezzi:


gppezzi:
This is not needed anymore, problem originally came from a system glitch and EB 4.0.0 seems to include a workaround for the issues we've seen with non-default Cray modules.

PR #1337, opened 2019-09-24
[daint-gpu] Create jupyterlab-1.1.4-CrayGNU-18.08.eb

twrobinson:
JupyterLab 1.1.4 with julia 1.2


twrobinson:
I was goijng to close this but then I noticed it is leet, so I'm going to keep it open

twrobinson:
Won't commit

PR #1338, opened 2019-09-25
PyCUDA: Fix for eb 4 compatibility

gppezzi:
This is to ensure that the pycuda easyblock is compatible with EB4 as well with older EB versions.

However we could potentially delete this custom easyblock (it seems that pycuda now builds with stock version, but I didn't test it).

Thanks @boegel !


boegel:
Context for this change: https://easybuild.readthedocs.io/en/develop/EasyBuild4-overview-of-changes.html#ingested-functionality-from-vsc-base-and-vsc-install ;)

PR #1340, opened 2019-09-26
[WIP] [dom-gpu] PoC CrayGNU 19.09 toolchain with CDT

gppezzi:
PoC using defaults from CDT. 

Goal is to minimise the changes needed in the `EXTERNAL_MODULES` when deploying a software stack with a new toolchain version. 

The idea is that, for a specific cdt, only dependencies requiring non-default versions should be pinned in the easyconfigs. By non-default I mean w.r.t the cdt defaults, not to be confused with the system defaults

Here's how to test:

```
export EASYBUILD_EXTERNAL_MODULES_METADATA=..../production/easybuild/cray_external_modules_metadata.cfg-19.09
```
Caveats so far:
- it would require one metadata file per CDT version
- module load is quite verbose (when it loads `cdt`)


gppezzi:
CDT 19.10 was released, following up in #1367

PR #1345, opened 2019-09-27
[daint-gpu] Update recipe for SIRIUS/6.4.0

toxa81:
Now we depend on SpFFT library, which EB script is also pushed with this commit.


toxa81:
test this please

toxa81:
Job seems to run out of time.

PR #1353, opened 2019-10-04
[dom-gpu] COSMO crClim - Update netCDF version in prebuildopts

kraushm:


gppezzi:
I'm merging since the CI will fail because of the known problem of using non-default modules (this build requires `module load cdt/19.08`)

PR #1358, opened 2019-10-09
[daint dom kesch] Add cpus-per-task for srun commands

teojgo:
Remove also fulen from the supported Machines


teojgo:
> could you please mention in the pr (title or comment) that this also removes fulen

done

PR #1368, opened 2019-10-11
[daint-gpu] Add support to hide hidden dependencies and exit on first error

victorusu:
Add option 
* `--hide-deps`, which adds `--hidden` to command line for easybuild in case of a hidden dependency recipe.
* `--exit-on-error` which exists the `production.sh` script on the first `eb` failure

Both options should be used in the TestingEB project. 
In this case, the CI will 
* build hidden dependencies only once
* exit the build procedure in case one of the recipes do not succeed.

Added GROMACS and Bison to the PR just to test the effects.

### DIFF
--- a/jenkins-builds/production.sh
+++ b/jenkins-builds/production.sh
@@ -165,7 +165,7 @@ if [ -n "${eb_lists}" ] && [ -n "${hidden_deps}" ]; then
   __eb_list=`eb --show-full-config | grep -i hide | awk -F'=' '{print $2}' | head -1`
   IFS=', ' read -r -a hidden_deps <<< ${__eb_list}
 
-# match force_list items with production lists: only matching items will be built using the EasyBuild flag '-f'
+# match  items with hide deps list: matching items will be built using the EasyBuild flag '--hidden'
  echo -e "Items matching hidden list and easybuild recipes to install (\"${eb_lists}\")"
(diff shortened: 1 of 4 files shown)


jenkins-builds/production.sh
```diff
@@ -10,18 +10,20 @@ scriptdir=$(dirname $0)
...
-    -u,--unuse    Module unuse colon separated PATH  (optional: default is null)
-    -x,--xalt     [yes|no] update XALT database      (optional: default is yes)
+    -a,--arch         Architecture (gpu or mc)           (mandatory: Dom and Piz Daint only)
+    -f,--force        Force build of item(s) in list     (optional: double quotes for multiple items)
+    -h,--help         Help message
+    -l,--list         Absolute path to production file   (mandatory: EasyBuild production list)
+    -p,--prefix       Absolute path to EasyBuild prefix  (mandatory: installation folder)
+    -u,--unuse        Module unuse colon separated PATH  (optional: default is null)
+    -x,--xalt         [yes|no] update XALT database      (optional: default is yes)
+    --hide-deps       Hide the hidden dependencies
```
gppezzi: if we call `eb file.eb`, then `file.eb` is not actually a dependency. How about:

- Force hide modules listed in 'hide-deps' (TestingEB only)

PR #1369, opened 2019-10-11
[dom] Add files for PE19.10

ekouts:
Add easyconfigs for cdt/19.10
Fix deprecated easyconfigs
Add metadata for cdt/19.10
Add jenkins files

### DIFF
--- a/easybuild/cray_external_modules_metadata-19.10.cfg
+++ b/easybuild/cray_external_modules_metadata-19.10.cfg
@@ -51,12 +51,12 @@ name = PETSc
 version = 3.11.2.0
 prefix = PETSC_DIR
 
-[cray-python2]
+[cray-python/2.7.15.7]
 name = Python
 version = 2.7.15.7
 prefix = /opt/python/2.7.15.7
 
-[cray-python3]
+[cray-python/3.6.5.7]
 name = Python
 version = 3.6.5.7
 prefix = /opt/python/3.6.5.7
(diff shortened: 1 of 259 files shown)


ekouts:
So that we don't include the version of an external dependencies when it's the standard.

vkarak:
@lucamar We've discussed this on Friday with @gppezzi and said that it would be better to have different metadata files for each cdt, so as not to grow a single file.

ekouts:
But actually the pull request should not be merged yet because I haven't changed the external dependencies for cray-python in many files. Will do it later on the afternoon.

lucamar:
@ekouts and @vkarak: I agree that it could be better to separate metadata and avoid appending data in a single file. 
However, we will need to have a different metadata file for each system and then export the corresponding variable before loading the EasyBuild custom modulefile, see #1370. 
While it is relatively easy for us to update the variable within the script `production.sh` at each PE update, it might be less immediate for users that load manually EasyBuild on the systems.

gppezzi:
Main goal of having one metadata file per cdt version is to simplify our recipes and allow using versionless external modules as dependency. The cdt module was not available when cray support was designed in EB, now that we have it I think we should try to rely on it in order to optimize our upgrade workflow.

lucamar:
> lgtm. I think we can merge this, can't we?
Yes, we need to merge it to have the correct metadata on Dom after the change of #1370

PR #1370, opened 2019-10-11
[dom-gpu] Production metadata dom

ekouts:
Change the metadata file for dom on production.sh in order for jenkins to install the new PE

### DIFF
--- a/jenkins-builds/production.sh
+++ b/jenkins-builds/production.sh
@@ -159,7 +159,7 @@ fi
 # --- BUILD ---
 # load module EasyBuild-custom
 module load EasyBuild-custom/cscs
-if [[ “$system” =~ “dom” ]]; then
+if [[ "$system" =~ "dom" ]]; then
  export EASYBUILD_EXTERNAL_MODULES_METADATA=${EB_CUSTOM_REPOSITORY}/cray_external_modules_metadata.cfg-19.10
 fi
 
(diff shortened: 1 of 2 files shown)


lucamar:
@ekouts Why do we need a custom metadata file for Dom with CrayPE 19.10? 
Adding the new modules to the current metadata file is not enough?

lucamar:
retest this please

jenkins-builds/production.sh
```diff
@@ -159,6 +159,9 @@ fi
 # --- BUILD ---
 # load module EasyBuild-custom
 module load EasyBuild-custom/cscs
+if [[ "$system" =~ "dom" ]]; then
+ export EASYBUILD_EXTERNAL_MODULES_METADATA=${EB_CUSTOM_REPOSITORY}/cray_external_modules_metadata.cfg-19.10
```
vkarak: Since `cfg` is the extension of the file, I would write the filename as `cray_external_modules_metadata-19.10.cfg`. Perhaps my comment should have been to another PR, but I have just spotted this.

lucamar:
@ekouts Considering the comments of #1369 as well, I would suggest to leave `production.sh` as it is and modify `https://github.com/eth-cscs/production/blob/master/easybuild/module/EasyBuild-custom/cscs` instead at line 49 as follows:
```
if [[ "$system" =~ "dom" ]]; then
 setenv EASYBUILD_EXTERNAL_MODULES_METADATA $::env(EB_CUSTOM_REPOSITORY)/cray_external_modules_metadata-19.10.cfg
else
 setenv EASYBUILD_EXTERNAL_MODULES_METADATA $::env(EB_CUSTOM_REPOSITORY)/cray_external_modules_metadata.cfg
```

PR #1378, opened 2019-10-21
[daint dom kesch] Add cray-mpich after module purge in production.sh

teojgo:
* Also remove loading of additional cray modules.

### DIFF
--- a/jenkins-builds/production.sh
+++ b/jenkins-builds/production.sh
@@ -113,6 +113,8 @@ if [[ "$system" =~ "daint" || "$system" =~ "dom" ]]; then
         echo -e "\n No architecture defined. Please use the option -a,--arch to define the architecture \n"
         usage
     else
+        module purge
+        module load craype craype-network-aries modules perftools-base ugni cray-mpich
         module load daint-${ARCH}
         eb_args="${eb_args} --modules-header=${scriptdir%/*}/login/daint-${ARCH}.h --modules-footer=${scriptdir%/*}/login/daint.footer"
     fi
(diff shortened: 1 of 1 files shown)


teojgo:
> @teojgo If you remove the `module purge` command, we will have problems building `daint-gpu` and `daint-mc` one after the other, unless we log out and log in again

@lucamar the problem is that currently the build fails since somehow mpich cannot be found

lucamar:
@teojgo Let me then update the list of modules to load after the `module purge` command, rather than removing it

lucamar:
@teojgo I have just added `cray-mpich` to the list of modules after the `module purge` command: this was not required on Piz Daint (see "Currently Loaded Modulefiles" in  https://jenkins.cscs.ch/blue/rest/organizations/jenkins/pipelines/ProductionEB/runs/915/nodes/34/steps/60/log/?start=0) and not even on Dom up to Cray PE 19.09, but it seems necessary for Cray PE 19.10 (the "Currently Loaded Modulefiles" are the same in https://jenkins.cscs.ch/blue/rest/organizations/jenkins/pipelines/ProductionEB/runs/915/nodes/35/steps/68/log/?start=0, but the builds fail).

I have tested the change on Dom and it works for Boost (it fails without the `cray-mpich`):
```
for ARCH in gpu mc; do
 /users/lucamar/GitHub/eth-cscs/production/jenkins-builds/production.sh --arch=$ARCH --list=/scratch/snx3000tds/lucamar/7.0.UP01-19.10-$ARCH --prefix=$SCRATCH/$ARCH/easybuild
done
```

The change should be backward compatible with Piz Daint.

PR #1391, opened 2019-11-06
Update Machines.groovy and soft links to latest cle7 for daint

teojgo:


jenkins/Machines.groovy
```diff
@@ -2,10 +2,10 @@ def daint = [name: 'daint',
              archs: ['gpu', 'mc'],
              cpus: ['gpu': 12, 'mc': 36],
              buildPath: '$XDG_RUNTIME_DIR/easybuild/build',
-             unusePath: '/apps/daint/UES/jenkins/6.0.UP07/ARCH/easybuild/tools/modules/all:/apps/daint/UES/jenkins/6.0.UP07/ARCH/easybuild/modules/all',
+             unusePath: '/apps/daint/UES/jenkins/7.0.UP01/ARCH/easybuild/tools/modules/all:/apps/daint/UES/jenkins/7.0.UP01/ARCH/easybuild/modules/all',
```
gppezzi: since we're currently using the software stack from Dom, I assume we need to unuse that path or else nothing will be built. This is what we have in daint-gpu for example:
```
module		 use /apps/dom/UES/jenkins/7.0.UP01/gpu/easybuild/modules/all
module		 use /apps/dom/UES/jenkins/7.0.UP01/gpu/easybuild/tools/modules/all
```

teojgo: For the build the first time I can pass it as a parameter to the jenkins build

PR #1395, opened 2019-11-06
[dom] Minor upgrade of VisIt for CLE19.10

jfavre:
== COMPLETED: Installation ended successfully (took 44 min 13 sec)
== Results of the build can be found in the log file(s) /scratch/snx3000tds/jfavre/daint/software/Visit/3.0.2-CrayGNU-19.10/easybuild/easybuild-Visit-3.0.2-20191106.125955.log

### DIFF
--- a/easybuild/easyconfigs/v/Visit/Visit-3.0.2-CrayGNU-19.10.eb
+++ b/easybuild/easyconfigs/v/Visit/Visit-3.0.2-CrayGNU-19.10.eb
@@ -1,0 +1,109 @@
+# contributed by Jean Favre (CSCS)
+#
+easyblock = 'CMakeMake'
+
+name = 'Visit'
+version = "3.0.2"
+
+homepage = 'https://wci.llnl.gov/codes/visit/'
+description = """VisIt is an Open Source, interactive, scalable, visualization,
+animation and analysis tool."""
+
+toolchain = {'name': 'CrayGNU', 'version': '19.10'}
+toolchainopts = {'usempi': True, 'verbose': False}
+
+#
+# Visit downloads all its dependencies by itself during the configuration.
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/v/Visit/Visit-3.0.2-CrayGNU-19.10.eb
```diff
@@ -0,0 +1,109 @@
...
+pyshortver = '%s.%s' % (py_maj_ver, py_min_ver)
+pysuff = '-python%s' % py_maj_ver
+
+dependencies = [
+    ('Boost', '1.70.0', pysuff),
+    ('cray-python/%s' % pyver, EXTERNAL_MODULE)
+]
+
+builddependencies = [
+    ('CMake', '3.9.3', '', True),
```
gppezzi: Is it a problem to use the CMake already available in the system?
```suggestion
    ('CMake', '3.14.5', '', True),
```

jfavre: Yes, it is a problem. VisIt will not configure itself. There is no choice.

jfavre:
the failure to build is seen only on the compute node. I have just redone a build on the login node. Investigating...Failure is with the building of Qt

jfavre:
the failure is due to the configuration of Qt detecting - rightly - because we are on a compute node:

**Checking for Desktop OpenGL... no**

However, the GUI runs on a desktop node. It is a catch 22 situation. We're compiling on a compute node something that will never run on a compute node.

jfavre:
this is for an older version. Version 3.1.0 is currently installed.

PR #1396, opened 2019-11-06
[daint-gpu dom-gpu] Magma-2.5.1 with CrayIntel-19.10

toxa81:


easybuild/easyconfigs/m/magma/magma-2.5.1-CrayIntel-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,37 @@
...
+toolchain = {'name': 'CrayIntel', 'version': '19.10'}
+toolchainopts = {'pic': True}
+
+sources = [SOURCE_TAR_GZ]
+source_urls = ['http://icl.cs.utk.edu/projectsfiles/magma/downloads/']
+patches = [('make.inc-2.5.1', './')]
+#dependencies = [('craype-accel-nvidia60', EXTERNAL_MODULE)]
+
+builddependencies = [
+  ('cudatoolkit', EXTERNAL_MODULE)
```
gppezzi: We suggest to pin the cuda version (see comments in #1404 )

toxa81: done

PR #1404, opened 2019-11-07
[daint-gpu] SpFFT: new version 0.9.8 for gnu/intel

simonpintarelli:


gppezzi:
@simonpintarelli could you please mention the target system in the PR subject, so that the CI can test it? Thanks in advance.

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,36 @@
...
+toolchain = {'name': 'CrayGNU', 'version': '19.10'}
+toolchainopts = {'usempi': True, 'openmp': True, 'pic': True, 'verbose': False, 'opt': True}
+
+source_urls = ['https://github.com/eth-cscs/SpFFT/archive']
+sources = ['v%(version)s.tar.gz']
+
+builddependencies = [('CMake', '3.14.5', '', True)]
+
+dependencies = [
+  ('cudatoolkit' , EXTERNAL_MODULE),
```
gppezzi: We updated the EB toolchains and, since Cray PE 19.10, you don't need to pin the version for the modules listed in the cray `cdt/19.10` module.

However,  `cudatoolkit` is not in the list and therefore we recommend to pin the cudatoolkit version here (for reproducibility, to ensure that this recipe will work when we change the cuda defaults on the system).

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayIntel-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,36 @@
...
+toolchain = {'name': 'CrayIntel', 'version': '19.10'}
+toolchainopts = {'usempi': True, 'openmp': True, 'pic': True, 'verbose': False, 'opt': True}
+
+source_urls = ['https://github.com/eth-cscs/SpFFT/archive']
+sources = ['v%(version)s.tar.gz']
+
+builddependencies = [('CMake', '3.14.5', '', True)]
+
+dependencies = [
+  ('cudatoolkit' , EXTERNAL_MODULE),
```
gppezzi: same as for the CrayGNU version (pin version here)

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,36 @@
...
+  ('gcc', EXTERNAL_MODULE),
```
gppezzi: `gcc` is already loaded by the CrayGNU toolchain, is this really needed?

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,36 @@
...
+  ('intel/19.0.1.144', EXTERNAL_MODULE),
```
gppezzi: Suggestion: since you are adding intel to a CrayGNU recipe, maybe you could add a hint in the versionsuffix (by adding for example a `-intel`).

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
simonpintarelli: I've removed it.

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayIntel-19.10-cuda-10.1.eb
simonpintarelli: Done.

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
simonpintarelli: Ok.

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
simonpintarelli: Thanks. I've added intel (it is just for mkl) in the version suffix.

gppezzi: You could also call use `-mkl` if you prefer. 

The last remark I have is that the file name should match the full version string (so that this recipe can be used as a dependency in another easyconfig), for example:

```
SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1-intel.eb
```

easybuild/easyconfigs/s/SpFFT/SpFFT-0.9.8-CrayGNU-19.10-cuda-10.1.eb
toxa81: So, then you need to have different EB recipes for each version of CUDA?

gppezzi: It was always the case to have different recipes for different versions of cuda: this is how reproducibility is guaranteed, we cannot rely on system defaults as they might change over time and break the existing recipes. 

The only thing we changed recently is that we include the cdt module in the toolchains, so that you don't need to explicitly pin (most of) the Cray dependency versions in the recipes and you will automatically get the default version from the specific cdt. The defaults for a specific cdt are defined by Cray and they don't change over time (this should not be confused with the system defaults: loading cdt will override system defaults).

The question 'what is special about cudatoolkit' should be asked to Cray, since they don't define a default for cuda in the cdt module (see for example `/opt/cray/pe/modulefiles/cdt/19.10).

We missed this detail when we deployed the 19.10 software stack (that's why you can see versionless dependencies on cudatoolkit in other recipes), but we'll fix that soon.

PR #1405, opened 2019-11-07
[daint-gpu, dom-gpu] adds a specific library path to fix ray-racing

jfavre:


jfavre:
it is a fix for https://jira.cscs.ch/browse/UES-601

gppezzi:
@jfavre since you are updating this easyconfig: we've noticed that this package is being installed with non-standard permissions (not writable for the user installing it).

could you please address this in this PR as well? You could try something like this (I didn't test, it is just a suggestion):
```
postinstallcmds = [
    ("chmod u+w -R %(installdir)s/include"),
    ("chmod u+w -R %(installdir)s/linux-x86-64"),
```

PR #1408, opened 2019-11-07
Update spack files

victorusu:


gppezzi:
Please only merge after the OK from @victorusu as it seems that there are still issues (and we don't have a CI test for this yet).

victorusu:
Problem fixed!

victorusu:
Ready to be merged!

PR #1414, opened 2019-11-14
[daint] DeePMD-0.10.2 CrayGNU 19.10

kraushm:
Based on ticket https://webrt.cscs.ch/Ticket/Display.html?id=37633


kraushm:
retest this please

PR #1418, opened 2019-11-15
[daint-gpu] Amber

victorusu:


victorusu:
@lucamar, I cannot because in this case the user will not have access to it... He cannot build Amber for himself.

lucamar:
Thanks @victorusu: then I would suggest to rename the recipe and the modulefile, providing a short description of the new features in the `module help`.

victorusu:
@lucamar, can we merge this one?

PR #1426, opened 2019-11-25
[daint-gpu] Fix GROMACS 2018.6 PLUMED-2.5.1 recipe

victorusu:


vkarak:
@victorusu There are a couple of warnings issued during building. Are we fine with them?

victorusu:
@vkarak, the warnings are related to Jenkins itself and the Haswell build for intel compilers on EB.

victorusu:
I will make a PR on EB to hide that warning for Cray

vkarak:
@kraushm Can you also approve?

gppezzi:
IMO the  versionless `[cudatoolkit]` entry in the metadata file should not be kept, as it does not ensure reproducibility as we change the system defaults. 

I'll fix that in #1427, unless you have a reason for keeping it?

vkarak:
Yep, it is not needed. I merged this as is, because otherwise we had to wait for #1427.

PR #1427, opened 2019-11-25
[daint dom] Pin cudatoolkit version on the metadata file 19.10

gppezzi:
`cudatoolkit` version is not defined in the cdt module and therefore we need to pin the version here and update the recipes for 19.10 accordingly.


gppezzi:
I'm skipping CI since intel compiler is slow and most of the builds already went through

PR #1428, opened 2019-11-25
[daint dom] Remove fftw from metadata (unused)

ekouts:
From what I can tell this module is not used in any production recipe, but fftw/2.1.5.9 is available on daint. If the entry is not useful I can remove it completely from the metadata file.


vkarak:
Hi @ekouts, by "any production recipe" do you mean any recipe file in the repo or just those that they are in the production list?

ekouts:
I mean any recipe in the repo, I didn't find any recipe with fftw/2.1.5.9.

vkarak:
@ekouts We could remove it then completely.

PR #1436, opened 2019-11-27
[daint-mc] OpenFOAM 6.0 + dependency SCOTCH

kraushm:
Build to support https://webrt.cscs.ch/Ticket/Display.html?id=37504


kraushm:
retest this please

PR #1439, opened 2019-11-28
[dom] Create Buildah recipe

gppezzi:
This only works on a particular node on Dom for the moment (CI will most likely fail)

### DIFF
--- a/easybuild/easyconfigs/b/Buildah/Buildah-1.11.5.eb
+++ b/easybuild/easyconfigs/b/Buildah/Buildah-1.11.5.eb
@@ -3,6 +3,7 @@ easyblock = 'Tarball'
 name = 'Buildah'
 version = '1.11.5'
 local_go_version = '1.13.4'
+local_runc_version = '1.0.0-rc9'
 
 homepage = 'https://buildah.io/'
 description = """Buildah is a tool that facilitates building OCI container images."""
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/b/Buildah/Buildah-1.11.5.eb
```diff
@@ -0,0 +1,38 @@
...
+
+postinstallcmds = ['mkdir %(builddir)s/go_path && export GOPATH=%(builddir)s/go_path && '
+                   '%(installdir)s/bin/go get github.com/opencontainers/runc && '
+                   '%(installdir)s/bin/go get github.com/containers/buildah ']
+
+# Make-install runc v1.0.0-rc9
+postinstallcmds += [
+    'cd %(builddir)s/go_path/src/github.com/opencontainers/runc && '
+    'export PATH=%(installdir)s/bin/:$PATH && export GOPATH=%(builddir)s/go_path && '
+    'git checkout v1.0.0-rc9 && make BUILDTAGS="" && make install PREFIX=%(installdir)s']
```
victorusu: Does it make sense to define a `local_` var for the open container and set it a the beginning of the recipe, like you did for go?

teojgo: Is it possible to mix `%()s` type strings and the python formated ones?

victorusu: yes. You can pass a dictionary at the end, which makes the mapping...

PR #1443, opened 2019-12-02
[daint] COSMO-POMPA-crclim cordex recipes

omlins:


easybuild/easyconfigs/c/COSMO_pompa/COSMO_pompa-f9916ba-28.11.19-CrayCCE-19.08-cordex-double.eb
```diff
@@ -0,0 +1,49 @@
+# kraushm, omlins (CSCS)
+easyblock = 'MakeCp'
+
+name = 'COSMO_pompa'
+version = 'f9916ba'
+versionsuffix = '-28.11.19-cordex-double'
```
kraushm: Do we mind, that the version suffix of the module is not equal the file name?

easybuild/easyconfigs/c/COSMO_pompa/COSMO_pompa-f9916ba-28.11.19-CrayCCE-19.08-cuda-10.0-cordex-double.eb
```diff
@@ -0,0 +1,51 @@
+# kraushm, omlins (CSCS)
+easyblock = 'MakeCp'
+
+name = 'COSMO_pompa'
+version = 'f9916ba'
+versionsuffix = '-28.11.19-cuda-10.0-cordex-double'
```
kraushm: Same comment regarding filename not equal module name.

easybuild/easyconfigs/c/COSMO_pompa/COSMO_pompa-f9916ba-28.11.19-CrayCCE-19.08-cordex-double.eb
omlins: Yes, I agree that it is not optimal, right now. As discussed offline, I think it logically rather belongs to the version (it is the essentially the same information as the hash) as Guilherme originally suggested. In order to avoid changing all the filenames that depend on the version, we can split create local variables for the hash and for the date and use them for the filenames... I suggest to apply these changes however in the scope of a general cleanup when creating the PE 19.11 recipes.

kraushm: Doing this "cleanup" with PE19.11 is fine with me.

easybuild/easyconfigs/c/COSMO_pompa/COSMO_pompa-f9916ba-28.11.19-CrayCCE-19.08-cuda-10.0-cordex-double.eb
kraushm: See discussion for the CPU version...

PR #1451, opened 2019-12-12
[dom-gpu] Add support for XALT installation in production.sh

victorusu:
One cannot have xalt loaded before installing xalt...

### DIFF
--- a/easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
+++ b/easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
@@ -1,0 +1,68 @@
+# contributed by Victor Holanda (CSCS)
+
+name = 'xalt'
+version = "2.7.24"
+
+homepage = 'https://xalt.readthedocs.io/'
+description = """XALT is a lightweight software tool for any Linux cluster,
+workstation, or high-end supercomputer to track executable information and
+linkage of static shared and dynamically linked libraries. When the code is
+executed, wrappers intercept both GNU linker (ld) to capture linkage
+information and environmental variables."""
+
+toolchain = SYSTEM
+
+source_urls = ['https://github.com/xalt/xalt/archive/']
+sources = [SOURCELOWER_TAR_GZ]
+
+patches = [('Daint_config.py', 'Config/')]
+
+local_transmission = "curl"
+
+# generic configurations
+configopts  = "--with-trackScalarPrgms=yes "
+configopts += "--with-computeSHA1=no "
+configopts += "--with-MySQL=no "
+configopts += "--with-tmpdir=${HOME}/.xalt.d/tmpdir "
+# system specific configurations
+configopts += "--with-trackGPU=yes "
+configopts += "--with-systemPath=/apps/daint/system/bin:/apps/dom/system/bin:/apps/common/system/bin::/usr/local/bin:/usr/bin:/bin:/usr/bin/X11 "
+configopts += "--with-config=Config/Daint_config.py "
+configopts += "--with-syshostConfig=read_file:/etc/xthostname "
+configopts += "--with-transmission=%s " % local_transmission
+# this is workaround to do not make public the logging_url data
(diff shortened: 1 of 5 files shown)


easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
```diff
@@ -0,0 +1,68 @@
...
+local_transmission = "curl"
+
+# generic configurations
+configopts  = "--with-trackScalarPrgms=yes "
+configopts += "--with-computeSHA1=no "
+configopts += "--with-MySQL=no "
+configopts += "--with-tmpdir=${HOME}/.xalt.d/tmpdir "
+# system specific configurations
+configopts += "--with-trackGPU=yes "
+configopts += "--with-systemPath=/apps/daint/system/bin:/apps/dom/system/bin:/apps/common/system/bin::/usr/local/bin:/usr/bin:/bin:/usr/bin/X11 "
```
jgphpc: can we avoid to depend on /system ?

easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
```diff
@@ -0,0 +1,68 @@
...
+preinstallopts = 'LDFLAGS="-L${EBROOTCUDA}/lib64" CPPFLAGS="-I${EBROOTCUDA}/include" '
+
+modextravars = {
+    'XALT_EXECUTABLE_TRACKING': 'yes',
+    'XALT_SCALAR_SAMPLING': 'no',
+    'XALT_GPU_TRACKING': 'no',
+    'XALT_DIR': '%(installdir)s',
+    'XALT_ETC_DIR': '%(installdir)s/etc',
+    'XALT_TRANSMISSION_STYLE': '%s' % local_transmission,
+    #'XALT_LOGGING_URL': 'asdfasdfasdf',
```
jgphpc: ?

easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
```diff
@@ -0,0 +1,68 @@
...
+configopts += "--with-tmpdir=${HOME}/.xalt.d/tmpdir "
+# system specific configurations
+configopts += "--with-trackGPU=yes "
+configopts += "--with-systemPath=/apps/daint/system/bin:/apps/dom/system/bin:/apps/common/system/bin::/usr/local/bin:/usr/bin:/bin:/usr/bin/X11 "
+configopts += "--with-config=Config/Daint_config.py "
+configopts += "--with-syshostConfig=read_file:/etc/xthostname "
+configopts += "--with-transmission=%s " % local_transmission
+# this is workaround to do not make public the logging_url data
+configopts += "--with-xaltLoggingUrl=`cat /apps/common/easybuild/sources/x/xalt/logging_url.txt` "
+# this is a limitation on EB
```
jgphpc: can you explain more ? why is it needed if gpu_tracking is off ?

easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
```diff
@@ -0,0 +1,68 @@
...
+
+modextrapaths = {
+    'LD_PRELOAD': 'lib64/libxalt_init.so',
+    'COMPILER_PATH': 'bin'
+}
+
+builddependencies = [
+    ('cudatoolkit/10.1.105_3.27-7.0.1.1_4.1__ga311ce7', EXTERNAL_MODULE),
+]
+dependencies = [
```
jgphpc: dependencies list is also a builddep no (hence builddep is not needed) ?

easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
victorusu: Unfortunately, no!

easybuild/easyconfigs/x/xalt/xalt-2.7.24.eb
victorusu: We do not need to define here, but I kept because it is part of our investigation.

PR #1452, opened 2019-12-12
[dom] Create jupyterlab-1.1.1b-CrayGNU-19.10.eb

twrobinson:
With the coming update of batchspawner to the asyncio version there is a new dependency of some compute-side batchspawner components.


twrobinson:
I have labelled this as "b", as all components of JuptyerLab are the same apart from the addition of batchspawner components which are needed for the new batchspawner. if you think there is a better naming convention let me know.

kraushm:
I would probably just add `-batchspawner` as module suffix, i. e. jupyterlab-1.1.1-CrayGNU-19.10-batchspawner.eb. Adding the `b` to the version risks to become confusing at some point.

twrobinson:
renamed the pr to avoid confusion

twrobinson:
retest this please

PR #1473, opened 2020-01-10
[daint] Mpi4py

kraushm:
Adding mpi4py 3.0.3 without CUDA for ticket: https://webrt.cscs.ch/Ticket/Display.html?id=37549


easybuild/easyconfigs/m/mpi4py/mpi4py-3.0.3-CrayGNU-19.10-python3.eb
```diff
@@ -0,0 +1,34 @@
...
+
+# force rebuilding everything, including patched files
+buildopts = '--force'
+
+req_py_majver = '3'
+req_py_minver = '6'
+
+sanity_check_paths = {
+    'files': [],
+    'dirs': ['lib/python3.6/site-packages/%(name)s'],
```
ekouts: I think this is better to prevent possible mistakes in automatically updated easyconfigs for future upgrades of the system.
```suggestion
    'dirs': ['lib/python%s.%s/site-packages/mpi4py' % (req_py_majver, req_py_minver)],
```
If you could keep all three variables even better.

kraushm: Used slightly different solution to keep all 3 variables.

PR #1482, opened 2020-01-16
[dom] Create jupyterlab-1.2.4-CrayGNU-19.10-batchspawner.eb

twrobinson:
Testing the compatibility of "JupyterLab GPU Dashboards" from the RAPIS folks at https://github.com/rapidsai/jupyterlab-nvdashboard.


twrobinson:
Demonstrated with a simple CUBLAS dgemm. @teojgo and/or @rsarm would you like to try some cupy, numba, etc when the machines come back up after the maintenance?

PR #1485, opened 2020-01-21
[daint-gpu] magma-2.5.2 with GNU

toxa81:
* recipe for magma-2.5.2 with GNU 
* as agreed with @gppezzi , I'm moving deprecated magma recipes to a 'deprecated' folder


easybuild/easyconfigs/m/magma/magma-2.5.2-CrayGNU-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,48 @@
...
+
+builddependencies = [
+  ('CMake', '3.14.5', '', True),
+  ('cudatoolkit/10.1.105_3.27-7.0.1.1_4.1__ga311ce7', EXTERNAL_MODULE),
+  ('intel/19.0.1.144', EXTERNAL_MODULE)
+]
+
+dependencies = [
+  ('cudatoolkit/10.1.105_3.27-7.0.1.1_4.1__ga311ce7', EXTERNAL_MODULE),
+  ('intel/19.0.1.144', EXTERNAL_MODULE)
```
gppezzi: Is this really using CrayGNU? It looks ambiguous by looking at the recipe: if users build software on top of this 'CrayGNU' module, they could end up using intel without noticing? 

you are often mixing both gcc and intel compilers in these recipes, maybe we should consider to create a specific toolchain?

toxa81: We want do compile with GNU, but link with MKL. MKL gives the best performance for linear algebra. For the second comment, we need to have a fresh gcc for Intel, otherwise it doesn't compile modern C++ code. So, we have to manually load gcc/8.2.0

gppezzi: In this case, is it a problem then to add `-mkl` to the versionsuffix?

Can you rename the PR to match the version (2.6.2 -> 2.5.2)?

As for the need of gcc when using intel, I think that should be included in the toolchain (but we'll take that offline since this is unrelated to this PR).

gppezzi:
retest this please

gppezzi:
@toxa81 CI script has been updated in the master branch in order to filter out recipes under the 'deprecated' folder, can you please update your branch?

easybuild/easyconfigs/m/magma/magma-2.5.2-CrayGNU-19.10-cuda-10.1-mkl.eb
```diff
@@ -0,0 +1,48 @@
+# created by Anton Kozhevnikov (CSCS)
+
+easyblock = 'CMakeMake'
+
+name = 'magma'
+version = '2.5.2'
+versionsuffix = '-cuda-10.1'
```
gppezzi: The versionsuffix should match the file name, please add `-mkl` here as well

easybuild/easyconfigs/s/SIRIUS/SIRIUS-6.4.4-CrayGNU-19.10-cuda-10.1.eb
```diff
@@ -24,7 +24,7 @@ dependencies = [
   ('ELPA','2019.05.001'),
   ('cray-hdf5', EXTERNAL_MODULE),
   ('cudatoolkit/%s.105_3.27-7.0.1.1_4.1__ga311ce7' % cudaversion , EXTERNAL_MODULE),
-  ('magma', '2.5.1', '-cuda-%s' % cudaversion),
+  ('magma', '2.5.2', '-cuda-%s' % cudaversion),
```
gppezzi: `-mkl` should appear here as well?

toxa81: fixed!

toxa81:
retest this please

toxa81:
I don't quite understand what is wrong now?

easybuild/easyconfigs/m/magma/magma-2.5.2-CrayGNU-19.10-cuda-10.1-mkl.eb
gppezzi:
```suggestion
versionsuffix = '-cuda-10.1-mkl'
```

PR #1488, opened 2020-01-24
[WIP][daint] Addition of Julia kernel to jupyterlab

omlins:


omlins:
As discussed offline, we should do two recipies: one for the GPU partition and one for the MC partition. The current one is for the GPU partition.

twrobinson:
@omlins Just doing some clean up. This PR is no longer needed as we are focusing on 1.2.4 for the IJulia deployment, correct?

omlins:
Yes, we can close it.

PR #1491, opened 2020-01-27
[daint dom] Update HPX to 1.4.0

msimberg:


msimberg:
As a side note, there seems to be something off with the HPX module. It seems like the wrong hwloc package is loaded at runtime, resulting in applications not start correctly. I still have to look into that but will do so separately from this PR.

lucamar:
retest this please

lucamar:
@msimberg The build of `HPX-APEX` failed on Piz Daint with the CMAke configuration error below: 
```
CMake Error at apex/src/apex/CMakeLists.hpx:211 (hpx_libraries):
  Unknown CMake command "hpx_libraries".
```
See https://jenkins.cscs.ch/blue/organizations/jenkins/TestingEB/detail/TestingEB/3090/pipeline

msimberg:
Thanks @lucamar, that's a good catch. I've updated the PR to hopefully fix that. Let's see how it goes.

lucamar:
retest this please

lucamar:
Thanks @victorusu, the modules are in the production list: I will trigger a force build after merging the new recipes.

msimberg:
@victorusu yes, I keep forgetting... I'll add them to the list.

PR #1497, opened 2020-02-07
[dom-gpu] Adding jupyterlab-1.2.4 to jenkins list

twrobinson:
Adds IPython to make it available as a module. Adds jupyterlab-1.2.4 with support for IJulia kernel. Change default version of JupyterLab to one that is supported by new batchspawner.


lucamar:
retest this please

twrobinson:
> Thanks @twrobinson:
> 
>     * as you know, we already build `IPython` as dependency, but it is currently hidden since it appears in the `hide-deps` field of `easybuild/easybuild.d/system_wide.cfg`. If you plan to have a single release only, would it be simpler to un-hide it instead of having it as a separate entry?
> 
I plan to have other releases. 

>     * do you still need to build `jupyterlab-1.1.1-CrayGNU-19.10.eb`?

Good point. Removed. Thanks for reviewing

PR #1510, opened 2020-02-20
[dom-gpu] Add configuration for RH7.6 

kraushm:
This is the configuration part for the production list of Tsa.

Should be merged after the branch tsa76 of https://github.com/lucamar/production is merged.

PR #1511, opened 2020-02-20
WIP [tsa] Recipes used for RH7.5 and RH7.6

kraushm:
This PR contains all the recipes used for Tsa/Arolla for the different OS versions.


jenkins/Machines.groovy
```diff
@@ -1,22 +1,27 @@
...
 
 def dom = [name: 'dom',
            archs: ['gpu', 'mc'],
-           cpus: ['gpu': 12, 'mc': 36],
            buildPath: '$XDG_RUNTIME_DIR/easybuild/build',
            unusePath: '/apps/dom/UES/jenkins/7.0.UP01/ARCH/easybuild/tools/modules/all:/apps/dom/UES/jenkins/7.0.UP01/ARCH/easybuild/modules/all',
            modulesProduction: '/apps/common/UES/jenkins/production/login',
            prefixProduction: '$APPS/UES/jenkins/7.0.UP01/ARCH/easybuild']
 
+def fulen = [name: 'fulen',
```
gppezzi: Probably irrelevant for TSA?

easybuild/easyblocks/craytoolchain.py
```diff
@@ -0,0 +1,84 @@
+##
```
gppezzi: This could have side effects on other systems, can you please comment why you need to override the craytoolchain easyblock?

easybuild/easyblocks/cp2kcray.py
```diff
@@ -1,806 +0,0 @@
-##
```
gppezzi: Is this relevant for TSA?

easybuild/easyblocks/boostcray.py
```diff
@@ -1,222 +0,0 @@
-##
```
gppezzi: Can you please confirm that you really need to remove this easyblock for TSA and that we don't need anymore on Daint?

easybuild/easyblocks/boost.py
```diff
@@ -1,240 +0,0 @@
-##
```
gppezzi: Can you please check if this won't break boost builds on other systems?

jenkins/Machines.groovy
lucamar: Yes, anyway the file `Machines.groovy` is addressed by #1510 that can be merged only after the recipes are available

easybuild/easyblocks/cp2kcray.py
lucamar: No, but we never used it anyway, so it's a good occasion to get rid of it

easybuild/easyblocks/boostcray.py
lucamar: It is a really outdated easyblock, we can now safely rely on the default easyblock for Boost

easybuild/easyblocks/boost.py
lucamar: See comment above

easybuild/easyblocks/craytoolchain.py
lucamar: This change was addressing an issue with Modules versions 4.0 and 4.1: 
```
On versions 4.0 and 4.1, unloading an unexistent modulefile generates an Unable to locate modulefile error. Starting with version 4.2, unloading a module only looks at loaded module list and does not trigger a modulefile search. So starting version 4.2 the same behavior than Modules compatibility version is obtained.
```
See section `unload` in https://modules.readthedocs.io/en/latest/diff_v3_v4.html

Since this issue was affecting MCH workflow, we have later introduced the compatibility mode of Modules, which reverts to Modules 3.2.1:
```
tsa-ln002:~$ which modulecmd
/cm/local/apps/environment-modules/4.0.0/bin/modulecmd
lucamar@tsa-ln002:~$ modulecmd -V | grep -w VERSION
VERSION=3.2.11 
```
Therefore, this custom easyblock is currently not necessary anymore.

jenkins/Machines.groovy
lucamar: Removed `Machines.groovy` from this pull request: any change will be addressed in #1510

gppezzi:
🎉

PR #1512, opened 2020-02-26
[daint,dom] Create GDAL-2.2.2-CrayGNU-19.10.eb

twrobinson:
Add GDAL 2.2.2 with cray-netcdf and cray-hdf5 corresponding to default versions on Daint as at 26 Feb 2020 (PE 19.10).

### DIFF
--- a/easybuild/easyconfigs/g/GDAL/GDAL-2.2.2-CrayGNU-19.10.eb
+++ b/easybuild/easyconfigs/g/GDAL/GDAL-2.2.2-CrayGNU-19.10.eb
@@ -1,0 +1,32 @@
+easyblock = 'ConfigureMake'
+
+name = 'GDAL'
+version = '2.2.2'
+
+homepage = 'http://www.gdal.org/'
+description = """GDAL is a translator library for raster geospatial data formats that is released under an X/MIT style
+ Open Source license by the Open Source Geospatial Foundation. As a library, it presents a single abstract data model
+ to the calling application for all supported formats. It also comes with a variety of useful commandline utilities for
+ data translation and processing."""
+
+toolchain = {'name': 'CrayGNU', 'version': '19.10'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['http://download.osgeo.org/gdal/%(version)s/']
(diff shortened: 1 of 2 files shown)


twrobinson:
In response for GDAL in webrt ticket 38426

easybuild/easyconfigs/g/GDAL/GDAL-2.2.2-CrayGNU-19.10.eb
```diff
@@ -0,0 +1,32 @@
...
+ data translation and processing."""
+
+toolchain = {'name': 'CrayGNU', 'version': '19.10'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['http://download.osgeo.org/gdal/%(version)s/']
+sources = [SOURCELOWER_TAR_XZ]
+
+dependencies = [
+    ('cray-netcdf/4.6.3.1', 'EXTERNAL_MODULE'),
```
kraushm:
```suggestion
    ('cray-netcdf/4.6.3.1', EXTERNAL_MODULE),
```

easybuild/easyconfigs/g/GDAL/GDAL-2.2.2-CrayGNU-19.10.eb
```diff
@@ -0,0 +1,32 @@
...
+ data translation and processing."""
+
+toolchain = {'name': 'CrayGNU', 'version': '19.10'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['http://download.osgeo.org/gdal/%(version)s/']
+sources = [SOURCELOWER_TAR_XZ]
+
+dependencies = [
+    ('cray-netcdf/4.6.3.1', EXTERNAL_MODULE),
```
gppezzi: Since the 19.10 deployment we rely on the default versions from the cdt/19.10 module and therefore the external modules should not have a version appended, unless you need a non-default version (in this last case, you should update the [metadata file](https://github.com/eth-cscs/production/blob/master/easybuild/cray_external_modules_metadata-19.10.cfg) accordingly)

PR #1513, opened 2020-02-26
[TSAtds] MCH non-CUDA GNU software stack

kraushm:
To align with MCH workflow, this non-CUDA version of the GNU programming environment is going to be deployed.

PR #1514, opened 2020-03-03
[TSAtds] Add PGI non-CUDA + update PrgEnv-*

kraushm:
This PR will add PGI 19.9 noCUDA, update the production list for Tsa RH7.7, and update PrgEnv* easyconfig files.

It should be merged after PR https://github.com/eth-cscs/production/pull/1513, since it builds on modifications of the Machine file.


kraushm:
retest this please

kraushm:
The Jenkins check fails for the newly added PrgEnv-pgi/gnu-nocuda modules, since the folders/modules invoked by the PrgEnv(s) do not exist yet.

easybuild/easyconfigs/o/OpenMPI/OpenMPI-4.0.2-GCC-8.3.0.eb
```diff
@@ -26,6 +26,7 @@ configopts = (
     "--without-alps "
     "--without-libfabric "
     "--without-lsf "
+    "--without-mxm "
```
gppezzi: you are modifying an existing eb recipe, can you confirm that this is not currently deployed in (pre-)production?

easybuild/easyconfigs/p/PrgEnv-gnu/PrgEnv-gnu-tsa-19.2.eb
```diff
@@ -15,7 +15,7 @@ description = """
 
 toolchain = SYSTEM
 
-local_path = '/apps/arolla/UES/jenkins/RH7.6/%s/%s/easybuild/modules/all' % (local_prgenv, version)
+local_path = '/apps/arolla/UES/jenkins/RH7.7/%s/%s/easybuild/modules/all' % (local_prgenv, version)
```
gppezzi: same here, are you sure this has no side-effects in the current deployed system?

kraushm:
retest this please

easybuild/easyconfigs/p/PrgEnv-gnu/PrgEnv-gnu-tsa-19.2.eb
kraushm: This one was reverted to it's original and a new module (with suffix RH7.7) was added.

easybuild/easyconfigs/o/OpenMPI/OpenMPI-4.0.2-GCC-8.3.0.eb
kraushm: This eb recipe was present in master, but in the end was not deployed on the system.

PR #1525, opened 2020-03-11
[dom-gpu] SIRIUS 6.5.1

simonpintarelli:
add SIRIUS v6.5.1 for GNU/Intel

PR #1526, opened 2020-03-12
[daint-gpu] Update libxc-4.3.4-CrayIntel-19.10.eb

lucamar:
I had to revert to the easyblock `ConfigureMake` since some library files requested by `CP2K 7.1` were missing in the CMake build performed with the easyblock `CMakeMake`. 
I have also fixed the download link which was not working any more to download the package (see #1525).


simonpintarelli:
The gitlab download link is working for me at home and also if I run easybuild manually on daint.

lucamar:
@simonpintarelli It might be possible that the compute nodes could not resolve the link temporarily (the test runs on the compute nodes)

PR #1528, opened 2020-03-12
[daint-gpu] libxc 4.3.4 with CrayGNU

lucamar:
I am submitting a new recipe for `libxc-4.3.4` with `CrayGNU`, since CP2K requires also the Fortran library files `libxcf03` and `libxcf90` in addition to the `libxc` file and the former ones were missing in the build with the easyblock `CMakeMake`. This recipe builds both the static and shared library files, that I have added to the sanity check list: the build will take a bit more time to complete, however I have removed the constraint `parallel = 1` to allow multiple threads with `make`.

The download from GitLab works correctly for manual builds, but recently fails with the Jenkins CI (see #1525): therefore I am reverting to the download from www.tddft.org, although in the past it was occasionally unreliable.


toxa81:
May be cmake requires additional parameters to build Fortran libraries? I'll investigate this later. We should stop using `configure` and switch to a more modern cmake.

PR #1530, opened 2020-03-13
[dom daint kesch tsa] Add recipe for reframe v3.0-dev2

teojgo:


vkarak:
Also what's wrong with Tsa?

PR #1533, opened 2020-03-16
[daint dom] CPMD 4.3 for CPU and GPU

ajocksch:
The new versions of CPMD (4.3 with GPU support and without) produce the correct results for our reframe test. However, it is still unclear whether the GPU version really uses the GPU. For our small test it does not.


vkarak:
retest this please

lucamar:
The build failed since the patch files were not found:
```
== FAILED: Installation ended unsuccessfully (build directory: /tmp/jenscscs/85a5f3cgpu/CPMD/4.3/CrayIntel-19.10-cuda-10.1-pat-7.1.1): build failed (first 300 chars): Couldn't find file patch.to.4612 anywhere, and downloading it didn't work either...
```
Probably it is better to separate in different pull requests the standard CPMD builds and the CrayPAT builds.

lucamar:
Thanks for fixing the patches @ajocksch: why have you added `PyFR` and `pycuda` to this pull request?

ajocksch:
This was a mistake. Shall I take them out?

lucamar:
> This was a mistake. Shall I take them out?

Yes, please. I have also two additional remarks:

- I think you should pack the patch files within the tarball `cpmd-v4.3.tar.gz`, maybe you can create the tarball as `cpmd-v4.3.tar.bz2` to distinguish it from the official one, which can be removed instead (I do that for `VASP`): in any case, the download has to be done manually by one of us and the patch files are intended for that version only (leaving them in the `sources` folder creates some confusion in my opinion)

- I think that the architecture files `CRAY-XE6-INTEL` and `CRAY-XE6-INTEL-CUDA` should be part of the repository and should not be stored in the `sources` folder, to allow checking the content. For instance, I have noticed that the `CUDA` file contains the following `NVCCFLAGS` without `sm_60`: 
 ````NVCCFLAGS='--default-stream per-thread -O3 -m64 -gencode arch=compute_35,code=sm_35 -gencode arch=compute_37,code=sm_37'
 ```

You could also rename the architecture files to match the version of the code that they refer to, e.g.:
```
CRAY-XE6-INTEL -> CPMD-4.3-CrayIntel-19.10.arch
CRAY-XE6-INTEL-CUDA -> CPMD-4.3-CrayIntel-19.10-cuda-10.1.arch
```
Then you would only need a single `patch` line in your EasyBuild script:
```
patches = [(
    '%(name)s-%(version)s-%(toolchain_name)s-%(toolchain_version)s%(versionsuffix)s.arch',
    '%(builddir)s/%(namelower)s-%(version)s'
)]
```

PR #1534, opened 2020-03-16
[dom-mc,daint-mc] Create LAMMPS-03Mar2020-CrayGNU-19.10.eb

twrobinson:
LAMMPS 3 March 2020 release cpu version


twrobinson:
I commented out the reax logic, and added "no-message".

PR #1536, opened 2020-03-17
[dom-gpu] LAMMPS-03Mar2020-CrayGNU-19.10-cuda-10.1-pat-7.1.1.eb

twrobinson:
LAMMPS 2 March 2020 with cuda and pat.


twrobinson:
It builds fine from command line as my user, but produces an error as jenscscs through github. Something is causing the makefile to become corrupt, or something similar.

.depend:1015: *** multiple target patterns. Stop

lucamar:
@twrobinson If you are able to build on your command line, I would merge the recipe: as a matter of fact, `jenscscs` does not build the instrumented applications, but we need to do that manually (see https://jira.cscs.ch/browse/UES-764 for step by step instructions).

PR #1540, opened 2020-03-18
[dom daint tsa] Use system python for ReFrame

teojgo:
* Add pip since not all system pythons have it.


easybuild/easyconfigs/r/reframe/reframe-3.0-dev2.eb
```diff
@@ -17,15 +20,28 @@ sources = [
...
+                ('pip', '20.0.2', '-py3')]
+builddependencies = [('pip', '20.0.2', '-py3')]
+
 keepsymlinks = True
 postinstallcmds = [
     'cp %(installdir)s/config/cscs.py %(installdir)s/reframe/settings.py',
     'cp -r %(builddir)s/reframe-tests-%(version)s/checks %(installdir)s/cscs-checks/private',
-    'mv %(installdir)s/cscs-checks %(installdir)s/checks'
+    'mv %(installdir)s/cscs-checks %(installdir)s/checks',
+    'PYTHONUSERBASE=%(installdir)s/installed_by_pip pip install --user -r %(builddir)s/reframe-%(version)s/requirements.txt',
```
vkarak: Isn't it better to have jsonschema as a depenency in `dependencies`?

teojgo: I install everything with pip via the `requirements.txt`.

teojgo:
retest this please

easybuild/easyconfigs/r/reframe/reframe-3.0-dev2.eb
```diff
@@ -17,15 +20,26 @@ sources = [
...
 ]
+
+modtclfooter = '''
+prepend-path PATH %(installdir)s/installed_by_pip/bin
```
vkarak: Why do you need that? As far as I know we are not using any binary from the dependencies.

easybuild/easyconfigs/r/reframe/reframe-3.0-dev2.eb
```diff
@@ -17,15 +20,26 @@ sources = [
...
+prepend-path PYTHONPATH %(installdir)s/installed_by_pip/lib/python3.6/site-packages
+'''
+
 sanity_check_paths = {
     'files': ['reframe.py', 'bin/reframe', 'config/cscs.py'],
     'dirs':  ['bin', 'checks', 'checks/private', 'unittests'],
 }
+
+sanity_check_commands = ['reframe -V', 'pytest --version']
```
vkarak: If you want to test `pytest`'s sanity, you can run `./test_reframe.py --version` and avoid changing the `PATH`.

vkarak: Or just do `python3 -c 'import pytest'`

teojgo: Ok I'll just do it with `python -m pytest --version`

easybuild/easyconfigs/r/reframe/reframe-3.0-dev2.eb
```diff
@@ -17,15 +20,25 @@ sources = [
...
 postinstallcmds = [
     'cp %(installdir)s/config/cscs.py %(installdir)s/reframe/settings.py',
     'cp -r %(builddir)s/reframe-tests-%(version)s/checks %(installdir)s/cscs-checks/private',
-    'mv %(installdir)s/cscs-checks %(installdir)s/checks'
+    'mv %(installdir)s/cscs-checks %(installdir)s/checks',
+    'PYTHONUSERBASE=%(installdir)s/installed_by_pip pip install --user -r %(builddir)s/reframe-%(version)s/requirements.txt',
 ]
+
+modtclfooter = '''
+prepend-path PYTHONPATH %(installdir)s/installed_by_pip/lib/python3.6/site-packages
```
victorusu: I think this one you can do with 
```
modextrapaths = {
    'PYTHONPATH': 'installed_by_pip/lib/python3.6/site-packages'
}
```

PR #1542, opened 2020-03-23
[daint dom] Update configurable-http-proxy and nodejs

rsarm:


twrobinson:
Thanks @rsarm . We will need to update on jupyter-dev as well. First we should try on Dom.

PR #1543, opened 2020-03-23
[dom] Add Spack files for dom

ekouts:


spack/scripts/update-metadata.py
```diff
@@ -0,0 +1,92 @@
...
+import yaml
+
+
+def get_spack_path(path):
+    return os.path.dirname(os.path.realpath(__file__)).split("scripts")[0] + path
+
+
+def main():
+    with open(get_spack_path("dom/packages.yaml"), "r") as infile, open(
+        get_spack_path("dom/packages_new.yaml"), "w"
```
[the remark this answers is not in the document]
ekouts: Yes I will change that, I forgot it because that's how I was testing it

spack/scripts/update-metadata.py
```diff
@@ -1,90 +1,100 @@
...
+                    cuda_version = default_versions[0].split(".")
+                    spack_version = f"{cuda_version[0]}.{cuda_version[1]}"
+                else:
+                    spack_version = default_versions[0]
+
+                new_data["packages"][spack_pkg_name]["version"] = [spack_version]
+
+            new_data["packages"][spack_pkg_name]["modules"] = new_modules
+
+    with open(args.output_file, "w") as outfile:
```
vkarak, line 97: You should check that `output_file` is not the same as `input_file`.

ekouts: Why not allow to overwrite a file?

vkarak: What if you accidentally overwrite the original `packages.yaml` file? But if that's a valid usage scenario, I have not a problem.

spack/scripts/update-metadata.py
```diff
@@ -0,0 +1,92 @@
...
+        new_data = copy.deepcopy(data)
+        for spack_name, pv in data["packages"].items():
+            if spack_name == "mpich":
+                # All mpich versions will be mapped to the default cray-mpich module
+                cmd = subprocess.run(
+                    "module avail cray-mpich", stderr=subprocess.PIPE, shell=True
+                )
+                regex_default_module = re.compile(r"cray-mpich\/(\S*?)\(default\)\s+")
+                default_version = regex_default_module.findall(
+                    cmd.stderr.decode("utf-8")
```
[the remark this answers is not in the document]
ekouts: I was getting a message like this `TypeError: cannot use a string pattern on a bytes-like object`

spack/scripts/update-metadata.py
ekouts: Right now we are keeping all the versions of the modules in one file, so in theory ` spack/dom/packages.yaml` will be getting bigger and be rewritten.

PR #1552, opened 2020-03-26
[dom daint tsa] Install pygelf using pip for ReFrame 3.0-dev2

teojgo:


victorusu:
@vkarak, can we merge this one, so that @teojgo can modify the files on #1554 ?

vkarak:
@victorusu @teojgo Guys, I don't understand: #1554 does not install pygelf, so I was asking that this modification should be done for the 3.0-dev3 recipe.

PR #1555, opened 2020-03-30
[dom daint] Remove ipykernel from production list

ekouts:
I am not sure why this recipe was introduced in the production list but, as far as I can tell, it is not used in any recipe of a supported application.


lucamar:
Thanks @ekouts: if we need to remove the modulefile from the system, we need to do that manually.

PR #1559, opened 2020-03-31
[dom daint tsa] Add the '--upgrade' option to pip

teojgo:


easybuild/easyconfigs/r/reframe/reframe-3.0-dev3.eb
```diff
@@ -28,7 +28,7 @@ postinstallcmds = [
     'cp %(installdir)s/config/cscs.py %(installdir)s/reframe/settings.py',
     'cp -r %(builddir)s/reframe-tests-%(version)s/checks %(installdir)s/cscs-checks/private',
     'mv %(installdir)s/cscs-checks %(installdir)s/checks',
-    'PYTHONUSERBASE=%(installdir)s/installed_by_pip pip install '
+    'PYTHONUSERBASE=%(installdir)s/installed_by_pip pip install --upgrade '
```
vkarak: Shouldn't it be `--upgrade pip`?

teojgo: This upgrades the pip itself which we don't need since pip is provided by another module.

PR #1565, opened 2020-04-03
[daint-gpu] Remove PROJ from hidden-modules list

kraushm:
In ticket https://webrt.cscs.ch/Ticket/Display.html?id=38747, the point is made that the PROJ module is not visible anymore on RH7.7 - compared to other systems.
Therefore, I suggest to remove PROJ from the system_wide.cfg.


lucamar:
retest this please

gppezzi:
I don't agree with changing a site wide setting for a request from one specific user, this change will have side effects in installations on other system as well. 

If PROJ is visible in other systems is because it is explicitly added to the production list. Why don't you do the same here?

lucamar:
> I don't agree with changing a site wide setting for a request from one specific user, this change will have side effects in installations on other system as well.
> 
> If PROJ is visible in other systems is because it is explicitly added to the production list. Why don't you do the same here?

@gppezzi This change was merged because there are more users on different systems looking for `PROJ`: rhe request mentioned by @kraushm is just the more recent one that we have received.

gppezzi:
This change in the settings will cause (or already caused) re-builds in all systems and we will end up with some modules that have a dependency in the hidden version of proj and some with the visible version. In order to fix that we would need to a force rebuild to generate all modules in a coherent way or manually tweak the module files.

As long as we don't have lmod, removing packages from the hide list can be very tricky and it should be carefully checked before. 

If there are more users asking for PROJ, then in my opinion it should appear in the production list (as any software that users need to load explicitly).

lucamar:
I agree about adding `PROJ` explicitly to the production list, however I also think that it should not appear in the `hide-deps`, since that would build two copies as we do for `CMake` and occasionally creates troubles.

gppezzi:
My point was to highlight that this change has potential impacts in every software deployment at CSCS and it should have been reviewed by a wider audience to make sure that it does not generate conflicts also on other systems. 

Removing packages from the site wide hide-deps setting is error-prone and require coordination, while adding a new module is much less disruptive, that's why I think that it should be the preferred solution. 

But now that change is done and since you prefer to keep it like this, then I just ask you to check in the other systems if this has caused side effects or not (I don't know if PROJ is deployed elsewhere).

lucamar:
In this specific case the impact is zero: only the production list of `Tsa` would have been affected, since `PROJ` is a depedency of `CDO` (but not the `Cray` module), `GDAL` and `R`. 
However, the production list of `Tsa` was already built with `PROJ` removed for the `hide-deps` list in the branch that @kraushm and I have been using so far.

PR #1568, opened 2020-04-03
[daint-gpu] QuantumESPRESSO 6.5a1 gpu with CrayIntel

lucamar:
I provide the EasyBuild recipe for QuantumESPRESSO 6.5a1: the GPU release needs the PGI compiler and can only build `pw.x`: see instructions at https://gitlab.com/QEF/q-e-gpu/-/wikis/home

The script is difficult to read, since I had to remove several compiler flags set by default by EasyBuild
(e.g.: `-O2 -Mflushz -mp`...). Do you know a quick toolchain option to remove all these default flags?


easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.5a1-CrayPGI-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,53 @@
...
+    ('cudatoolkit/10.1.105_3.27-7.0.1.1_4.1__ga311ce7', EXTERNAL_MODULE),
+]
+
+dependencies = [
+    ('intel/19.0.1.144', EXTERNAL_MODULE)
+]
+
+# Scalapack disabled: the GPU eigensolver can outperform the parallel CPU one 
+configopts = ' CC=cc F77=pgf90 MPIF90=ftn '
+configopts += ' --enable-openmp --enable-parallel --with-scalapack=no '
```
jgphpc, line 31:
```suggestion
configopts += ' --enable-openmp --enable-parallel --with-scalapack=no '
```

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-6.5a1-CrayPGI-19.10-cuda-10.1.eb
```diff
@@ -0,0 +1,53 @@
...
+]
+
+# Scalapack disabled: the GPU eigensolver can outperform the parallel CPU one 
+configopts = ' CC=cc F77=pgf90 MPIF90=ftn '
+configopts += ' --enable-openmp --enable-parallel --with-scalapack=no ' 
+configopts += ' --with-cuda=$CUDATOOLKIT_HOME --with-cuda-cc=60 '
+configopts += ' --with-cuda-runtime=10.1 '
+
+# experimental CUDA-Aware MPI: add -D__GPU_MPI in make.inc
+prebuildopts = """
```
jgphpc, line 36: maybe need a patch file instead here but not mandatory

lucamar: Thanks for the comment @jgphpc: the problem with the patch is that EasyBuild applies it before building. On the other hand, I need to adapt the file `make.inc` after it is generated in the `configure` step, since it does not exist when the build tarball is unpacked.

PR #1570, opened 2020-04-04
[daint-gpu] CP2K 7.1 gpu with CrayIntel

lucamar:
Minor change with a fix: adding MKL include and link line options and unloading `cray-libsci` before building.


lucamar:
retest this please

victorusu:
@lucamar, I just wonder if there was any performance impact by changing from fftw to mkl. Because, GROMACS is slower when using mkl.

lucamar:
> @lucamar, I just wonder if there was any performance impact by changing from fftw to mkl. Because, GROMACS is slower when using mkl.

Thanks for your comment @victorusu: as a matter of fact, I have some performance regression that you can check in https://confluence.cscs.ch/display/GENERAL/CP2K+7.1 vs. https://confluence.cscs.ch/display/GENERAL/CP2K+6.1 and I am still investigating: sure, part of the regression can be due to the MKL FFT, which works otherwise better with other applications.

lucamar:
retest this please

PR #1575, opened 2020-04-10
[daint-gpu] Add GROMACS 2020.1 (GPU version)

victorusu:


jenkins-builds/7.0.UP01-19.10-gpu
```diff
@@ -13,6 +13,7 @@
  GROMACS-2018.6-CrayGNU-19.10-cuda-10.1.eb          --set-default-module
  GROMACS-2019.1-CrayGNU-19.10-cuda-10.1.eb
  GROMACS-2020-CrayGNU-19.10-cuda-10.1.eb
```
lucamar, line 15:
```suggestion
```

lucamar:
Given the reply of @victorusu in #1576, I suggest to remove from the production list the "intermediate" builds here as well, since the modules won't be removed from the system anyway.

PR #1576, opened 2020-04-10
[daint-mc] Add GROMACS 2020.1 MC

victorusu:


lucamar:
Thanks @victorusu: after this pull request, we are going to have the following non-default GROMACS releases in production, on top of the default (older) one:
```
 GROMACS-2019.1-CrayGNU-19.10.eb
 GROMACS-2020-CrayGNU-19.10.eb
 GROMACS-2020.1-CrayGNU-19.10.eb
```
Do you need to keep all of them?

victorusu:
Ciao @lucamar,
We haven't announced any removal of GROMACS versions, so I would like to keep all.
We can announce the removal and keep only the current default and the most up to date one.

lucamar:
Thanks for your reply @victorusu: I suggest to remove from the production list the "intermediate" builds: the modules won't be removed from the system anyway, until you do that manually.

jenkins-builds/7.0.UP01-19.10-mc
```diff
@@ -14,6 +14,7 @@
  GROMACS-2018.6-CrayGNU-19.10.eb                    --set-default-module
  GROMACS-2019.1-CrayGNU-19.10.eb
  GROMACS-2020-CrayGNU-19.10.eb
```
lucamar, line 16:
```suggestion
```

PR #1578, opened 2020-04-14
[daint-mc] Add deepmd-kit-1.1.4

rsarm:


rsarm:
adding `--local_ram_resources=60000` to `bazel build` because the jenkins test creashed with
```
slurmstepd: error: Detected 1 oom-kill event(s) in step 21922566.0 cgroup. Some of your processes may have been killed by the cgroup out-of-memory handler.
srun: error: nid00025: task 0: Out Of Memory
srun: Terminating job step 21922566.0
```

lucamar:
Thanks for providing the package @rsarm: in the previous release `0.10.2`  that was addressed by an EasyBuild recipe there used to be also other executables together with `dp_ipi`: 
```
ls /apps/daint/UES/sandbox/lucamar/easybuild/software/DeePMD/0.10.2-CrayGNU-19.10/bin
dp_frz  dp_ipi  dp_mdnn  dp_test  dp_train
```
Do you know if they can still be built of have they been discontinued? 
I have also asked about a user of the code in CSCS RT #38858.

rsarm:
The way deepmd is built has changed and I think now they put all the functionalities together through the executables `dp` and `dp_ipi`. Now one does `dp train ...` like here https://github.com/deepmodeling/deepmd-kit#write-the-input-script

lucamar:
Thanks for your reply @rsarm: do you plan to make `deepmd` available in production or just as a recipe?

rsarm:
@lucamar just a recipe. I did this for 38732

PR #1582, opened 2020-04-17
[dom-gpu] XALT cce/9.0.2 possible workaround

victorusu:
This is a possible workaround, but not complete


ekouts:
Hi @victorusu , is the GROMACS recipe in this PR by mistake? Also, should the tag in the title be [dom]?

victorusu:
@ekouts, yes it is...
I just merged from master and it has disappeared.

easybuild/easyconfigs/x/xalt/xalt-2.8.eb
```diff
@@ -0,0 +1,77 @@
...
+source_urls = ['https://github.com/xalt/xalt/archive/']
+#source_urls = ['https://github.com/victorusu/xalt.git']
+sources = [SOURCELOWER_TAR_GZ]
+
+patches = [('Daint_config.py', 'Config/')]
+
+local_transmission = "curl"
+
+# generic configurations
+preconfigopts  = 'cd %(builddir)s && cp xalt-xalt-%(version)s/Config/Daint_config.py . && '
```
vkarak, line 24: This `xalt-xalt` name is a bit strange.

easybuild/easyconfigs/x/xalt/xalt-2.8.eb
```diff
@@ -0,0 +1,77 @@
...
+preconfigopts += "rm -rf xalt-xalt-%(version)s && "
+preconfigopts += 'git clone https://github.com/victorusu/xalt.git xalt-xalt-%(version)s && '
```
vkarak, line 26: Why not putting that in `source_urls`?

easybuild/easyconfigs/x/xalt/xalt-2.8.eb
```diff
@@ -0,0 +1,77 @@
+# contributed by Victor Holanda (CSCS)
+
+name = 'xalt'
+version = "2.8"
```
vkarak, line 4: I would suggest adding a suffix to make sure that this is our patched version.

easybuild/easyconfigs/x/xalt/xalt-2.8.eb
victorusu: This is out-dated

easybuild/easyconfigs/x/xalt/xalt-2.8.eb
victorusu: I am not sure if we should use a suffix, due to the fact the xalt installation is already daint/dom directed and the patch is almost the same as of `Config.py` patch that we apply to get the config. This is just an additional config.

victorusu: @vkarak, what do you think?

vkarak: Fine for me.

PR #1585, opened 2020-04-20
[daint-gpu] DeePMD dependency backports.weakref

lucamar:
Fixing the recipe of the `DeePMD` dependency `backports.weakref` to comply with the current EasyBuild standard and updating DeePMD with a tentative recipe to address the latest release.


lucamar:
I remove the recipe for version `1.1.4` since this might be addressed by #1578

PR #1594, opened 2020-04-24
[dom-mc] Renaming EasyBuild recipe according to versionsuffix

lucamar:
I propose to rename the easyconfig file of `LAMMPS-deepmd` according to the `versionsuffix` defined  within it.
@rsarm and @twrobinson: do you know if the `LAMMPS` package `user-deepmd` is compatible with the LAMMPS `gpu` package? I could not find `user-deepmd` in the list https://lammps.sandia.gov/doc/Packages_user.html


rsarm:
@lucamar I don't know. I have always installed it for users of the mc partition. On their anaconda recipe I see they don't use the gpu package either https://github.com/njzjz/lammps-dp-feedstock/blob/master/recipe/build.sh for the gpu version. Maybe is not compatible.

PR #1597, opened 2020-04-27
[dom-gpu] deprecating older .ebs

jgphpc:


jgphpc:
[eff.txt](https://github.com/eth-cscs/production/files/4540850/eff.txt)

gppezzi:
Thanks for the initiative @jgphpc !

I've been quickly through to the list and the only remark I have is that you are deprecating two M4 versions with Cray PE 19.09 (current default) and they could be used as dependency in other recipes (or by user recipes that we are not aware of).  

I would suggest to keep the files from valid PEs which we are unsure if they are used or not, unless you have another reason do deprecated it?

jgphpc:
will keep old pe to keep safe

PR #1601, opened 2020-04-29
[daint-gpu] more updates to run on Jupyer-lab

jfavre:
one more missing ParaView recipe for Jupyterlab


twrobinson:
Looks good to me. I tested it with the new jupyterlab 1.1 1 and I was able to run without manually setting the PYTHONPATH

PR #1603, opened 2020-04-30
[tsa, TSAtds] Downgrade rstudio version to 1.1.463

kraushm:
Rstudio v1.2.1335 exhibits issue on Red Hat - see https://webrt.cscs.ch/Ticket/Display.html?id=38532, v1.1.463 does not show the issue.


kraushm:
The test of this PR does not take into account the `--disable-rpath` otpion of the production list file. Building manually as the Jenkins user did work.

kraushm:
> The test of this PR does not take into account the `--disable-rpath` otpion of the production list file. Building manually as the Jenkins user did work.

The recipe was changed to already contain the `--disable-rpath` option of the command line in the `toolchainopts`, therefore, it is not needed anymore in the production list.

PR #1616, opened 2020-05-07
[dom] Eb templates 19.10

ekouts:
The purpose of this PR is to homogenize the recipes that have python and cuda dependencies, so that automatic updates are easier.

Main changes:
1. Whenever possible we user the python and cuda template variables (`pyver`, `pyshortver`, `pymajver`, `cudaver`, `cudashortver`, `cudamajver`)
2. For this reason cudatoolkit and cray-python are now dependencies, and not builddependencies. Files affected: `VASP`, `Amber`, `CP2K`, `GROMACS`, `Julia`, `LAMMPS`, `NAMD`, `nvtop`, `QuantumESPRESSO`
3. Use `PythonBundle` instead of `Bundle` when appropriate (no need to set ext_default_class  and modextspaths )
4. For PythonBundle easybuild will fail if some extension tries to download more extension. We have to add all of them explicitly, but this is good for reproducibility (see for example `PyExtensions-2.7.15.6-CrayGNU-19.10.eb` recipe, that was downloading python-dateutil, pytz, kiwisolver, cycler because of matplotlib, but it wasn't failing as a `Bundle`
5. Use `exts_default_options` to make the new recipes a bit shorter
6. Bazel doesn't depend on python, so it shouldn't have it on the suffix. It is used only in the tensorflow recipe so we have to change this as well
7. Add `cray-python/2.7.15.6` and `cray-python/3.6.5.6` entry in the 19.10 metadata since they are used in their respective PyExtensions recipes

Issues:
1. Still missing recipes: Visit-3.1.0-CrayGNU-19.10.eb, ParaView-5.7
2. WIP pycuda-python2, oq-engine, CPMD, VTK
3. Eb >=4.2 has to be the default in order for the PR to pass the jenkins tests
4. there is no `pyminver` template, so when it’s unavoidable, we use local variable `_pyminver`

### DIFF
--- a/easybuild/easyconfigs/p/PyExtensions/PyExtensions-2.7.15.6-CrayGNU-19.10.eb
+++ b/easybuild/easyconfigs/p/PyExtensions/PyExtensions-2.7.15.6-CrayGNU-19.10.eb
@@ -1,12 +1,8 @@
 # contributed by Luca Marsella (CSCS)
-easyblock = 'Bundle'
+easyblock = 'PythonBundle'
 
 name = 'PyExtensions'
-local_py_maj_ver = '2'
-local_py_min_ver = '7'
-local_py_rev_ver = '15.6'
-local_pyver = '%s.%s' % (local_py_maj_ver, local_py_min_ver)
-version = '%s.%s' % (local_pyver, local_py_rev_ver)
+version = '%(pyver)s'
 
 homepage = 'https://pypi.python.org/pypi'
 description = """This module is a bundle of Python packages on Cray systems based on the module cray-python"""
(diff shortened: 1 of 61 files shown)


easybuild/easyconfigs/p/PyExtensions/PyExtensions-2.7.15.6-CrayGNU-19.10.eb
```diff
@@ -15,42 +11,33 @@ toolchain = {'name': 'CrayGNU', 'version': '19.10'}
 toolchainopts = {'pic': True, 'verbose' : False }
 
 dependencies = [
-    ('cray-python/%s' % version, EXTERNAL_MODULE),
+    ('cray-python/2.7.15.6', EXTERNAL_MODULE),
```
henrique, line 14:
```suggestion
    ('cray-python/2.7.15.7', EXTERNAL_MODULE),
```
shouldn't we update to 2.7.15.7 instead?

ekouts: No, because we have two `PyExtensions` recipes for python2. One for 2.7.15.6 and one for 2.7.15.7. And both are used in the production list

ekouts:
retest this please

PR #1638, opened 2020-05-14
[dom] add easybuild metadata for 20.05

ekouts:


easybuild/cray_external_modules_metadata-20.05.cfg
```diff
@@ -0,0 +1,122 @@
...
+version = 3.7.3.2
+prefix = /opt/python/3.7.3.2
+
+[cray-python/3.8.2.0]
+name = Python
+version = CRAY_PYTHON_VERSION
+prefix = CRAY_PYTHON_PREFIX
+
+[cray-R]
+name = CrayR
```
victorusu, line 85: name should be `R`, not `CrayR`.
```
eb -S R
...
* $CFGS2/r/R/R-3.6.2-fosscuda-2019b.eb
```

easybuild/cray_external_modules_metadata-20.05.cfg
```diff
@@ -0,0 +1,122 @@
...
+name = HDF5
+version = CRAY_HDF5_VERSION
+prefix = CRAY_HDF5_PREFIX
+
+[cray-hdf5-parallel]
+name = HDF5
+version = CRAY_HDF5_PARALLEL_VERSION
+prefix = CRAY_HDF5_PARALLEL_PREFIX
+
+[cray-libsci]
```
victorusu, line 19: Not sure if we need it. But if we do, why not also add `cray-libsci_acc`?

ekouts: Sure, I only added whatever was in the old metadata. As soon as dom is back I will check if the cray variables are in the module and if they are, I will add this as well

easybuild/cray_external_modules_metadata-20.05.cfg
```diff
@@ -0,0 +1,122 @@
+# metadata useful to EasyBuild (hpcugent.github.io/easybuild) for modules provided by Cray on Cray systems
+# authors: Guilherme Peretti-Pezzi (CSCS), Petar Forai (IMP/IMBA), Kenneth Hoste (HPC-UGent), Eirini Koutsaniti (CSCS)
```
vkarak, line 2: If the file is automatically generated, I think it is good to mention it here.

easybuild/cray_external_modules_metadata-20.05.cfg
```diff
@@ -0,0 +1,122 @@
...
+
+[cray-fftw]
```
victorusu, line 4: @ekouts, I think we can drop almost all variables from the metadata file.
What about
```
[cray-fftw]
name = FFTW

[cray-hdf5]
name = HDF5

[cray-hdf5-parallel]
name = HDF5

[cray-libsci]
name = LibSci

[cray-netcdf]
name = netCDF, netCDF-Fortran

[cray-netcdf-hdf5parallel]
name = netCDF, netCDF-Fortran

[cray-petsc]
name = PETSc

[cray-petsc-64]
name = PETSc

[cray-petsc-complex]
name = PETSc

[cray-petsc-complex-64]
name = PETSc

[cray-python/2.7.15.6]
name = Python
version = 2.7.15.6
prefix = /opt/python/2.7.15.6

[cray-python/2.7.15.7]
name = Python
version = 2.7.15.7
prefix = /opt/python/2.7.15.7

[cray-python/3.6.5.6]
name = Python
version = 3.6.5.6
prefix = /opt/python/3.6.5.6

[cray-python/3.6.5.7]
name = Python
version = 3.6.5.7
prefix = /opt/python/3.6.5.7

[cray-python/3.7.3.2]
name = Python
version = 3.7.3.2
prefix = /opt/python/3.7.3.2

[cray-python/3.8.2.0]
name = Python

[cray-R]
name = R

[cray-tpsl]
name = TPSL

[cray-tpsl-64]
name = TPSL

[cudatoolkit/10.1.105_3.27-7.0.1.1_4.1__ga311ce7]
name = CUDA
version = CRAY_CUDATOOLKIT_VERSION
prefix = CRAY_CUDATOOLKIT_DIR

[gcc]
name = GCC

[papi]
name = PAPI

[pmi]
name = pmi

[slurm/.default]
name = slurm
version = .default
prefix = SLURMDIR
```

ekouts: I think we still have to keep some of those, at least `cray-petsc-64`, `cray-petsc-complex`, `cray-petsc-complex-64` and `cray-tpsl-64`, but I will try it, no problem

easybuild/cray_external_modules_metadata-20.05.cfg
ekouts: This was not automatically generated, because this time I had to check which variables were added to the new modules and will be recognized by easybuild. But in theory, after we drop the old versions of python, it should be only CUDA that needs to be updated, since we will have to include the version in the index of the entry.

easybuild/cray_external_modules_metadata-20.05.cfg
ekouts: I can add it to the metadata but what would be the equivalent `cray-libsci_acc` name for EasyBuild?

ekouts:
retest this please

easybuild/cray_external_modules_metadata-20.05.cfg
ekouts: Removed all of them, except for the ones I mention in the comment above

easybuild/cray_external_modules_metadata-20.05.cfg
victorusu: There is none

victorusu: @gppezzi may know more

gppezzi: I wouldn't add entries if we're not sure if they are actually needed, the only reference I see to `cray-libsci_acc` is on the Cosmo recipes. 

My guess is that it worked so far without this entry is because we don't need to explicitly rely on variables from `libsci_acc` (the cray wrappers are taking care of swapping the libraries), but I personally never used it so I would gladly take this statement back if you find an use case where just loading the libsci_acc module is not enough.

gppezzi:
LGTM but I have a couple of questions: can you explain how do you plan to roll out the new entries to production? These entries can co-exist with the ones needed for PE 19.x?  

PE 19.x will most probably be available on the system until the upgrade to CLE7UP02 and meanwhile we should be able to build software with both PEs 19.x and 20.05 (if possible, without having to explicitly switch between two different metadata files).

ekouts:
> LGTM but I have a couple of questions: can you explain how do you plan to roll out the new entries to production? These entries can co-exist with the ones needed for PE 19.x?
> 
> PE 19.x will most probably be available on the system until the upgrade to CLE7UP02 and meanwhile we should be able to build software with both PEs 19.x and 20.05 (if possible, without having to explicitly switch between two different metadata files).

Currently we have the switch metadata files for different PEs, because in the old modules (and some of the new ones) we have to specify the version in the entry, for example [cray-petsc-complex](https://github.com/eth-cscs/production/pull/1638/files#diff-9df6d9324dbd238890b9ac5550fca7d4R32). As soon as we get the environment variables <module>_VERSION and <module>_PREFIX in all the modules and stop using the old ones we should be able to have one metadata file, in which only the index and the name of the module are specified like this:
```
[cray-netcdf]
name = netCDF, netCDF-Fortran
```
Since we are still using the old modules in 19.10 (without the variables) and we want the recipes to have the default, versionless modules I am not sure how we could avoid switching.

gppezzi:
Thanks @ekouts for the explanation. Since PE 20.x are not yet available in production I think we can merge this. Anyone else has comments?

We can later on think how to document and expose this metadata on the system (for example add a new module `Easybuild-cscs/2020`).

lucamar:
We can create a module wrapper in sync with the Cray Programming Environments, if needed. E.g.:
```
EasyBuild-custom/19.10
EasyBuild-custom/20.05
```
> We can later on think how to document and expose this metadata on the system (for example add a new module `Easybuild-cscs/2020`).

PR #1642, opened 2020-05-22
[daint TSAtds] Add recipe for ReFrame version 3.0-dev6

jenkins-cscs:


teojgo:
Ok to test

easybuild/easyconfigs/r/reframe/reframe-3.0-dev6.eb
```diff
@@ -0,0 +1,52 @@
...
+    'cp %(installdir)s/config/cscs.py %(installdir)s/reframe/settings.py',
+    'cp -r %(builddir)s/reframe-tests-%(version)s/checks %(installdir)s/cscs-checks/private',
+    'mv %(installdir)s/cscs-checks %(installdir)s/checks',
+    'PYTHONUSERBASE=%(installdir)s/installed_by_pip pip install --upgrade '
+    '--user -r %(builddir)s/reframe-%(version)s/requirements.txt pygelf',
+    'python3 -m venv docs_env && source docs_env/bin/activate && '
+    'pip install -r docs/requirements.txt && make -C docs man && deactivate && '
+    'mkdir -p %(installdir)s/share/man/man1 %(installdir)s/share/man/man8 && '
+    'cp docs/man/reframe.1 %(installdir)s/share/man/man1 && '
+    'cp docs/man/reframe.settings.8 %(installdir)s/share/man/man8'
```
vkarak, line 37: Is there a way that this can done also from pip? Because then the pip installation diverges. Also the tutorial should also be installed by the pip installation.

teojgo: Then the `setup.py` script of reframe should be changed to include the manual as well as the tutorial.

vkarak: I will open an issue in the reframe repo.

PR #1651, opened 2020-05-27
[daint-gpu] Add VASP 6.1.0 in production

lucamar:
The new modules of `VASP 6.1.0` will be installed on Dom and Piz Daint as non default.

### DIFF
--- a/easybuild/easyconfigs/v/VASP/VASP-6.1.0-CrayIntel-19.10-cuda-10.1.eb
+++ b/easybuild/easyconfigs/v/VASP/VASP-6.1.0-CrayIntel-19.10-cuda-10.1.eb
@@ -1,0 +1,39 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'MakeCp'
+
+name = 'VASP'
+version = '6.1.0'
+local_cudaversion = '10.1'
+versionsuffix = '-cuda-%s' % local_cudaversion
+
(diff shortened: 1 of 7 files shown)


lucamar:
Before merging this pull request and installing `VASP 6.1.0` in production, I should clarify if we need an additional unix group for the last release, since the license is different according to the news on the web page: see https://vasp.at/post/new-release-vasp-610

lucamar:
I have uploaded the recipes for `VASP 6.1.0` GPU with `CrayIntel` as the corresponding executable performs better with respect to the one built with `CrayPGI`. 
The latter should become the default compiler using OpenACC for the in future GPU releases, according to the Wiki: see https://www.vasp.at/wiki/index.php/Installing_VASP.6.X.X
However I have found some issues in the compilation and I have tried to contact the developers in the VASP Forum at https://www.vasp.at/forum: unfortunately I cannot post and I got no reply so far.

lucamar:
The build has completed successfully: see https://jenkins.cscs.ch/blue/organizations/jenkins/TestingEB/detail/TestingEB/3554/pipeline

lucamar:
I have added a modified version of the script `production.sh` that will take into account the new group `vasp6`, which will be enforced after June 30th: in the meanwhile, all `vasp` users can use the new release. I have added a `Warning` on the [dedicated page](https://user.cscs.ch/computing/applications/vasp) of the User Portal.

PR #1652, opened 2020-05-27
[dom daint TSAtds] Add ReFrame env variables

teojgo:


easybuild/easyconfigs/r/reframe/reframe-3.0-dev6.eb
```diff
@@ -38,14 +36,16 @@ postinstallcmds = [
 ]
 
 modextrapaths = {
-    'PYTHONPATH': 'installed_by_pip/lib/python3.6/site-packages'
+    'PYTHONPATH': 'installed_by_pip/lib/python3.6/site-packages',
+    'RFM_CONFIG_FILE': 'config/cscs.py',
```
vkarak, line 40: Shouldn't that have been prefixed with `$EBROOTREFRAME`?

easybuild/easyconfigs/r/reframe/reframe-3.0-dev6.eb
```diff
@@ -38,14 +36,16 @@ postinstallcmds = [
...
+    'RFM_CHECK_SEARCH_PATH': 'cscs-checks',
```
vkarak, line 41: Same here.

PR #1654, opened 2020-05-27
[TSAtds] Adapt PrgEnv-gnu to prevent errors on module purge

kraushm:
`module purge` raised errors on RH7.7 for the GNU modules. The reason is, that the deps. `foss` and `fosscuda` try to unload modules on their own.


lucamar:
retest this please

PR #1657, opened 2020-05-28
[dom] Add small utilities

jgphpc:


jgphpc:
thks @kraushm 
I will push @ekouts new format

PR #1659, opened 2020-06-02
[dom-gpu, daint-gpu] Eb templates in dask, numpy, SWIG and PyExtensions

ekouts:

### DIFF
--- a/easybuild/easyconfigs/n/numpy/numpy-1.17.2-CrayGNU-19.10.eb
+++ b/easybuild/easyconfigs/n/numpy/numpy-1.17.2-CrayGNU-19.10.eb
@@ -3,13 +3,6 @@ easyblock = 'PythonPackage'
 name = 'numpy'
 version = '1.17.2'
 
-local_py_maj_ver = '3'
-local_py_min_ver = '6'
-local_py_rev_ver = '5.7'
-
-local_pyver = '%s.%s.%s' % (local_py_maj_ver, local_py_min_ver, local_py_rev_ver)
-local_pyshortver = '%s.%s' % (local_py_maj_ver, local_py_min_ver)
-
 homepage = 'https://github.com/numpy/numpy'
 description = "Horovod is a distributed training framework for TensorFlow."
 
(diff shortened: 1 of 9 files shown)


lucamar:
The build of `SWIG` failed because it could not download the package from Sourceforge. I have also tried a manual download, without success: the following command hangs both on Dom and Piz Daint
`srun -C gpu --ntasks=1 wget https://sourceforge.net/projects/swig/files/latest/download`

easybuild/easyconfigs/n/numpy/numpy-1.17.2-CrayGNU-19.10.eb
```diff
@@ -3,13 +3,6 @@ easyblock = 'PythonPackage'
...
 
-local_py_maj_ver = '3'
-local_py_min_ver = '6'
-local_py_rev_ver = '5.7'
-
-local_pyver = '%s.%s.%s' % (local_py_maj_ver, local_py_min_ver, local_py_rev_ver)
-local_pyshortver = '%s.%s' % (local_py_maj_ver, local_py_min_ver)
-
 homepage = 'https://github.com/numpy/numpy'
 description = "Horovod is a distributed training framework for TensorFlow."
```
henrique, line 7: a bit unrelated but as we are at it...
```suggestion
description = """NumPy is the fundamental package for scientific computing with Python. It contains among other things:
 a powerful N-dimensional array object, sophisticated (broadcasting) functions, tools for integrating C/C++ and Fortran
 code, useful linear algebra, Fourier transform, and random number capabilities. Besides its obvious scientific uses,
 NumPy can also be used as an efficient multi-dimensional container of generic data. Arbitrary data-types can be
 defined. This allows NumPy to seamlessly and speedily integrate with a wide variety of databases."""
```

ekouts: thanks @henrique !

ekouts:
retest this please

PR #1660, opened 2020-06-02
[dom-gpu, daint-gpu] Eb templates in boost

ekouts:

### DIFF
--- a/easybuild/easyconfigs/b/Boost/Boost-1.70.0-CrayGNU-19.10-python2.eb
+++ b/easybuild/easyconfigs/b/Boost/Boost-1.70.0-CrayGNU-19.10-python2.eb
@@ -2,11 +2,8 @@
 name = 'Boost'
 version = '1.70.0'
 
-local_py_maj_ver = '2'
-local_py_min_ver = '7'
-local_py_rev_ver = '15.7'
-
-versionsuffix = '-python%s' % local_py_maj_ver
+_pyminver = '7'
+versionsuffix = '-python%(pymajver)s'
 
 homepage = 'http://www.boost.org/'
 description = """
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/b/Boost/Boost-1.70.0-CrayGNU-19.10-python2.eb
```diff
@@ -45,7 +40,7 @@ dependencies = [
 # for python3 the corresponding lib is libboost_python36.so
 sanity_check_paths = {
     'files': ['lib/libboost_system.so', 'lib/libboost_mpi.so',
-              'lib/libboost_python%s%s.so' % (local_py_maj_ver, local_py_min_ver)],
+              'lib/libboost_python%%(pymajver)s%s.so' % _pyminver],
```
lucamar, line 43: Thanks @ekouts: is it possible to use the same format for both `major` and `minor` versions, as in my suggestion ? 
```suggestion
              'lib/libboost_python%s%s.so' % (pymajver, _pyminver)],
```

easybuild/easyconfigs/b/Boost/Boost-1.70.0-CrayGNU-19.10-python3.eb
```diff
@@ -46,7 +41,7 @@ patches = ['Boost-1.70.0-python%s.patch' % local_py_maj_ver]
 # for python3 the corresponding lib is libboost_python36.so
 sanity_check_paths = {
     'files': ['lib/libboost_system.so', 'lib/libboost_mpi.so',
-              'lib/libboost_python%s%s.so' % (local_py_maj_ver, local_py_min_ver)],
+              'lib/libboost_python%%(pymajver)s%s.so' % _pyminver],
```
lucamar, line 44: Thanks @ekouts: same comment here as above for `majver` and `minver`, just to have the same formatting, if possible

easybuild/easyconfigs/b/Boost/Boost-1.70.0-CrayGNU-19.10-python2.eb
ekouts: `pymajver` is a template variable, while `_pyminver` is a local variable. The problem is template variables don't work with this syntax.

lucamar: Ok, never mind!

PR #1662, opened 2020-06-02
[dom-gpu, daint-gpu] Eb template values in nvtop and line_profiler

ekouts:

When `cray-python` and `cudatoolkit` are direct dependencies, we can use the easybuild template variables (pyver, pyshortver, pymajver, cudaver, cudashortver, cudamajver), in order to update the recipes more easily in the future.


jgphpc:
i guess my review is not essential

PR #1664, opened 2020-06-02
[dom, daint] use eb templates in VTK

henrique:


jfavre:
There is some confusion here. As a general rule for visualization packages (ParaView and VTK), recipes with an EGL string in the name, are only applicable for daint-gpu and dom-gpu, because they use the GPU for their offscreen rendering. So the original recipe, must not have been ever compiled on {daint,dom}-mc

the reciprocal (or complementary I should say) recipes, are called with the OSMesa string in the name, and are indeed targeted at {daint,dom}-mc.the compilation took place on teh login node?

I am confused that you could indeed compile, but perhaps that is because the compilation took place on the login node? Is that so?

henrique:
@jfavre I think it runs with srun, so it's actually on a compute node: https://github.com/eth-cscs/production/blob/experimental/jenkins/JenkinsfileTestingEB#L66
However, the dependency on a gpu as you mentioned might be only at run-time, as all the required libraries during build are available anyway.

easybuild/easyconfigs/v/VTK/VTK-8.2.0-EGL-CrayGNU-19.10-python3.eb
```diff
@@ -71,19 +66,25 @@ configopts += "-DCMAKE_CXX_COMPILER:PATH=`which CC` "
...
+    configopts += '-DEGL_LIBRARY:FILEPATH=/usr/lib64/libEGL.so '
+    configopts += '-DEGL_opengl_LIBRARY:FILEPATH=/usr/lib64/libOpenGL.so '
+
+modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
+
 # a long discussion with Victor, Matthias and Rafael lead to the following hard-wired variable.
 # should the cray-python module be fixed?
 modtclfooter = """
-prepend-path LD_LIBRARY_PATH "/opt/python/3.6.5.7/lib"
+prepend-path LD_LIBRARY_PATH "/opt/python/%(pyver)s/lib"
```
ekouts, line 82: > a long discussion with Victor, Matthias and Rafael lead to the following hard-wired variable

@jfavre is this comment still valid? Do you need to hard-wire the python version or is it fine to get it from `pyver`?

henrique:
retest this please

easybuild/easyconfigs/v/VTK/VTK-8.2.0-EGL-CrayGNU-19.10-python3.eb
jfavre: should be absolutely fine to get it from pyver

ekouts: Thanks. So @Henrique I think it's better omit that comment from the new recipe.

PR #1649, opened 2020-05-26
[tsa] Change production configuration

kraushm:
MCH requested UDUNITS accessible on RH7.7 (see https://webrt.cscs.ch/Ticket/Display.html?id=39013)

PR #1668, opened 2020-06-04
[TSAtds] Remove reframe 2.18 from TSAtds production lists

teojgo:


teojgo:
Closing since this will be included in https://github.com/eth-cscs/production/pull/1649

PR #1675, opened 2020-06-08
[TSAtds] Remove Python release suffix

kraushm:
Based on the new PE release strategy using Git tags, the naming scheme of modules is moved back to the default one.


lucamar:
retest this please

PR #1684, opened 2020-06-09
[tsa] Change buildPath to prevent EB locks during concurrent builds

kraushm:
When different Jenkins projects run at the same time, EB might abort the build of a particular module in case it detects build folders/files for the same module from the other project present in the buildPath. Therefore, a separate buildPath is suggested.

### DIFF
--- a/jenkins/Machines.groovy
+++ b/jenkins/Machines.groovy
@@ -24,7 +24,7 @@ def kesch = [name: 'kesch',
 
 def tsa = [name: 'tsa',
            archs: [],
-           buildPath: '/tmp/$USER/easybuild/build',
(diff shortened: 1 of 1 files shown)


teojgo:
> Thanks @kraushm. In Jenkins projects `RegressionEB` and `TestingEB` we avoid the lock by appending a commit or a build tag to the path, while `ProductionEB` uses just the path provided in `Machines.groovy`. Do you think that it makes sense to change the `ProductionEB` pipeline file to fix it everywhere, @teojgo?

I can change by doing the same for the ProductionEB as well. Should I proceed?

kraushm:
Is fixed in #1687

PR #1685, opened 2020-06-09
[tsa] Add release tag to path variable

kraushm:
The path for the PE is hard coded in the module and was missing the subfolder corresponding to the release tag.


lucamar:
The error causing the failure is the following:
```
== FAILED: Installation ended unsuccessfully (build directory: /tmp/jenscscs/435e82e/PrgEnvgnu/19.2/system-system-nocuda): build failed (first 300 chars): prgenv-gnu/19.2-nocuda(51):ERROR:105: Unable to locate a modulefile for 'foss/.2019b' (took 1 sec)
```

PR #1687, opened 2020-06-09
[tsa] Adding BUILD_ID to the build path in ProductionEB

lucamar:
The path will become unique enough to prevent the issue reported by @kraushm in #1684.

### DIFF
--- a/jenkins/JenkinsfileProductionEB
+++ b/jenkins/JenkinsfileProductionEB
@@ -80,7 +80,7 @@ stage('Build Stage') {
                                  """
                 }
 
-                withEnv(["EASYBUILD_BUILDPATH=$buildPathExpanded",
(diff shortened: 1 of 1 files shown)


jenkins/JenkinsfileProductionEB
```diff
@@ -80,7 +80,7 @@ stage('Build Stage') {
                                  """
                 }
 
-                withEnv(["EASYBUILD_BUILDPATH=$buildPathExpanded",
+                withEnv(["EASYBUILD_BUILDPATH=$buildPathExpanded/$env.BUILD_ID",
```
teojgo, line 83:
```suggestion
                withEnv(["EASYBUILD_BUILDPATH=$buildPathExpanded/${env.BUILD_ID}",
```

teojgo: Just so that it is clearer

PR #1688, opened 2020-06-10
[tsa] Add ddt 20.0.3

kraushm:


easybuild/easyconfigs/d/ddt/ddt-20.0.3-Redhat-7.0.eb
```diff
@@ -0,0 +1,56 @@
...
+To avoid conflict, module unload Visit Cube darshan
+Licensing : http://velan:4241/status.html"""
+
+toolchain = SYSTEM
+source_urls = ['http://content.allinea.com/downloads/']
+sources = ['arm-forge-{}{}-x86_64.tar'.format(version, versionsuffix)]
+
+postinstallcmds = [
+    'bash ./textinstall.sh --accept-licence %(installdir)s && '
+    'echo -e "type=2\nserial_number=6784\nhostname=velan\nserverport=4241" > '
```
jgphpc, line 27: may need to use the new serial number after we clarify the license contract but for now it will be fine.

PR #1700, opened 2020-06-16
[tsa] Align OpenMPI easyconfig files

kraushm:
Return to original module name and remove node specific variable setting.


victorusu:
@kraushm, `lgtm`. I agree with you. Since the files are independent of the software stack, I think it makes sense to remove them to avoid future confusion.

PR #1703, opened 2020-06-17
[dom-gpu] Add sourcing to default kernel

rsarm:


twrobinson:
As this is potentially a breaking change, I wonder if we should wait for 20.08? What do you think @rsarm

rsarm:
ok, let's move it to 20.08. We can test it there. I think for the default kernel there won't be any changes from the user's perspective. With custom kernels that weren't created with `kernel-create`, users need to be aware that there won't be any sourcing happening before launching jupyter.

lucamar:
@rsarm and @twrobinson Should we close this pull request, then?

rsarm:
@lucamar yes. I think so. I will open a new one for 20.08.

PR #1709, opened 2020-06-18
[daint-mc] Add Amber 20 CPU

victorusu:


jenkins-builds/7.0.UP01-19.10-mc
```diff
@@ -1,5 +1,6 @@
  Amber-18-14-14-CrayGNU-19.10.eb                    --set-default-module
```
gppezzi, line 1: if keeping 3 version on the supported list, please mention explicitly in the PR.

lucamar: @victorusu following @gppezzi's comment, may I suggest to remove at least `Amber-19-9-17-ff19SB-CrayGNU-19.10.eb` from the list? If I understood correctly, it's a version with a bug fix. The module won't be removed from the system unless you remove it manually as `jenscscs`.

victorusu: @gppezzi. sorry, I do not follow. Do I have to justify why we have three versions?
And @lucamar can we remove from the production list when we revisit which software version will go to the next PE installation? Because, if we install using PE20.06, on the current daint, I think it makes sense to keep all software currently in production.

lucamar: Thanks for your reply @victorusu, but I still believe that if the version `Amber-19-9-17-ff19SB-CrayGNU-19.10.eb` only provides bug fixes to the older default one, I think it wouldn't be any more necessary in production. I also think that when we update to `20.06` it makes sense to get rid of the older versions and keep the most recent ones only.

victorusu: @lucamar Amber19 is not just bug fixes, it is a complete new AmberTools version and we have users using this version.

victorusu: Amber18 != Amber19 != Amber20.
The only reason why we have the `ff19SB`in the name is that we have added a new force field into the Amber19 folder.

lucamar: I see, thanks for clarifying.

gppezzi: @victorusu my comment was just to confirm that it was your decision to have three versions, since you didn't mention anything it could have been just forgotten. 

And if you explain beforehand the reason why two version are not enough, it will speed up your PR in the future. Thanks!

PR #1710, opened 2020-06-22
[dom-gpu] Update jupyterlab-1.1.1-CrayGNU-19.10-batchspawner.eb

twrobinson:
Testing system monitor 0.6.0


twrobinson:
Closing. jupyterlab-system-monitor supports cpu from version 0.5 but these versions require JupyterLab 2.0 and will not be backported. It would be better to start to consider JupyterLab 2.0 deployment.

PR #1713, opened 2020-06-23
[dom-gpu, daint-gpu] eb py templates jupyterlab-1.2.4

henrique:


easybuild/easyconfigs/j/jupyterlab/jupyterlab-1.2.4-CrayGNU-19.10-cuda-10.1-batchspawner.eb
```diff
@@ -1,18 +1,12 @@
...
+easyblock = 'PythonBundle'
 
 name = 'jupyterlab'
 version = '1.2.4'
-_cudaversion = '10.1'
-versionsuffix = '-cuda-%s-batchspawner' % _cudaversion
+versionsuffix = '-cuda-%(cudashortver)s-batchspawner'
 
-_py_maj_ver = '3'
 _py_min_ver = '6'
```
ekouts, line 9:
```suggestion
_pyminver = '6'
```
Better change this, because I think `_pyminver` is what we are replacing in the ExperimentalEB

easybuild/easyconfigs/j/jupyterlab/jupyterlab-1.2.4-CrayGNU-19.10-cuda-10.1-batchspawner.eb
```diff
@@ -28,585 +22,198 @@ toolchain = {'name': 'CrayGNU', 'version': '19.10'}
...
-        'req_py_majver': '3',
-        'req_py_minver': '6',
-        'source_urls': ['https://pypi.python.org/packages/source/c/colorcet/'],
-        }),
-    ('zstandard', '0.13.0', {
-        'req_py_majver': '3',
-        'req_py_minver': '6',
-        'source_urls': ['https://pypi.python.org/packages/source/z/zstandard/'],
-        }),
+    ('traittypes', '0.2.1'),
```
ekouts, line 118: Is this dependency needed earlier? Because it's already in the recipe later in the list.

henrique, line 118:
```suggestion
```

henrique: jupyterlab 1.1 already used 'configurable-http-proxy', '4.2.1' (together with 'jupyterlab-nvdashboard', '0.2.1')
perhaps @twrobinson can shed some light on this
Do we need the old versions?

twrobinson: As this was never in production it is probably lagging in versioning. It should work with the configurable http proxy used in our 1.1.1 build. As for traittypes, it is a dep of itkwidgets and bqplot. Whichever occurs earlier in the eb file you can add it there. thanks!

henrique: Thx Tim
traittypes was my bad, we don't need it twice. All fixed now

victorusu:
`lgtm`
@twrobinson do you want to take a last look before we merge?

PR #1725, opened 2020-07-02
[daint] Remove termgraph from Daint production list

kraushm:


kraushm:
Fix for https://jira.cscs.ch/browse/UES-932

PR #1619, opened 2020-05-07
WIP [dom-gpu] ParaView for EasyBuild 4.2 and CLE7.UP01.PS09 

jfavre:
it uses the new Python template variables and the new cmake variables for the NVIDIA driver

COMPLETED: Installation ended successfully (took 20 min 35 sec)
== Results of the build can be found in the log file(s) /scratch/snx3000tds/jfavre/daint/software/ParaView/5.8.0-CrayGNU-19.10-EGL/easybuild/easybuild-ParaView-5.8.0-20200507.112609.log

PR #1726, opened 2020-07-03
[daint] Remove python variables from ParaView

ekouts:


jfavre:
Well done @ekouts . Thanks

ekouts:
@jfavre I only changed the python variables, but I just remembered you had an open PR ( #1619 ) fixing one of the recipes in order for them to build on dom as well. Should we make similar fixes in all the recipes? I only put daint as a target

jfavre:
LGTM

jfavre:
I'll review the recipe for Dom later since its environment is in constant change. "constant change". Is this an oxymoron?

PR #1730, opened 2020-07-15
[tsa] PE modules based on PGI 20.4

kraushm:
New PE based on most recent version of PGI. (see https://jira.cscs.ch/browse/LRIII-115)

In addition, all easyconfig files for this PE are now added to this repository.


kraushm:
retest this please

PR #1739, opened 2020-07-23
[dom-gpu] Set modules tool and metadata

lucamar:
Fix modules tool with TMOD 4 and Cudatoolkit version

### DIFF
--- a/easybuild/cray_external_modules_metadata-20.06.cfg
+++ b/easybuild/cray_external_modules_metadata-20.06.cfg
@@ -41,36 +41,10 @@ name = PETSc
 [cray-petsc-complex-64]
 name = PETSc
 
-[cray-python/2.7.15.6]
+[cray-python]
 name = Python
-version = 2.7.15.6
-prefix = /opt/python/2.7.15.6
-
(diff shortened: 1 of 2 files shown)


easybuild/cray_external_modules_metadata-20.06.cfg
```diff
@@ -0,0 +1,75 @@
...
+[cray-petsc-complex]
+name = PETSc
+
+[cray-petsc-complex-64]
+name = PETSc
+
+[cray-python]
+name = Python
+version = 3.8.2.1
+prefix = CRAY_PYTHON_PREFIX
```
ekouts, line 47:
```suggestion
```
Is the version and prefix necessary? Since the module has both the variables `CRAY_PYTHON_VERSION` and `CRAY_PYTHON_PREFIX` Easybuild should be able to get those variables from the module

lucamar: Thanks @ekouts, I just added them to see if they were fixing issues with Python recipes that failed. 
If EasyBuild can automatically detect both values, we can remove them.

PR #1741, opened 2020-07-24
[daint] Replace local variable _pyminver with the eb template parameter

ekouts:

PR #1742, opened 2020-07-24
[daint-gpu] Starting point for ipcmagic and PyExtensions

lucamar:
The recipe of `ipcmagic` was failing in a previous pull request #1741 and this pull request should fix it. 
I also include an updated recipe of `PyExtensions` provided by @ekouts in the branch `experimental`. 

I have removed the version of `cray-python` from the recipes, since the default is loaded: however, this might be wrong, so please feel free to suggest to re-insert the version.


lucamar:
I'm closing this pull request, I will address the issue in a new one, targeting directly `cdt/20.06` on Dom.

PR #1744, opened 2020-07-25
[dom-gpu] Revert "[daint] Replace local variable _pyminver with the eb template parameter"

lucamar:
Please revert eth-cscs/production#1741 as it breaks `ProductionEB`: after fixing #1742 the changes of this pull request can be merged again.


henrique:
`ProductionEB` was already broken in `1219` actually, and this build fails with the same EB issue of conflicting modules (with the old `PyExtensions/3.6.5.7` that shouldn't exist anymore), so perhaps we should look on why the production is failing first before reverting this?

lucamar:
Thanks for your comment @henrique: I'll close this pull request

PR #1745, opened 2020-07-27
[dom-gpu] Revert Dom back to modules 3.2

victorusu:
I think we should wait for Matteo to revert back on dom101.


teojgo:
lgtm, You can merge now.

PR #1746, opened 2020-07-28
WIP [dom] CLE7UP02 First batch

gppezzi:
This branch contains:
- Metadata files and base toolchains for 20.06
-  the 19.10 recipes with the modifications needed for the UP02 upgrade with 20.06
-- namely python and cuda versions removed from file name (default is set for a specific PE)
-- python 2 is removed (out of support)


easybuild/easyconfigs/j/jupyterlab/jupyterlab-1.1.1-CrayGNU-19.10-batchspawner.eb
```diff
@@ -110,22 +114,22 @@ exts_list = [
     ('colorcet', '2.0.2'),
     ('zstandard', '0.13.0'),
     ('itk-core', '5.0.1', {
-        'source_tmpl': 'itk_core-%(version)s-cp36-cp36m-manylinux1_x86_64.whl',
+        'source_tmpl': 'itk_core-%(version)s-cp38-cp38-manylinux1_x86_64.whl',
```
henrique, line 117:
```suggestion
        'source_tmpl': 'itk_core-%(version)s-cp%(pymajver)s%(pyminver)s-cp%(pymajver)s%(pyminver)s-manylinux1_x86_64.whl',
```
Ca we use %(pymajver)s%(pyminver)s here and below @twrobinson ? or is it not the python version?

lucamar:
We don't need to merge the modified 19.10 recipes, as we are now merging the new 20.06 recipes directly in production. We might need to merge `JenkinsfileExperimentalEB` and `util.groovy` only for the automated build: since we plan to merge only few files, it makes sense to close this pull request and open a new one instead.

PR #1751, opened 2020-07-30
WIP [dom] Bulk 20.06 recipes based on experimental 19.10 branch

gppezzi:
Creating base recipes for 20.06 (This branch won't be merged as it is)


lucamar:
We don't need to merge the modified `19.10` recipes, as we are now merging the new `20.06` recipes directly in production. We might need to merge `JenkinsfileExperimentalEB` and `util.groovy` only for the automated build.

PR #1792, opened 2020-08-03
[dom-gpu] Remove old versions of applications

lucamar:
Removing old versions of CP2K LAMMPS and VASP as agreed for the update.


lucamar:
@victorusu What about removing `GROMACS-2018.6-CrayGNU-20.06.eb` for the list `dom-mc`?

victorusu:
> @victorusu What about removing `GROMACS-2018.6-CrayGNU-20.06.eb` from the list `dom-mc`?

@lucamar, I think we can safely remove it. But we haven't announced it before.

PR #1818, opened 2020-08-06
[dom] Updated ipcmagic recipe and production lists

lucamar:
The new recipe uses the current metadata file and loads external dependencies without versions. 

I don't understand why `ipcmagic` builds the extension `ipyparallel 6.3.0`, while we have a stand-alone modulefile ` ipyparallel/6.2.4-CrayGNU-20.06-python3` in the production list. Any idea?

### DIFF
--- a/easybuild/easyconfigs/i/ipcmagic/ipcmagic-0.1-CrayGNU-20.06.eb
+++ b/easybuild/easyconfigs/i/ipcmagic/ipcmagic-0.1-CrayGNU-20.06.eb
@@ -1,0 +1,43 @@
+# modified by Luca Marsella (CSCS)
+easyblock = "PythonPackage"
+
+name = 'ipcmagic'
+version = '0.1'
+
+homepage = 'https://github.com/eth-cscs/jupyter-utils'
+description = "Utilities for working with Jupyter at CSCS"
+
+toolchain = {'name': 'CrayGNU', 'version': '20.06'}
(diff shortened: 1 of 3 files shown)


teojgo:
> The new recipe uses the current metadata file and loads external dependencies without versions.
> 
> I don't understand why `ipcmagic` builds the extension `ipyparallel 6.3.0`, while we have a stand-alone modulefile ` ipyparallel/6.2.4-CrayGNU-20.06-python3` in the production list. Any idea?

It is in the exts list

lucamar:
> > The new recipe uses the current metadata file and loads external dependencies without versions.
> > I don't understand why `ipcmagic` builds the extension `ipyparallel 6.3.0`, while we have a stand-alone modulefile ` ipyparallel/6.2.4-CrayGNU-20.06-python3` in the production list. Any idea?
> 
> It is in the exts list

Thanks @teojgo, I have noticed that it is in the extension list, but I don't understand why, since we have a stand-alone modulefile.

henrique:
> Thanks @teojgo, I have noticed that it is in the extensions list, but I don't understand why, since we have a stand-alone modulefile.
> I guess that the stand-alone modulefile was created before the extension, since the latter is more recent: if this is the case, we could probably get rid of `ipyparallel/6.2.4-CrayGNU-20.06-python3` from production.

good call, +1

PR #1822, opened 2020-08-06
[dom-gpu] VTK with CrayGNU 20.06 EGL

lucamar:
The pull request provides a draft recipe for VTK with `CrayGNU 20.06` and the `EGL` library.

Unfortunately, the library is available under the name `libEGL_nvidia.so.0` and not as `libEGL.so` as requested by the build process and the build fails with the following error:
```
make[2]: *** No rule to make target '/opt/cray/nvidia/default/lib64/libEGL.so', needed by 'lib64/libvtkglew-9.0.so.9.0.0'.  Stop.
```

### DIFF
--- a/easybuild/easyconfigs/v/VTK/VTK-9.0.0-CrayGNU-20.06-EGL-python3.eb
+++ b/easybuild/easyconfigs/v/VTK/VTK-9.0.0-CrayGNU-20.06-EGL-python3.eb
@@ -1,13 +1,12 @@
-##
+#
 # This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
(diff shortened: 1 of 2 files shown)


lucamar:
Even if I fix the configuration option of `CMake` and point to `libEGL_nvidia.so.0` instead of `libEGL.so`, still the library `libOpenGL.so` that is required in the configure steps is missing. Therefore I have opened CSCS RT #39877.

lucamar:
I have adapted the configure EGL and OpenGL library path to use `usr/lib64` instead of `/opt/cray/nvidia/default/lib64`, as in `ParaView-5.8.0-CrayGNU-20.06-EGL-python3.eb`.

PR #1826, opened 2020-08-07
[dom] Jupyterlab with cdt 20.06

lucamar:
Draft recipe for `juypterlab` with `cdt/20.06`.


twrobinson:
Hi, this is not needed. We already have "-batchspawner" in the production lists.

lucamar:
Thanks for the commen @twrobinson, the "-batchspawner" version is already in the production list of Dom: it might be worth removing this recipe from the production list of Piz Daint, if it is not needed.

PR #1827, opened 2020-08-07
[dom-mc] Visit 3.1.2 for cdt/20.06

lucamar:
The pull request provides a draft recipe for `Visit` 3.1.2, with a tentative dependency `CMake 3.14.5` and trying to use the default `gcc 8.3.0` instead of `gcc 7.3.0` as in the previous version `Visit 3.1.0` on Piz Daint. 
If needed, we can use the system `/usr/bin/gcc 7.5.0` on Dom and build an older `CMake` modulefile.

### DIFF
--- a/easybuild/easyconfigs/v/Visit/Visit-3.1.0-CrayGNU-20.06-python3.eb
+++ b/easybuild/easyconfigs/v/Visit/Visit-3.1.0-CrayGNU-20.06-python3.eb
@@ -1,0 +1,94 @@
+# contributed by Jean Favre (CSCS)
+easyblock = 'CMakeMake'
+
+name = 'Visit'
+version = "3.1.2"
+versionsuffix = '-python%(pymajver)s'
+
+homepage = 'https://wci.llnl.gov/codes/visit/'
+description = """VisIt is an Open Source, interactive, scalable, visualization,
+animation and analysis tool."""
+
+toolchain = {'name': 'CrayGNU', 'version': '20.06'}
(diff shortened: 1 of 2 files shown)


jfavre:
As far as I remember, VisIt will not build with CMake 3.14.5. That is why the Visit *custom* build recipe set cmake to be 3.9.
The newer Visit was supposed to be release around now, so I need more time to verify availability and feasability on daint, and perhaps we'll jump directly to it. Note that his is less urgent that deploying other visualization apps, so we will converge less quickly to the final solution.

lucamar:
Thanks for the comment @jfavre: I have tried a tentative recipe with the latest release that I could find of `Visit`, to see if a more recent version of `CMake` could be used to build it. 
`Visit` this is the last supported application that is left to build on Dom with the Cray PE 20.06, while `ParaView` and `VTK` have already been adapted and built: see https://jira.cscs.ch/browse/UES-953
If `Visit` is not urgent, I will close this pull request and the corresponding issues, so it will not appear in the production list used on Dom to test the upgrade to Cray PE 20.08.

jfavre:
VisIt will not build with any version of cmake that is different than what the build_visit script requires, and that version is 3.9.3
Further VisIt 3.1 is not ready for python3. That is for the upcoming version.
Further, the current version of the EB recipe works only  if we have gcc 7.3,which we do not have on cdt/20.06 as I understand. This is why I need much more time, and we should not block the release cycle with the rest of the production list. I agree to close the corresponding PR

PR #1829, opened 2020-08-10
[dom-gpu] Default module for Horovod

victorusu:
There is no default module for Horovod.
This is in principle ok, but it is not symmetrical to the other applications that always have a default.
On daint, for instance, it shows...
```
module av Horovod
...
Horovod/0.16.4-CrayGNU-19.10-tf-1.14.0 Horovod/0.18.1-CrayGNU-19.10-tf-2.0.0  Horovod/0.19.1-CrayGNU-19.10-tf-2.2.0
```


lucamar:
retest this please

PR #1834, opened 2020-08-11
[dom-gpu] Cray PE 20.08 metadata file

lucamar:
The metadata file is needed by the Jenkins project `UpdateEB` to update the software stack.

### DIFF
--- a/easybuild/cray_external_modules_metadata-20.08.cfg
+++ b/easybuild/cray_external_modules_metadata-20.08.cfg
@@ -1,0 +1,75 @@
+# metadata useful to EasyBuild (hpcugent.github.io/easybuild) for modules provided by Cray on Cray systems
+# authors: Guilherme Peretti-Pezzi (CSCS), Petar Forai (IMP/IMBA), Kenneth Hoste (HPC-UGent), Eirini Koutsaniti (CSCS)
+
+[cray-fftw]
+name = FFTW
+
+[cray-hdf5]
+name = HDF5
+version = 1.12.0.0
+prefix = HDF5_DIR
+
+[cray-hdf5-parallel]
+name = HDF5
+version = 1.12.0.0
+prefix = HDF5_DIR
+
+[cray-libsci]
+name = LibSci
+version = 20.06.1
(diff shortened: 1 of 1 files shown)


easybuild/cray_external_modules_metadata-20.08.cfg
```diff
@@ -0,0 +1,75 @@
+# metadata useful to EasyBuild (hpcugent.github.io/easybuild) for modules provided by Cray on Cray systems
+# authors: Guilherme Peretti-Pezzi (CSCS), Petar Forai (IMP/IMBA), Kenneth Hoste (HPC-UGent), Eirini Koutsaniti (CSCS)
+
+[cray-fftw]
+name = FFTW
+
+[cray-hdf5]
+name = HDF5
+version = 1.12.0.0
```
victorusu, line 9: I do not think we should set the version. EB should be able to take it

easybuild/cray_external_modules_metadata-20.08.cfg
```diff
@@ -0,0 +1,75 @@
...
+prefix = HDF5_DIR
+
+[cray-hdf5-parallel]
+name = HDF5
+version = 1.12.0.0
```
victorusu, line 14: do not think we should set the version. EB should be able to take it

easybuild/cray_external_modules_metadata-20.08.cfg
```diff
@@ -0,0 +1,75 @@
...
+prefix = HDF5_DIR
+
+[cray-libsci]
+name = LibSci
+version = 20.06.1
```
victorusu, line 19: do not think we should set the version. EB should be able to take it

easybuild/cray_external_modules_metadata-20.08.cfg
```diff
@@ -0,0 +1,75 @@
...
+prefix = CRAY_LIBSCI_PREFIX
```
victorusu, line 20: do not think we should set the prefix. EB should be able to take it

PR #1841, opened 2020-08-12
[dom-gpu] nvhpc/cdt-20.08

jgphpc:


easybuild/easyconfigs/n/nvhpc/nvhpc-2020_207-cuda-10.2.eb
```diff
@@ -0,0 +1,80 @@
+# jg (CSCS)
+easyblock = 'Tarball'
+
+name = 'nvhpc'
+version = '2020_207'
+# this recipe installs cuda, do NOT change this line:
+local_cudaversion = '10.2'
+versionsuffix = '-cuda-%s' % local_cudaversion
```
lucamar, line 8:
```suggestion
versionsuffix = '-cuda'
```

jgphpc: By default, nvhpc will install latest cuda:
```
Couldn't find a CUDA version based off the driver. Pick a default:
No selection/hitting enter will choose CUDA 11.0
1) CUDA 11.0
2) CUDA 10.2
3) CUDA 10.1

Please choose an option [1-3]: 2
```
We need 10.2, not 11.0 hence i must add the version in the recipe

jgphpc:
I will make a new pr to avoid conflicts

PR #1856, opened 2020-08-13
[dom] use - instead of _ for Intel tools

jgphpc:
Fix for @lucamar comment
Will update the jenkins list in a separate pull request


easybuild/easyconfigs/a/advisor/advisor-2020_update2.eb
```diff
@@ -28,11 +28,12 @@ sanity_check_paths = {
 allow_prepend_abs_path = 'True'
 modextrapaths = {
     'PATH': '%(name)s/bin64',
+    'INTEL_PYTHONPATH': '%(name)s/pythonapi',
     'PYTHONPATH': '%(name)s/pythonapi',
     'INTEL_LICENSE_FILE': '/opt/intel/licenses',
 }
 modextravars = {
-    'ADVISOR_2019_DIR': '%(installdir)s/%(name)s',
+    'ADVISOR_2020_DIR': '%(installdir)s/%(name)s',
```
ekouts, line 36:
```suggestion
    'ADVISOR_%(version)s_DIR': '%(installdir)s/%(name)s',
```
Would this work? So that you don't have to change it manually in the future?

jgphpc: yes, i'll do it next time

PR #1858, opened 2020-08-13
[dom] Update the packages of PyExtensions

teojgo:


teojgo:
> why two sets of ext_lists?

Oh by mistake

twrobinson:
What's the need for downgrading six compared to the cray provided one? Ans zipp - cray's zipp installed as 0.0.0 but it probably isn't...  (likely a setuptools cock up)

teojgo:
> why two sets of ext_lists?
> also, why cython when cray-python already has a newer version, is it a deliberate downgrade? (edit: my mistake it is not newer. but is it new enough or do we need the newer one?)

The cython that is already there does not compile the latest pandas. It neads at least 0.29.16 for that

teojgo:
> What's the need for downgrading six compared to the cray provided one? And as for zipp - cray's zipp installed as 0.0.0 but it probably isn't... (likely a setuptools cock up)

Should I remove six and zipp completely?

twrobinson:
> > What's the need for downgrading six compared to the cray provided one? And as for zipp - cray's zipp installed as 0.0.0 but it probably isn't... (likely a setuptools cock up)
> 
> Should I remove six and zipp completely?

I reckon you can probably remove six. zipp is problematic... I think because it installed as 0.0.0 then other packages won't consider it being matched when doing dependency analysis. It is a cray bug really, due to building with source and old setuptools (I think).

lucamar:
I take the opportunity to comment here, since we are discussing the Python extensions to be added or removed. 
The module `PyExtensions` on Dom depends on `cray-python/3.8.2.1`, which provides `dask 2.13.0`.
However, we build `dask 2.2.0` as a stand-alone modulefile as a dependency of `jupyterlab`: Why do we need to build a much older release? Jupyterlab does not support newer versions of `dask`?

twrobinson:
> I take the opportunity to comment here, since we are discussing the Python extensions to be added or removed.
> The module `PyExtensions` on Dom depends on `cray-python/3.8.2.1`, which provides `dask 2.13.0`.
> However, we build `dask 2.2.0` as a stand-alone modulefile and dependency of `jupyterlab`.
> Why do we need to build a much older release? Jupyterlab does not support newer versions of `dask`?

Yep, I noticed that. I will comment in UES-1035.

twrobinson (approved):
I think this can be merged. We will need to double check that the python modules it provides are compatible with all the applications, but as a zeroth order approximation I say merge.

PR #1865, opened 2020-08-14
[dom-gpu] Adding ELPA with GPU support

lucamar:
I'm adding the recipe for the `ELPA` library with GPU support and updating the recipe of `CP2K` where it is currently used as a dependency.
Extending the new `ELPA` recipe to build it on the current Cray PE 19.10 of Piz Daint is straightforward: the library can be compiled using the EasyBuild flag `--try-toolchain-version=19.10`.

### DIFF
--- a/easybuild/easyconfigs/c/CP2K/CP2K-7.1-CrayGNU-20.08-cuda.eb
+++ b/easybuild/easyconfigs/c/CP2K/CP2K-7.1-CrayGNU-20.08-cuda.eb
@@ -32,7 +32,7 @@ builddependencies = [
     ('flex', '2.6.4'),
 ]
 dependencies = [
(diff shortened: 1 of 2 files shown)


lucamar:
Linking `ELPA` with gpu support could be also envisaged for `SIRIUS`: linking the new library will not change the behaviour of the code, as described in the Wiki: 
> If ELPA is build with GPU support, users can choose at RUNTIME, whether to use the GPU version or not.

Only if the configure option `--with-gpu-support-only` is used when building `ELPA`, then the use of the GPU version of the library will be enforced at runtime:
```
--with-gpu-support-only | Compile and always use the GPU version
```

See https://gitlab.mpcdf.mpg.de/elpa/elpa/-/wikis/INSTALL for more information.

PR #1862, opened 2020-08-13
[daint-gpu] jupyterlab debugging

twrobinson:
Do not merge

PR #1875, opened 2020-08-18
[daint-gpu] Force remove cray-libsci_acc

omlins:
Force remove the `cray-libsci_acc` module as it is not needed by `CUDA.jl` or Julia in general, but it breaks the jupyterlab module (see #1862).


omlins:
retest this please

PR #1876, opened 2020-08-19
[daint-gpu] jupyterlab-1.2.16-CrayGNU-20.08-batchspawner-cuda.eb

twrobinson:


omlins (approved):
approved to enable rebuild, but you need to undo this change later again!

PR #1878, opened 2020-08-19
[daint-gpu] Julia 1.5.0 without craype-accel-nvidida60 dependency

omlins:
This PR removes the craype-accel-nvidida60 dependency from Julia 1.5.0 as it might be the reason for issues with Dask and numpy at runtime in Jupyterlab.


twrobinson:
We can close this one now that we confiorm an unload of libsci_acc is sufficient, right @omlins ?

omlins:
> We can close this one now that we confirm an unload of libsci_acc is sufficient, right @omlins ?

yes

PR #1888, opened 2020-08-21
[dom-gpu] Update ELPA dependency of SIRIUS

lucamar:
Linking `ELPA` with GPU support as already done for `CP2K`: linking the new library will not change the behaviour of the code, as described in the `ELPA` Wiki:
> If ELPA is build with GPU support, users can choose at RUNTIME, whether to use the GPU version or not.

Only if the configure option `--with-gpu-support-only` is used when building `ELPA`, then the use of the GPU version of the library will be enforced at runtime:
```
--with-gpu-support-only | Compile and always use the GPU version
```
See https://gitlab.mpcdf.mpg.de/elpa/elpa/-/wikis/INSTALL for more information.


toxa81:
Thanks @lucamar ! I have nothing against this changes, I just have a strong feeling that dealing with ELPA is hopeless :) It never works as expected.

toxa81 (approved):

PR #1890, opened 2020-08-22
[dom]  jupyterlab-2.0.2-CrayGNU-20.08-batchspawner-cuda.eb

twrobinson:
This is JupyterLab 2.0.2. 
JupyterLab 2.x is required for a fix for ipyparaview zooming (not backported to jupyterlab 1.x). It requires the update of some python modules and widgets, as jupyterlab extensions are incompatible between jupyterlab 1.x and 2.x. I have tried to keep as many python modules the same as for previous versions, but because of some forced changes and bug fixes there are a few updates, which are commented in the recipe.


twrobinson:
retest this please

jfavre (approved):

PR #1892, opened 2020-08-25
[daint-mc] Add OpenFOAM-7 and SCOTCH 6.0.9

kraushm:


lucamar (approved):
Thanks @kraushm: let's try to use `cray-tpsl` for `SCOTCH` in the future, though

PR #1895, opened 2020-08-26
[daint-mc] Changing SCOTCH dependency of OF7

kraushm:
Changing the dependencies to use the SCOTCH library from the `cray-tpsl` module, instead of building SCOTCH apart.


lucamar (approved):
Thanks @kraushm

PR #1897, opened 2020-08-28
WIP: COSMO-POMPA test ACC fix

omlins:
This is not a PR to merge but only to share the recipe to test the OpenACC fix.


victorusu:
@omlins, are we still deploying recipes using 20.08?

omlins:
Let's keep it open until we have a working recipe for 21.05

PR #1898, opened 2020-08-28
[daint-gpu,dom-gpu] provided the correct LD_LIBRARY_PATH 

jfavre:
thanks to Victor for his tip

### DIFF
--- a/easybuild/easyconfigs/m/mdl/mdl-317500.2554-CrayGNU-19.10.eb
+++ b/easybuild/easyconfigs/m/mdl/mdl-317500.2554-CrayGNU-19.10.eb
@@ -20,8 +20,7 @@ sanity_check_paths={
 }
 
 # set custom LD_LIBRARY_PATH
-modextravars = { 'LD_LIBRARY_PATH':'%(installdir)s/linux-x86-64/lib:$::env(LD_LIBRARY_PATH)', 
-               }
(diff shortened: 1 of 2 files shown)


victorusu (approved):
`lgtm`

victorusu:
The failure is associated to the lack of 20.08 on Piz Daint. So, I am merging...

jfavre:
builds have failed because I was too pretentious and thought jenkins was smart enough to check that one recipe marked 19.10 was for daint to test, and the other marked 20.08 was for dom to test. I guess I should have split them into two different PRs. Advice?

victorusu:
@jfavre, yes! It is better to split the recipes into two in this case.
I merged also because I have also tested it manually on dom and it works.
I could not wait to see for myself the fix for the `LD_LIBRARY_PATH` generation... 😂

lucamar:
> builds have failed because I was too pretentious and thought jenkins was smart enough to check that one recipe marked 19.10 was for daint to test, and the other marked 20.08 was for dom to test. I guess I should have split them into two different PRs. Advice?

Thanks @jfavre, I also agree with @victorusu, splitting would be the best approach for the next pull requests, since the current one has been already merged by @victorusu. 
Every system that you specify in the title of the pull request will build every script that you add to the pull request, therefore it is up to you to verify that the each build makes sense.

PR #1901, opened 2020-08-30
[daint-gpu] Fix srcdir for Dycore-551f6b4 cdt-19.08

teojgo:


lucamar:
@omlins I think that we can merge the pull request, what do you think?

lucamar (approved):
Thanks @teojgo

omlins (commented):
@lucamar, I discussed it with Theo offline and I think you should decide together with him if this should be merged as you introduced this change.

lucamar:
> @lucamar, I discussed it with Theo offline and I think you should decide together with him if this should be merged as you introduced this change going from 20.06 to 20.08.

I changed that line, since it was not working otherwise for me: as it worked also with the recipes that were built successful in this pull request, I think it is safe to merge it.

PR #1904, opened 2020-09-01
[dom] CPMD running with old Intel compiler

ajocksch:


lucamar (changes requested):
Thanks @ajocksch: since you use the toolchain `CrayIntel`, you don't need `PrgEnv-Intel` in the dependency list. Please also remove the old Intel module from the dependencies and try using `prebuildopts = module switch intel intel/19.0.1.144 && ` instead.

lucamar (approved):
Thanks @ajocksch

PR #1905, opened 2020-09-01
[dom-gpu] COSMO-POMPA with cdt 20.08

lucamar:
This is a work in progress tentative to unify the recipes of COSMO-POMPA (cpu and gpu), since we test the executables together. If this approach succeeds, it might make sense to do the same for the `cordex` executables.


kraushm (approved):
lgtm.

lucamar:
The build of `COSMO` and all the dependencies succeeded, the second build of `Serialbox` alone failed instead: see https://jenkins.cscs.ch/blue/organizations/jenkins/TestingEB/detail/TestingEB/4150/pipeline.

omlins:
@lucamar :
For me the Serialbox build failed when in the past days, because I had set
```
export INSTALLDIR=$SCRATCH/cosmo_pe2008
export EASYBUILD_BUILDPATH=$INSTALLDIR
```
Once, I did not set `EASYBUILD_BUILDPATH` anymore to use the default it worked. Looking at the log of the failing build that you noted, it seems to have happened about at the same place. I remember that the error message was not at all pointing to the buildpath problem. So, I would suppose here it is also a buidpath problem somehow when building with Jenkins.

lucamar:
Thanks for your comment @omlins: there could be some bugs related to the build path. Let's see anyway if removing parallel make would change the outcome of multiple builds within the same pull request.

lucamar:
Removing parallel threads of make for `Serialbox` worked twice in a row, with `Serialbox` built twice each time: I suspect that the parallel build of the tool is not robust... well, it makes sense for a tool called `Serailbox` ;-9

gppezzi:
👍  for attempting to remove the usage of two PrgEnv's!

I don't know if the users validated this approach already, so my only concern in this PR is that it is modifying existing recipes and it could potentially cause conflict with the working Cosmo recipes. You are also deleting a patch. 

Maybe adding a new version suffix is safer?

omlins (changes requested):
That looks great!
I agree with Guilherme that no existing file should be deleted, i.e. the deleted patch should be restored.
Besides that, do I understand right, Luca, that you abandoned the idea of merging also CPU and GPU recipes. To my understanding, this would not be possible.

PR #1906, opened 2020-09-02
[dom-gpu] Fix integer overflow on xalt data

victorusu:
This fixes an integer overflow on XALT
Introduces the '?###?' separator for the symboles
Removes symbols that start with '.' or '__'
Strips the ' [code .*]' entries at the end of the function names


kraushm (approved):
lgtm.

PR #1909, opened 2020-09-04
[dom] CPMD 4.1

ajocksch:
we set CPMD 4.1 as default since our test case is executed faster with the old version

UES-1052


lucamar (approved):
Thanks @ajocksch

PR #1914, opened 2020-09-07
[WIP] [dom-gpu] COSMO-pompa with ACC fix for PE 20.08

omlins:
This uses the following new source: https://github.com/C2SM-RCM/cosmo-pompa/tree/crclim_cray_acc_fix


victorusu:
@omlins, are we still deploying recipes using 20.08?

omlins:
Let's keep it open until we have a working recipe for 21.05.

PR #1917, opened 2020-09-08
[dom-gpu] cpmd + pat-20.08.0

ajocksch:
This is CPMD with the most recent CrayPAT version.


lucamar (changes requested):
Thanks @ajocksch: please have a look at https://github.com/eth-cscs/production/pull/191 for the naming scheme. We don't need to use any longer the perftools version in the file name and in the `builddependencies`.

lucamar (approved):
Thanks @ajocksch

PR #1938, opened 2020-09-15
[daint,dom] jupyterlab-1.2.16-CrayGNU-20.08-batchspawner-cuda.eb

twrobinson:
This PR adds the plotly labextension, which was missing from this build. The production builds are already fixed by hand, so this PR should be merged, but should not be force installed.


twrobinson:
retest this please

rsarm (approved):

PR #1940, opened 2020-09-16
[daint-gpu] Unhide CMake when defined as a dependency

victorusu:
This is a proposal to remove CMake from the hidden dependencies... It makes no sense to have CMake hidden since it is shown to the users as a module on the systems.

It is a bit difficult to see the change in the file...
But I have removed CMake from the list of `hide-deps`.


gppezzi (approved):
I'm fine with this, but we should agree on a policy for avoiding an explosion of cmake versions on the module list and reviewers should check if PRs are not adding random CMake versions to production (if people use try-toolchain we can easily add old versions without noticing)

lucamar:
retest this please

kraushm (approved):
lgtm.

gppezzi:
As usual when changing the hide-deps list, the only caveat is that we might see conflicts at some point when adding new recipes that depend on cmake: if the new recipe has dependencies that are already deployed and have a hidden cmake as dependency. 

If cmake is always used as build dependency, we should be good.

> This is a proposal to remove CMake from the hidden dependencies... It makes no sense to have CMake hidden since it is shown to the users as a module on the systems.

Well just for the record, it actually made sense in our first deployments when we didn't bother to try to match with the existing CMake version before adding a new recipe and we had several CMake versions for every toolchain + dummy toolchain.

lucamar (commented):
In principle I agree that we can have the CMake modules visible, but then we should check that we are not building additional CMake versions when it is not needed and correct the recipes, since the automated update will keep it. For instance, I see that the recipe of GROMACS still uses CMake 3.14.1, while the default 3.14.5 works as well.

PR #1942, opened 2020-09-16
[daint, dom] Integration of ADIOS with Cray wrappers and refinement

omlins:


omlins:
It successfully builds using the wrappers - all the compile an linking flags are added automatically:
omlins@dom101:~/10_io_perf/0_adios2_helloworld> cc helloBPWriter.c
omlins@dom101:~/10_io_perf/0_adios2_helloworld> CC helloBPWriter.cpp
omlins@dom101:~/10_io_perf/0_adios2_helloworld> ftn helloBPWriter.F90
omlins@dom101:~/10_io_perf/0_adios2_helloworld> CC helloBPWriter.c -fpermissive
helloBPWriter.c: In function 'int main(int, char**)':
helloBPWriter.c:54:22: warning: invalid conversion from 'void*' to 'float*' [-fpermissive]
     myFloats = malloc(sizeof(float) * Nx);
                ~~~~~~^~~~~~~~~~~~~~~~~~~~
omlins@dom101:~/10_io_perf/0_adios2_helloworld>

jfavre (changes requested):
should moduleclass of blosc be changed from 'vis' to 'tools' to match similar other modules such as the bzip2?

Other than that, it lgtm

omlins:
> should moduleclass of blosc be changed from 'vis' to 'tools' to match similar other modules such as the bzip2?
> 
> Other than that, it lgtm

@gppezzi, what do you think?

Besides that, note that this is still WIP as I am adding more compression codecs.

omlins:
What is missing is to set the `moduleclass` - maybe `lib` could also be used... Besides that, everything should be fine.

omlins:
@jfavre and @gppezzi : with the exception of the moduleclass issue it now it is completely ready for merging from my point of view. Please share what you think about this issue (when to use tools, vis and lib) in order to merge it afterwards.

easybuild/easyconfigs/a/adios/adios-2.6.0-CrayGnu-20.08.eb
```diff
@@ -1,4 +1,4 @@
-# contributed by Jean Favre (CSCS)
+# contributed by Jean Favre (CSCS); extended by Samuel Omlin (CSCS)
```
gppezzi, line 1: I've just noticed that the file name does not match the toolchain (case): CrayGNU

gppezzi (approved):

easybuild/easyconfigs/a/adios/adios-2.6.0-CrayGnu-20.08.eb
jfavre: @omlins Can you fix the filename? Thanks

omlins: Done.

omlins:
Concerning the moduleclasses: I think the primary purpose of all the software dealt with in this PR with is to be used as a library. Thus, I changed the module class for all to `lib`. I hope you agree with this. Besides that @gppezzi pointed out that this is essentially cosmetics as long as we don't use lmod and does not have any practical consequences for us.

omlins:
@jfavre please let us know if you agree with this.

jfavre (approved):

PR #1944, opened 2020-09-16
[daint-gpu] VASP CUDA recipe with perftools-lite

lucamar:
EasyBuild recipe of VASP CUDA with perftools-lite (default from `cdt/20.08`).


victorusu (approved):
`lgtm`

PR #1945, opened 2020-09-16
[daint-mc] VASP recipe with perftools-lite

lucamar:
EasyBuild recipe of VASP with `perftools-lite` (default from `cdt/20.08`).


lucamar:
Same error as for #1946, quota reached on `$HOME` of user `jenscscs`: 
```
WARNING: PerfTools is saving object files from a temporary directory into directory '/users/jenscscs/.craypat/vasp/60262'
cp: error writing '/users/jenscscs/.craypat/vasp/60262/profiling.o': Disk quota exceeded
```

lucamar:
retest this please

twrobinson (approved):
Looks good. I tested the new module and the performance report was written correctly. The run time for a simple test case was the same as the non-perftools version in production, and energies, wavefunctions etc identical.

PR #1947, opened 2020-09-16
[daint-mc] QuantumESPRESSO recipe with perftools-lite

lucamar:
EasyBuild recipe of QuantumESPRESSO with `perftools-lite` (default from `cdt/20.08`).


henrique (approved):
lgtm

PR #1950, opened 2020-09-18
[daint-mc] GROMACS double precision

victorusu:


lucamar (approved):
Thanks @victorusu: does it make sense to add the double precision executables of GROMACS in the supported modules as well, using additional iterations of the EasyBuild recipe?

victorusu:
@lucamar, I do not know. We can also do that...

PR #1951, opened 2020-09-18
[daint-gpu] switch to SIRIUS-6.5.6

toxa81:
* Add new recipe for SIRIUS
 * bump version of SIIRUS in QE EB file


lucamar (approved):
Thanks @toxa81: as discussed, I suggest to re-introduce SIRIUS version in the suffix: in this way, you will be able to update the SIRIUS dependency of the same QE-SIRIUS release tag, without overwriting the existing installation. The module name is longer with `SIRIUS` appearing twice (name of the recipe and suffix), but that makes the updates clean

PR #1952, opened 2020-09-18
[daint-gpu] Remove Sirius from hide-deps in system_wide.cfg 

lucamar:
Remove SIRIUS from the hide list: it will be visible when built as a dependency.


toxa81:
Hi @gppezzi ! I forgot why sirius was made hidden. It is a blocker now to update its version and rebuild QE. The proposal is to make the SIRIUS module normal again.

toxa81 (approved):
looks good

PR #1954, opened 2020-09-18
[dom daint] Add MATLAB to production lists

lucamar:
MATLAB recipe on `daint-gpu`, `daint-mc`, `dom-gpu` and `dom-mc` production lists (the latter to keep the systems in sync).


kraushm (approved):
lgtm.

PR #1955, opened 2020-09-18
[daint-mc] Include double precision build to GROMACS non GPU

victorusu:
@lucamar, this implements your suggestion to include the double precision build in the "normal" GROMACS build.


lucamar (approved):
Thanks @victorusu

PR #1957, opened 2020-09-21
[daint-gpu dom-gpu] Latest QE-SIRIUS in production

lucamar:
The pull request adds the recipe `QuantumESPRESSO-SIRIUS-6.5-rc4-CrayIntel-20.08-SIRIUS-6.5.6-cuda.eb` to the GPU production lists of Dom and Piz Daint, setting it as the new default and removing the old build. 
The old modulefile will still be available as non default: the documentation might need to be updated.


toxa81:
this has to be dropped in favor of GNU-based EB recipie

lucamar:
Thanks for your comment @toxa81, I'm closing the pull request and deleting the corresponding branch.

PR #1960, opened 2020-09-21
[daint-gpu dom-gpu] QE/SIRIUS

toxa81:
This is the update of all EB recipes related to QE and SIRIUS

* old QE and SIRIUS recopies move to `deprecated` folder
* SIRIUS-6.5.7 is now the default and stable version
* QuantumESPRESSO-SIRIUS-6.5-rc4-CrayGNU-20.08-cuda.eb was changed to use SIRIUS-6.5.7
* Support of Intel programming environment EBs is dropped, from now on only GNU programming environment is supported


lucamar (approved):
Thanks @toxa81, please let me remind you that:
- the old module with Intel is removed from the production list, but it will be available as non-default on the system until we remove it manually 
- there is no reference to the version of `SIRIUS` in the name of the new modulefile, therefore if you updated SIRIUS without changing QE-SIRIUS version, you will have to overwrite the existing module

toxa81:
> Thanks @toxa81, please let me remind you the following important notes:
> 
> * the old module QE-SIRIUS built with `CrayIntel` is removed from the production list, but it will still be available as non-default module on the system, until we remove it manually

@lucamar How you manage the updates of dependencies? Let's say there is a hot-fix in libxc. All codes have to update the dependency. But the codes themselves didn't change. How this situation is handled in EB?

> * there is no reference to the version of SIRIUS in the name of the new modulefile; therefore, if you update SIRIUS without changing QE-SIRIUS version, you will have to overwrite the existing module in production
That is probably fine.

lucamar:
> @lucamar How you manage the updates of dependencies? Let's say there is a hot-fix in libxc. All codes have to update the dependency. But the codes themselves didn't change. How this situation is handled in EB?

We force the re-build of the codes with the EasyBuild flag `-f` in the Jenkins project ProductionEB, after we update the necessary dependencies. 

Talking about dependencies, can we remove the commented lines with dependencies `ELPA` and `magma` in the recipe of SIRIUS?

PR #1963, opened 2020-09-23
[daint-mc] QuantumESPRESSO 6.6

lucamar:
I provide the recipe to build the latest release of QuantumESPRESSO (multicore only): I have linked `HDF5` and `libxc`: I have tested `pw.x` with `AUSURF112` and the performance is comparable to the previous release.


toxa81 (approved):
Looks good! I wish they had CMake build system

PR #1967, opened 2020-09-24
[daint-gpu, dom-gpu] remove an old reference to a library

jfavre:
this addresses https://webrt.cscs.ch/Ticket/Display.html?id=40226


lucamar (approved):
Thanks @jfavre

PR #1969, opened 2020-09-25
[dom-mc] Update jupyterlab-2.0.2-CrayGNU-20.08-batchspawner.eb

twrobinson:
Do not merge.


twrobinson:
jupyterlab-plotly extension does not want to install when explicit version is requested

twrobinson:
Might have been a temporary glitch with the npm repository as it works now.

PR #1972, opened 2020-09-28
[daint] LAMMPS 20.08 with pat

twrobinson:
Adds the perftools-lite versions of the production LAMMPS software.


twrobinson:
This is ready to be merged, as the local build works. See https://github.com/eth-cscs/production/pull/1536#issuecomment-603230178 for a previous reference.

lucamar (approved):
Thanks @twrobinson, let's merge it!

PR #1934, opened 2020-09-14
[daint] Update HPX to 1.5.0

msimberg:

PR #1973, opened 2020-09-29
[daint,dom] Add HPX 1.5.0 to production lists

msimberg:
Forgot this in #1934.


victorusu (approved):
`lgtm`

lucamar:
Thanks @msimberg: in general we announce a change of default (e.g. on the Latest News of User Portal or by e-mail if you have a users mailing list). The old release will not be removed automatically, would you like to remove it?

msimberg:
> Thanks @msimberg: in general we announce a change of default (e.g. on the Latest News of User Portal or by e-mail if you have a users mailing list). The old release will not be removed automatically, would you like to remove it?

Ok, I don't mind. I think the old release can stay. I changed the default since I think we have have few enough users. However, might be safer to keep it at 1.4.0 just in case someone is actually relying on the default being 1.4.0. I'll change back the default.

PR #1974, opened 2020-09-30
[dom] Update Cray metadata for CDT 20.09

lucamar:
I provide an updated version of the Cray metadata file with CUDA 11.0 and GCC 9.3.0 supported by Nvidia: please let me know if the versions of Cray modulefiles still need to be hardcoded or if we can use the defaults.

### DIFF
--- a/easybuild/cray_external_modules_metadata-20.09.cfg
+++ b/easybuild/cray_external_modules_metadata-20.09.cfg
@@ -6,23 +6,18 @@ name = FFTW
 
 [cray-hdf5]
 name = HDF5
-prefix = HDF5_DIR
 
 [cray-hdf5-parallel]
 name = HDF5
-prefix = HDF5_DIR
 
 [cray-libsci]
 name = LibSci
-prefix = CRAY_LIBSCI_PREFIX
 
 [cray-netcdf]
 name = netCDF, netCDF-Fortran
-prefix = NETCDF_DIR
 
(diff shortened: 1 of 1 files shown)


lucamar:
I have updated the GCC version in the Cray metadata file to `9.3.0` as this is supported by CUDA 11.0:
```
/usr/local/cuda-11.0/bin/../targets/x86_64-linux/include/crt/host_config.h:138:2: error: #error -- unsupported GNU version! gcc versions later than 9 are not supported!
  138 | #error -- unsupported GNU version! gcc versions later than 9 are not supported!
      |  ^~~~~
```

ekouts:
Thanks @lucamar, the version and prefix variables are missing in modules `cray-petsc-complex`, `cray-petsc-complex-64` and `cudatoolkit`. We can use the defaults in the rest.

lucamar:
Thanks @ekouts, I have removed all the versions, except for the modules that you have mentioned.

ekouts (approved):

ekouts:
@lucamar , I think we don't need the prefix variable in modules `cray-hdf5` etc? Because if the module has both the variables <name>_VERSION and <name>_PREFIX it should ignore the metadata

lucamar:
Thanks @ekouts, I have removed the `PREFIX` variables from the metadata, when `_VERSION` and `_PREFIX` were both defined in the external module.

PR #1987, opened 2020-10-06
[daint,dom] ipcmagic is back

twrobinson:
Adding ipcmagic and dependencies back into the build. It was too messy with a separate module, and it led to a bug where the use of ipython at the command line failed with: ModuleNotFoundError: No module named 'ipyparallel'. The reason for this was the startup file /apps/daint/UES/jenkins/7.0.UP02/gpu/easybuild/software/IPython/7.7.0-CrayGNU-20.08-python3/startup/01-ipcmagic-cscs.py. So, previously, for this not to fail you needed to load the ipcmagic module in your .jupyterhub.env.
This should fix a bug needed for JF's course next week.


twrobinson:
retest this please

rsarm (approved):

PR #1988, opened 2020-10-06
[daint-mc] LAMMPS with user-mofff

lucamar:
I'm providing an updated recipe for LAMMPS on the multicore nodes, adding the package `user-mofff`: I have tested the build successfully with the ReFrame checks (small and large production), therefore I don't see conflicts with the current packages provided.


twrobinson (approved):
Looks good to me, thanks

PR #2011, opened 2020-11-06
[eiger] Updated CDO and metadata with production list for Eiger

lucamar:
I had to insert `prefix` and `version` in the metadata file for the `CDO` dependency `cray-netcdf`, otherwise I was getting an error from the EasyBuild function `get_software_root`.

### DIFF
--- a/easybuild/cray_external_modules_metadata-20.10.cfg
+++ b/easybuild/cray_external_modules_metadata-20.10.cfg
@@ -15,6 +15,8 @@ name = LibSci
 
 [cray-netcdf]
 name = netCDF, netCDF-Fortran
+prefix = CRAY_NETCDF_DIR
+version = 4.7.4.2, 4.7.4.2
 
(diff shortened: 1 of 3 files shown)


easybuild/cray_external_modules_metadata-20.10.cfg
```diff
@@ -15,6 +15,8 @@ name = LibSci
 
 [cray-netcdf]
 name = netCDF, netCDF-Fortran
+prefix = CRAY_NETCDF_DIR
+version = 4.7.4.2, 4.7.4.2
```
jgphpc, line 19: why `4.7.4.2` twice ?

lucamar: Because you have two match the two packages listed in the field `name`

jgphpc (approved):

ekouts (approved):

PR #2012, opened 2020-11-06
[eiger] Last stable release of LAMMPS

lucamar:
I had to add include and library flags for `cray-python`, since the Cray wrappers do not seem to add them to the compile and link lines, when the corresponding module is loaded.

I also had to remove some packages that are no longer available: see https://lammps.sandia.gov/doc/Commands_removed.html


twrobinson (approved):
Great, thanks Luca

PR #2027, opened 2020-11-18
[eiger] QuantumESPRESSO

lucamar:
Recipe of the latest release of QuantumESPRESSO with dependencies.


easybuild/easyconfigs/l/libxc/libxc-5.0.0-cpeGNU-20.10.eb
```diff
@@ -0,0 +1,36 @@
...
+source_urls = ['https://www.tddft.org/programs/%(name)s/down.php?file=%(version)s']
+sources = [SOURCE_TAR_GZ]
+
+builddependencies = [
+    ('bzip2', '1.0.6'),
+    ('CMake', '3.18.4', '', True),
+]
+
+configopts = [
+    " -DENABLE_FORTRAN=ON -DCMAKE_INSTALL_LIBDIR=lib ",
```
teojgo, line 23: Are both these lines needed?

teojgo (changes requested):

easybuild/easyconfigs/l/libxc/libxc-5.0.0-cpeGNU-20.10.eb
lucamar: Yes, this is a multi-iteration build: the first iteration build the static libraries, while the second line (and second iteration) builds the shared ones.

teojgo (approved):

PR #2028, opened 2020-11-19
[eiger] Update 1.3.1-20.10-eiger with Jupyter

twrobinson:
Adds JupyterLab (not working yet) to force build of nodejs, needed for testing purposes


twrobinson (commented):
This is not a working version of JupyterLab, but is rather to trigger install of nodejs

PR #2031, opened 2020-11-19
[eiger] Julia, JuliaExtensions and JupyterLab

twrobinson:
JupyterLab 2.2.8. The recipe includes Julia kernel but hides the kernel (Julia is not working due to ENOTSUP). The associated Julia and JuliaExtenstion builds should be hidden in the production lists (or not included at all in the production lists if that would also work)


twrobinson:
retest this please

PR #2041, opened 2020-11-25
[dom-gpu] Minor changes in UpdateEB

lucamar:
I have changed `BUILDPATH` to the default one under `$XDG_RUNTIME_DIR` since we use the login node and `/tmp` is not reliable: see my comments in some issues of https://jira.cscs.ch/browse/UES-1240. I have also fixed `TMPDIR` general access.


victorusu (approved):
`lgtm`

PR #2044, opened 2020-11-25
[eiger] Create production list 1.3.2-20.10-eiger

twrobinson:
add "Default" tag for jupyterlab module


lucamar:
Thanks @twrobinson: I think we should create a new production list `1.3.2-20.10-eiger` to match the recent upgrade of the operating system.

lucamar (approved):

PR #2048, opened 2020-11-27
[eiger] Workaround for Lmod on Eiger

lucamar:
The two recipes are a workaround for Lmod on Eiger after the upgrade.


twrobinson (approved):
Thanks Luca

jgphpc (commented):
:+1:

PR #2049, opened 2020-11-29
[eiger] jupyterlab-2.2.8-cpeGNU-20.10.eb

twrobinson:
JupyterLab 2.2.8 without Julia kernel. We will reintroduce Julia kernel when we have a working Julia and JuliaExtensions (currently suffering from ENOTSUP).
Ready for merging.


lucamar (approved):
Thanks @twrobinson, I have also removed Julia and JuliaExtensions from the production list of Eiger

PR #2058, opened 2020-12-09
[daint-gpu] Add git-lfs 2.12.1

victorusu:


jgphpc (approved):
can not remember why keeping the install script but lgtm

PR #2069, opened 2020-12-14
[daint] Updated NAMD modules added to production lists

ChristopherBignamini:


lucamar (approved):
Thanks @ChristopherBignamini

PR #2083, opened 2020-12-22
[dom-gpu] Remove Amber from dom-gpu

lucamar:
Amber GPU fails to build on Dom with CUDA 11 with the error:
```
Error: Unsupported CUDA version 11.0 detected
```


victorusu (approved):
`lgtm`

PR #2092, opened 2020-12-23
[dom-gpu] Update tensorflow and horovod

rsarm:


rsarm:
retest this please

henrique (approved):
lgtm, thanks @rsarm !
(as discussed, we should add it to the prod list after adding tensorboard support)

PR #2094, opened 2021-01-05
[daint dom eiger tsa] Update system_wide.cfg

victorusu:
EasyBuild has updated its list of dependencies for installing the `foss` toolchain. This PR hides most of the "unnecessary" ones
```
UCX
GMP
OpenMPI
PMIx
libpciaccess
Perl
DB
groff
makeinfo
libfabric
UnZip
```

This has the potential to affect the non-Cray systems we have.
For some reason that I am not sure, we do not hide GMP on Daint.
So, this PR also affects Daint.


kraushm:
On the machines with FOSS (at least the ones I'm aware of) the recommended workflow is through meta-modules for the tool chains instead of loading the single modules like `OpenMPI`. Hence, I would assume these should not be affected immediately.

Does this PR require the software stack to be redeployed anew or is this optional?

victorusu:
It is optional

lucamar:
retest this please

lucamar:
retest this please

lucamar (approved):
Please @kraushm @gppezzi let me know if we can merge this pull request

kraushm (approved):
lgtm.

PR #2102, opened 2021-01-11
[dom] updated GROMACS to version 2020.4 to compile with CUDA 11

sebkelle1:


lucamar (approved):
Thanks @sebkelle1, I have also updated the production lists in the pull request

PR #2105, opened 2021-01-13
[dom-gpu] QuantumESPRESSO 6.6a2 with CUDA 11

lucamar:
Recipe of QuantumESPRESSO 6.6a2 adapted for CUDA 11 with Cray PE 20.10 on Dom.

PR #2106, opened 2021-01-14
[dom-gpu] QuantumESPRESSO 6.6a2 with CrayPGI 20.12 and CUDA 11

lucamar:
I'm adding to the recipe the hard coded HDF5 library path, to circumvent the issue with `HDF5_ROOT` that is not expanded correctly in some cases and leads to a wrong HDF5 library path with `PrgEnv-PGI` (see my last comment in #2105).


teojgo (approved):

lucamar:
I have added the following lines to the EasyBuild metadata file for Cray modules:
```
prefix = CRAY_HDF5_PREFIX
version = 1.12.0.0
```
Without explicitly stating the `prefix`, the variable `EBROOTHDF5` was defined as `/opt/cray/pe/hdf5/1.12.0.0/CRAYCLANG/9.0` instead of as `/opt/cray/pe/hdf5/1.12.0.0/PGI/20.1`, although the toolchain of the recipe is `CrayPGI`. 
In this pull request I have addressed only the `cray-hdf5` modulefile in the metadata file, but I think that we might need to address also other external modules: what is your opinion @ekouts?

ekouts (approved):

PR #2110, opened 2021-01-15
[dom] Switch to cdt-cuda in UpdateEB

lucamar:
Replace `cdt` with `cdt-cuda` in `UpdateEB` to build the software stack on Dom.


ChristopherBignamini (approved):
lgtm

PR #2116, opened 2021-01-19
[dom-gpu] Fix ELPA cuda download

lucamar:
Download from ELPA web server is often unreliable, use local downloaded file instead (see https://jira.cscs.ch/browse/UES-1364).


victorusu (approved):
`lgtm`

PR #2117, opened 2021-01-19
[dom-mc] Fix ELPA download (multicore)

lucamar:
Download from ELPA web server is often unreliable, use local downloaded file instead (see https://jira.cscs.ch/browse/UES-1365).


jjotero (approved):
lgtm

PR #2118, opened 2021-01-19
[dom] Fix UDUNITS 2.2.26 download

lucamar:
The tarball of version 2.2.26 is no longer available for download from the FTP server: see https://jira.cscs.ch/browse/UES-13[66-69].


jgphpc (approved):
2.2.26 was released in 2017-12-22,
it may be time to move to 2.2.28 (released in 2020-12-28) ? 
https://www.unidata.ucar.edu/software/udunits/CHANGE_LOG

PR #2120, opened 2021-01-19
[daint] Add CP2K 8.1 in production

lucamar:
I have added the recipes of the latest CP2K stable release 8.1 in the production lists on Piz Daint.


ChristopherBignamini (approved):
lgtm

PR #2127, opened 2021-02-01
[daint-gpu] Lammps easyconfig with plumed support created

ChristopherBignamini:


easybuild/easyconfigs/l/LAMMPS/LAMMPS-03Mar20-CrayGNU-20.08-PLUMED-cuda.eb
```diff
@@ -0,0 +1,46 @@
+# Contributed by TWR (CSCS)
+easyblock = 'MakeCp'
+
+name = 'LAMMPS'
+version = '03Mar20'
+local_plumedversion = '2.5.1'
+versionsuffix = '-PLUMED-cuda'
```
lucamar, line 7:
```suggestion
versionsuffix = '-PLUMED-%s-cuda' % local_plumedversion
```

lucamar: I would keep the version of PLUMED in the `versionsuffix`, since you load a specific version in the dependencies and not just the default like `cudatoolkit`.

lucamar (approved):
Thanks @ChristopherBignamini, I have removed `dom` from the subject, since the default Cray PE is 20.11 on that system. Please see my suggestion about `local_plumedversion`.

twrobinson (approved):

PR #2128, opened 2021-02-01
[daint] Add Julia 1.5.3 and improve Julia admin path handling

omlins:
## The problem
Building and testing of certain packages from within the Julia Pkg manager can fail due to the Julia startup file. Either because 1) it gives an error if the DEPOT_PATH/LOAD_PATH were modified and the length does not correspond to length(USER_DEPOT_PATH) + length(ADMIN_DEPOT_PATH) or length(USER_LOAD_PATH) + length(ADMIN_LOAD_PATH)  or 2) the startup file is deactivated (according to the documentation, building happens without startup file by default except if julia is called explicitly with --startup-file=yes).

## Solution
1. The modules leave the DEPOT_PATH and LOAD_PATH essentially as defined by default by the Julia installation (introducing though mc and daint as well as using #.#.#).
2. The modules define the ADMIN_DEPOT_PATH and ADMIN_LOAD_PATH environment variables as before, with the exception that the standard Julia  installation depot paths  (/apps/daint/UES/omlins/julia-1.5.3/software/Julia/1.5.3-CrayGNU-20.08/local/share/julia and /apps/daint/UES/omlins/julia-1.5.3/software/Julia/1.5.3-CrayGNU-20.08/share/julia) and load path (@stdlib) are not added (as already in the DEPOT_PATH  and LOAD_PATH)
3. USER_DEPOT_PATH and USER_LOAD_PATH are not defined or just not used (and kept for possible other usage).
4. STD_DEPOT_PATH and STD_LOAD_PATH environment variables are additionally defined.
5. The startup file injects the entries of ADMIN_DEPOT_PATH  and ADMIN_LOAD_PATH to the DEPOT_PATH and LOAD_PATH after any user path, but before the standard Julia installation depot and load paths.
6. The startup file does nothing in certain conditions to ensure proper and consisting working in any case (Pkg build, test ...):  if paths empty (or only "@" for LOAD_PATH) or all entries in std path.

## Result
1. Standard usage of Julia is as before. 
2. For specific operations which are run with --startup-file=no  (implicitly or explicitly) the admin packages are simply deactivated. 
3. For specific operations which are run with --startup-file=yes  (implicitly or explicitly) the admin packages are also deactivated if the  conditions listed in point 6 above apply. Else, the admin paths are activated, but there is no constraint on the DEPOT_PATH/LOAD_PATH defined by the user or Pkg manager.
4. In addition, each user has the choice to easily completely deactivate the usage of admin packages  by creating a julia alias with --startup-file=no.

PR #2133, opened 2021-02-03
[dom] Update jupyterlab-1.2.16-CrayGNU-20.11-batchspawner-cuda.eb

twrobinson:
JupyterHub 1.3 (and deps) to align with update of JupyterHub on VM.


omlins (changes requested):
This should include the changes from #2128

omlins:
The changes from master and in the PE 20.08 recipe are now done here too.

omlins (approved):

PR #2137, opened 2021-02-04
[dom-mc] CP2K 8.1 with PE 20.11 (multicore)

lucamar:
CP2K latest stable release with the corresponding architecture file and ELPA dependency updated (multicore).


kraushm (approved):
lgtm.

PR #2142, opened 2021-02-09
[daint-gpu] Fix a typo in the intel and pgi compilers version

victorusu:


spack/daint/compilers.yaml
```diff
@@ -63,6 +63,8 @@ compilers:
     modules:
     - PrgEnv-intel
     - intel/19.0.1.144
+    - cdt/20.08
+    - intel/19.0.1.144
```
ekouts, line 67: You have added this twice

spack/daint/compilers.yaml
```diff
@@ -78,6 +80,8 @@ compilers:
     modules:
     - PrgEnv-intel
     - intel/19.1.1.217
+    - cdt/20.08
+    - intel/19.1.1.217
```
ekouts, line 84: same here

spack/daint/compilers.yaml
```diff
@@ -93,6 +97,8 @@ compilers:
     modules:
     - PrgEnv-pgi
     - pgi/20.1.0
+    - cdt/20.08
+    - pgi/20.1.0
```
ekouts, line 101: same

spack/daint/compilers.yaml
```diff
@@ -108,6 +114,8 @@ compilers:
     modules:
     - PrgEnv-pgi
     - pgi/20.1.1
+    - cdt/20.08
+    - pgi/20.1.1
```
ekouts, line 118: same

ekouts (changes requested):

spack/daint/compilers.yaml
victorusu: Yes! Due to Spack's limitation, you cannot have the `cdt` defined before the first `intel` entry. And whenever we load the `cdt` module the compiler is reset. So we need to load again the `intel` module.

ekouts (approved):

PR #2144, opened 2021-02-11
[daint dom eiger tsa] Add createJiraSD to util.groovy

lucamar:
Function to create a Jira SD ticket (based on `create_jsd_issue` created by @GiuseppLoRe at https://git.cscs.ch/misc/storage_scripts/-/blob/master/cscs_tools/cscs_tools.py).

### DIFF
--- a/jenkins/util.groovy
+++ b/jenkins/util.groovy
@@ -99,3 +99,22 @@ void failedJiraTask(String projkey, String recipe, String machine) {
                           issuetype: [name: 'Task']]]
    def newIssue = jiraNewIssue issue: issue, site: 'JIRA_SITE'
 }
+
+/**
+* Create a Jira Service Desk ticket with custom message
+*
+* @param projkey Jira project key (only project members can create issues).
+* @param recipe EasyBuild recipe that failed to build.
+* @param machine Computing system where the recipe failed to build.
+*/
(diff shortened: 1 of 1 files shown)


jenkins/util.groovy
```diff
@@ -99,3 +99,23 @@ void failedJiraTask(String projkey, String recipe, String machine) {
...
+* Create a Jira Service Desk ticket with custom message
+*
+* @param title Subject of the Jira Service Desk ticket appended to [${machine}].
+* @param machine Computing system where the build took place.
+* @param message Content of the Jira Service Desk ticket prepended to Jenkins job details.
+* @param queue Queue where the Jira Service Desk ticket will be dispatched.
+*/
+void createJiraSD(String title, String machine, String message, String queue) {
+
+   def title = "[${machine}] ${title}"
```
teojgo, line 113: Remove this and put it directly in the `summary` below because there is already a `title` defined.

teojgo (changes requested):

jenkins/util.groovy
lucamar: Thanks @teojgo, I've changed the parameter name to `subject`, in order to avoid conflicts

teojgo (approved):

PR #2152, opened 2021-02-16
[dom] Reset CP2K 7.1 as default on Dom

lucamar:
The change is due to a performance regression with CP2K 8.1: see https://jira.cscs.ch/browse/UES-1424 for more details.


ChristopherBignamini (approved):
lgtm

PR #2153, opened 2021-02-16
[dom] Remove PLUMED Intel

lucamar:
I remove PLUMED built with the Intel compiler, due to recurring issues when building from scratch the dependencies and the code (see https://jira.cscs.ch/browse/UES-1425). We have set PLUMED built with the GNU compiler as default, so we don't need PLUMED with Intel in production.


victorusu (approved):
`lgtm`

PR #2155, opened 2021-02-19
[eiger] Configure Pilatus

lucamar:
I provide the configuration of Pilatus login node (`uan04`) using EasyBuild HierarchicalMNS with Lmod: therefore the `prefixProduction` parameter in `Machines.groovy` is not using the Cray PE version 21.02, since users will be able to switch PE by swapping the corresponding toolchain modulefiles. For the same reason, the `cpe` toolchains are no longer hidden.  
 
The metadata file defines `PREFIX` for some important external modules, to avoid the issue that we have experienced with EasyBuild setting the corresponding `EBROOT` variable before switching the programming environment.


jenkins/Machines.groovy
```diff
@@ -22,6 +22,14 @@ def eiger = [name: 'eiger',
            modulesProduction: '',
            prefixProduction: '/apps/eiger/UES/jenkins/1.3.2/20.10']
 
+def pilatus = [name: 'pilatus',
```
teojgo, line 25: Just format the dictionary so that the keys fall in the same column.

teojgo (changes requested):

jenkins/Machines.groovy
lucamar: Thanks @teojgo, does may last commit address your remark?

teojgo (approved):

PR #2156, opened 2021-02-19
[pilatus] Toolchain recipes and GSL in production list

lucamar:
I provide the recipes of the toolchains for Cray PE 21.02 on Pilatus and the easyconfig files of GSL.


teojgo (approved):
lgtm

PR #2157, opened 2021-02-19
[pilatus] Draft recipe for Amber on Pilatus

lucamar:
I provide a draft easyconfig file for Amber with the toolchain `cpeIntel/21.02` on Pilatus: 
- I have replaced `module load gcc/8.3.0` from `prebuildopts` with `module load gcc 9.3.0`, since `8.3.0` is not available on the system. We might also try loading the default `10.1.0`, if GCC 10 won't be too strict for the build
- We could probably remove the patch `amber20-nvcc.patch`, since we don't have a CUDA build: what do you think, @victorusu?


lucamar:
The error is reported in the log file `/scratch/e1000/jenscscs/TestingEB/pilatus/278b2bb-4639/tmp/eb-8KRDJw/easybuild-Amber-20-11-7-20210219.141018.XuyMY.log`:
```
Obtaining the intel compiler suite versions, e.g.:
     cc -v
Error: cc is not well formed or produces unusual version details!
       Check for a CC environment variable.
```
However, when I load `cpeIntel` on the system, I get a reasonable output for `cc -v`:
```
$ ml cpeIntel/21.02

The following have been reloaded with a version change:
  1) cray-mpich/8.0.16 => cray-mpich/8.1.2     2) craype/2.7.2 => craype/2.7.5     3) perftools-base/20.10.0 => perftools-base/21.02.0

$ cc -v
icc version 19.1.3.304 (gcc version 7.5.0 compatibility)
```

lucamar:
I'm closing this pull request since #2194 has been submitted with a newer release of Amber, removing `cray-netcdf` from the dependencies.

PR #2163, opened 2021-02-23
[dom daint eiger pilatus tsa] Add recipe for ReFrame version 3.4.2

jenkins-cscs:


easybuild/easyconfigs/r/reframe/reframe-3.4.2.eb
```diff
@@ -0,0 +1,48 @@
...
+]
+
+keepsymlinks = True
+postinstallcmds = [
+    'cp -r %(builddir)s/reframe-tests-%(version)s/checks %(installdir)s/cscs-checks/private',
+    './bootstrap.sh +docs +pygelf && '
+    'cp -r external %(installdir)s/external && '
+    'mkdir -p %(installdir)s/share/man/man1 %(installdir)s/share/man/man8 && '
+    'cp docs/man/man1/reframe.1 %(installdir)s/share/man/man1 && '
+    'cp docs/man/man8/reframe.settings.8 %(installdir)s/share/man/man8'
```
vkarak, line 31: Do we need to copy now also the shell competions? Check here: https://reframe-hpc.readthedocs.io/en/stable/started.html#enabling-auto-completion

vkarak (changes requested):

easybuild/easyconfigs/r/reframe/reframe-3.4.2.eb
vkarak: No, they will not work, but they should be there so that users can source them according to the documentation.

teojgo: Ok, I will add the corresponding command and push

teojgo:
Ok to test

vkarak (approved):
lgtm

PR #2164, opened 2021-02-23
[eiger] Fix Paraview 5.9.0 recipe for eiger

teojgo:


jfavre (approved):
Am not sure it can build with maxparallel=64. In the past, there has been issues. Curious to see.

lucamar (approved):
Thanks @jfavre and @teojgo

PR #2165, opened 2021-02-23
[eiger] Set GALLIUM_DRIVER to llvmpipe

teojgo:


jfavre (approved):
we tested together and it is right. Fixing also the client-server connection, and making it explicit that the default value is indeed "llvmpipe"

PR #2169, opened 2021-02-24
[pilatus] Jupyterlab and dependencies

lucamar:
I have updated the dependencies using the latest stable release available: I did not update the versions of the Python extensions and the version of JupyterLab itself, though.


twrobinson (approved):
Tested by launching on pilatus and tunneling the connection to my browser. Works fine. Thanks @lucamar

PR #2172, opened 2021-02-25
[pilatus] QuantumESPRESSO 6.7.0

lucamar:
I provide the recipe of QuantumESPRESSO 6.7.0 with the toolchain `cpeGNU/21.02`: the more recent version of the `libxc` library is added for future use, since it could not be used immediately as a dependency.
The `prefix` and `version` of some key modules provided by the Cray Programming Environment have been added to the metadata file, in order to get the correct library path of the corresponding library with the `EBROOT` variables.


kraushm (approved):
lgtm. Thanks @lucamar

PR #2173, opened 2021-02-25
[pilatus] Update production list 1.3.2-21.02-pilatus

lucamar:
I have added `jupyter-utils` and `juypterlab` on Pilatus: the first one is built with the system toolchain, therefore the corresponding modulefile will be installed under the `Core` folder. 
The `jupyterlab` modulefile will be installed under to corresponding toolchain folder instead and it will be available for loading only after the toolchain modulefile `cpeGNU/21.02` is loaded (Lmod hierarchy).


twrobinson (approved):
Great, thanks Luca

PR #2177, opened 2021-02-25
[pilatus] NAMD (no smp) build with PE 21.02 

ChristopherBignamini:


lucamar (approved):
Thanks @ChristopherBignamini

PR #2178, opened 2021-02-26
[eiger] LAMMPS python fix

twrobinson:
Fix for python. Previous recipe included a workaround for missing python, and although it built it did not execute.


lucamar (approved):
Thanks @twrobinson

PR #2181, opened 2021-03-01
[pilatus] ecCodes on Pilatus

lucamar:
I provide the recipe of the latest release of ecCodes and the dependencies with PE 21.02 on Pilatus. I'm not adding the module explicitly to the production list, since it will be a dependency of the CDO installation.


kraushm (approved):
lgtm. Thanks @lucamar

PR #2182, opened 2021-03-01
[pilatus] PROJ 8 and dependencies

lucamar:
I provide the recipe and dependencies of `PROJ` (latest stable release 8.0.0) with Cray PE 21.02 on Pilatus: I'm not adding the module directly in the production list, since it will be a dependency of the new module `CDO` (to be submitted soon).


kraushm (approved):
lgtm. Thanks @lucamar

PR #2184, opened 2021-03-02
[daint-gpu] update VMD to the latest 1.9.4

jfavre:
== COMPLETED: Installation ended successfully (took 12 sec)
== Results of the build can be found in the log file(s) /scratch/snx3000/jfavre/daint/software/VMD/1.9.4a51/easybuild/easybuild-VMD-1.9.4-20210302.150958.log

### DIFF
--- a/easybuild/easyconfigs/v/VMD/VMD-1.9.4.eb
+++ b/easybuild/easyconfigs/v/VMD/VMD-1.9.4.eb
@@ -1,0 +1,44 @@
+# @author: jfavre
+
+easyblock = "CmdCp"
+
+name = 'VMD'
+version = '1.9.4'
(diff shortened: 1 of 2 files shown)


lucamar:
Thanks @jfavre: are you sure to set the latest VMD as default? Would you like to add the same on Dom?

jfavre:
> Thanks @jfavre: are you sure to set the latest VMD as default? Would you like to add the same on Dom?

Yes, there was no previous default. And v1.9.3 is from november 2016, so I feel strongly that it is time to move up.

lucamar (approved):
Thanks @jfavre: I would add the same version on Dom as well, to keep both systems in sync.

PR #2185, opened 2021-03-02
[dom-gpu] Add VMD to 7.0.UP02-20.11-dom-gpu

jfavre:
add VMD-1.9.4 to dom, as well as suggested by @lucamar


lucamar (approved):
Thanks @jfavre, I have added `VMD` to the `20.11` production list, that is the current target of the symbolic link `dom-gpu`.

PR #2187, opened 2021-03-04
[pilatus] LAMMPS with PE 21.02

lucamar:
I have updated the recipe of LAMMPS using CMake: I build `lmp_mpi` only, which I have tested successfully on Pilatus. 

The packages selected with the preset `nolib.cmake` match previous builds, except for the packages `COMPRESS`, `MPIIO` and `PYTHON`, that are added with the CMake flag `-DPKG_<name>=ON` (see https://lammps.sandia.gov/doc/Build_package.html#cmake-presets).


twrobinson (approved):
Very nice. Thanks @lucamar

PR #2209, opened 2021-03-11
[dom-gpu] force autoload of ParaView NVIDIA Index lib plugin

jfavre:
work-around for https://gitlab.kitware.com/paraview/paraview/-/issues/20487


lucamar (approved):
Thanks @jfavre: if you want to rebuild the module on Dom, we need to force it on Jenkins

jfavre:
> Thanks @jfavre: if you want to rebuild the module on Dom, we need to force it on Jenkins

Yes indeed, we need to force a rebuild. Can you do that please?

PR #2211, opened 2021-03-11
[dom-mc] CP2K 7.1 and 8.1 with CrayPAT 20.11

lucamar:
The default module in production will still be CP2K 7.1 due to a performance issue with the gpu release of CP2K 8.1: for the sake consistency, I keep the same default also with the mc release. I provide a minor fix for production recipe of CP2K 7.1 as well.


henrique (approved):
lgtm
thanks @lucamar

PR #2222, opened 2021-03-16
[dom] Update jupyterlab-2.0.2-CrayGNU-20.11-batchspawner-cuda.eb

twrobinson:
Adds a single requirement for EMPA course. This was added to 20.08 after the freeze so was not caught by the auto-update to 20.11. Thus needs to be deployed by hand.


twrobinson:
retest this please

victorusu (approved):
`lgtm`

PR #2232, opened 2021-03-23
[daint-mc] Materials were missing from install dir

jfavre:
== COMPLETED: Installation ended successfully (took 25 min 35 sec)
== Results of the build can be found in the log file(s) /scratch/snx3000/jfavre/daint/software/ParaView/5.9.0-CrayGNU-20.11-OSMesa-python3/easybuild/easybuild-ParaView-5.9.0-20210323.203118.log


lucamar (approved):
Thanks @jfavre: we will need to force the build of the recipe to make the change effective on the system after the request is merged.

PR #2246, opened 2021-04-08
[daint] Update Boost recipes to 20.11 toolchains

msimberg:
Also add them to production lists.

I haven't tested the CrayCCE recipe (I'm assuming the two others will work just fine). Hoping to have CI tell me if I've done something wrong.

### DIFF
--- a/easybuild/easyconfigs/b/Boost/Boost-1.75.0-CrayCCE-20.11.eb
+++ b/easybuild/easyconfigs/b/Boost/Boost-1.75.0-CrayCCE-20.11.eb
@@ -1,0 +1,29 @@
+# contributed by Luca Marsella (CSCS)
+name = 'Boost'
+version = "1.75.0"
+
+homepage = 'http://www.boost.org/'
+description = """
+    Boost provides free peer-reviewed portable C++ source libraries.
+"""
+
+toolchain = {'name': 'CrayCCE', 'version': '20.11'}
+toolchainopts = {'usempi': True, 'pic': True, 'verbose': False}
+
+source_urls = [SOURCEFORGE_SOURCE]
+sources = ['%(namelower)s_1_75_0.tar.gz']
+
+dependencies = [
+    ('bzip2', '1.0.8'),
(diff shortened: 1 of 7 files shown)


aurianer (approved):
LGTM thanks!

msimberg:
Does someone know how the boost recipe actually does the configuration before building (boost uses b2 but I see no mention of it in the recipe)? Is there some other place I can place the `--with-toolset=clang` option for the CrayCCE build?

teojgo:
> Does someone know how the boost recipe actually does the configuration before building (boost uses b2 but I see no mention of it in the recipe)? Is there some other place I can place the `--with-toolset=clang` option for the CrayCCE build?

Can you try `toolset = 'clang'`? It is one parameter accepted by the EasyBlock for Boost.

msimberg:
> > Does someone know how the boost recipe actually does the configuration before building (boost uses b2 but I see no mention of it in the recipe)? Is there some other place I can place the `--with-toolset=clang` option for the CrayCCE build?
> 
> Can you try `toolset = 'clang'`? It is one parameter accepted by the EasyBlock for Boost.

Thanks @teojgo, I didn't know about that option. It seems to have worked nicely!

I tend to forget to add the recipes to the production lists, but this time I remembered. Do they look ok? bzip and zlib don't need to be added, right?

teojgo:
> > > Does someone know how the boost recipe actually does the configuration before building (boost uses b2 but I see no mention of it in the recipe)? Is there some other place I can place the `--with-toolset=clang` option for the CrayCCE build?
> > 
> > 
> > Can you try `toolset = 'clang'`? It is one parameter accepted by the EasyBlock for Boost.
> 
> Thanks @teojgo, I didn't know about that option. It seems to have worked nicely!
> 
> I tend to forget to add the recipes to the production lists, but this time I remembered. Do they look ok? bzip and zlib don't need to be added, right?

Yeap they seem fine now. You don't need to include bzip and zlib in the production lists. They are going to be built as hidden modules since Boost depends on them.

PR #2252, opened 2021-04-13
WIP EasyBuild recipe for Cosmo Pompa 24d1ac6 + dependencies cdt/20.08

teojgo:


easybuild/easyblocks/boostcray.py
```diff
@@ -0,0 +1,222 @@
+##
```
victorusu, line 1: @teojgo, I do not understand the existence of this file...
Why is it here? None of this recipes use boost... There is no boost recipe...
Is it just to make it Python3 compatible?

teojgo: Yeah I need to add boost as well for 20.08, version 1.67

teojgo: @victorusu Boost is used by SerialBox

omlins: @victorusu , boostcray easyblock is needed, else COSMO does not build - tried several times to remove it, but it is required.

victorusu:
@teojgo and @omlins, are we still deploying recipes using 20.08? 

omlins:
As it is Theo's PR, he will know if he still needs that for some reason.

PR #2253, opened 2021-04-14
[daint-gpu] VASP 6.2.0 OpenACC

lucamar:
I have adapted the instructions to build the [OpenACC port](https://www.vasp.at/wiki/index.php/OpenACC_GPU_port_of_VASP) of VASP 6.2.0 on Piz Daint and I provide the recipes of the dependencies as well.


lucamar:
I have removed `[dom-gpu]` since the `cudatoolkit` modulefile of CUDA 10.2 has a different revision string on that system.

victorusu (approved):
`lgtm`

PR #2260, opened 2021-04-19
[dom-gpu daint-gpu] Use ConfigureMake easyblock for Magma

teojgo:
Changing to configure make since the pckgconfig is configured with `make install`


lucamar (approved):
Thanks @teojgo

PR #2270, opened 2021-04-21
[dom] Drop xalt when loading julia, since it hijacks libcurl.so and is incompatible

haampie:
xalt hijacks libcurl.so through LD_LIBRARY_PATH as julia tries to open a vendored version of the library.

This results in warning messages when using Downloads.jl (which is used by the package manager).

Julia has good reasons to vendor libcurl: https://julialang.org/blog/2021/03/julia-1.6-highlights/#downloads_networkingoptions

So we should drop xalt here.

Pinging @lucamar, @omlins, @gppezzi

### DIFF
--- a/easybuild/easyconfigs/j/Julia/Julia-1.6.0-CrayGNU-20.11-cuda.eb
+++ b/easybuild/easyconfigs/j/Julia/Julia-1.6.0-CrayGNU-20.11-cuda.eb
@@ -37,6 +37,13 @@ modextravars = {
     'JULIA_MPI_TEST_ARRAYTYPE': 'CuArray',
 }
 
-modtclfooter = "module unload cray-libsci_acc"
-
 moduleclass = 'lang'
+
+# Julia 1.6.0+ uses Downloads.jl with a vendored libcurl.so that's dlopen'ed.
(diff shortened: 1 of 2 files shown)


haampie:
retest this please

teojgo:
ok to test

lucamar:
Thanks @haampie: if it's just a warning, can we simply ignore it for the moment? 
We can use a recent `libcurl` as well in the list of dependencies, if needed.

haampie:
It's a warning but it looks like an error in all red; it also comes up when doing distributed computing without mpi, cluttering the output, so it's not a great user experience :(.

>  We can use a recent libcurl as well in the list of dependencies, if needed.

It shouldn't be necessary and maybe only asking for problems, since they build libcurl in a very specific way: https://github.com/JuliaPackaging/Yggdrasil/blob/8c858d9b3a13ea285e147d929a3707c8764401bf/L/LibCURL/build_tarballs.jl and distribute cross-compiled binaries themselves: https://github.com/JuliaBinaryWrappers/LibCURL_jll.jl/releases/, as is done for all binary dependencies of julia :upside_down_face:

victorusu (approved):
`lgtm`

PR #2269, opened 2021-04-21
[daint] Update 7.0.UP02-20.11-daint-gpu

twrobinson:
This changes the default Julia to 1.6 which offers a bunch of enhancements, and very little regression.. if any! JupyterLab kernel is already at 1.6, so this will align termianl and jupyter behaviour nicely.

PR #2272, opened 2021-04-21
[daint] Bump default version of Julia to 1.6.0

haampie:
1.6.0 is working just fine, and is a quite a bit faster compared to 1.5.x in terms of package manager, package precompilation, and compiler latency; so it'd be good to default to 1.6.0

Pinging @lucamar, @omlins, @gppezzi


haampie:
Probably #2270 should go first

lucamar:
Thanks @haampie: in general we prevent users before changing default, but if the change has no adverse impact on the workflows with Julia, we can go on. 
I'd suggest to remove the previous version from the production list: the module on the system won't be removed, though, but it makes it cleaner and easier for automated updates.

lucamar:
retest this please

haampie:
Closing because of #2269

PR #2273, opened 2021-04-21
[dom] Fix syntax of JuliaBundle easyblock

teojgo:


lucamar (approved):
Thanks a lot @teojgo, that fixes the conflict!

PR #2274, opened 2021-04-21
[daint dom] Fix stdin syntax error in juliapackage.py

haampie:
Seems like you need to use sys.stdin


lucamar:
retest this please

lucamar (approved):
Thanks for fixing it @haampie

PR #2277, opened 2021-04-21
[dom-gpu] 7.0.UP02-20.11-dom-gpu

twrobinson:
Bugfix. I had added the mc version to the gpu prod list. Sorry about that!


lucamar (approved):
Thanks @twrobinson

PR #2290, opened 2021-04-28
[daint] Add SDL and freeglut

victorusu:


lucamar (approved):
Thanks @victorusu: have you already asked @jfavre?

PR #2291, opened 2021-04-28
[dom daint] Remove JULIA_PKG_SERVER

haampie:
It was added to work around some weird issue where curl would complain about incomplete downloads, but in the end it happened with & without this variable.

Turns out it is useful to unset it, because then the julia package registry gets updated by just a download into ~/.julia/... when you run Pkg.update(), whereas when this variable is set to empty git clone the registry into ~/.julia/..., which curently fails because git tries to mmap something and that fails on the $HOME filesystem.


haampie:
retest this please

twrobinson (approved):
Yep, looks good to merge, thanks

PR #2294, opened 2021-04-28
[eiger] QuantumESPRESSO with cpeIntel

lucamar:
QuantumESPRESSO and dependencies with the Intel compiler to address the issue with `cray-libsci` reported in [SD-52008](https://jira.cscs.ch/browse/SD-52008): `cray-hdf5` is not available with `PrgEnv-intel`, therefore it has been removed with respect to the build with `cpeGNU`.


lucamar:
retest this please

lucamar:
I'm merging since the check passed and we need to build the new module on the system before opening to early users again.

PR #2297, opened 2021-04-30
[eiger] Update jupyterlab-2.2.8-cpeGNU-21.04.eb

twrobinson:
Adding ijulia kernel. If you could check the syntax I've used for the modluafooter and improve if necessary, that would be great (you can compare against the tcl version in e.g. the easyconfig jupyterlab-1.2.16... for daint).


lucamar (approved):
Thanks @twrobinson

PR #2302, opened 2021-05-05
[dom] Update dom production lists with julia 1.6.1

twrobinson:
Upgrade julia to 1.6.1 for testing SD-52054


lucamar:
Thanks @twrobinson: do you want to deploy it only in the gpu production list?

lucamar (approved):
Thanks @twrobinson

PR #2313, opened 2021-05-08
[eiger] scalasca

jgphpc:
trying to pass https://github.com/eth-cscs/production/pull/2311


lucamar (changes requested):
Please @jgphpc add the recipe to the pull request, since it is currently missing:
```
production/easybuild/easyconfigs$ ls s/Scalasca/Scalasca-2.6-cpeGNU-21.04.eb
ls: cannot access 's/Scalasca/Scalasca-2.6-cpeGNU-21.04.eb': No such file or directory
```

PR #2317, opened 2021-05-08
[dom] otf2

jgphpc:


easybuild/easyconfigs/o/otf2_cli_profile/otf2_cli_profile-cd4e452-CrayGNU-21.02.eb
```diff
@@ -0,0 +1,42 @@
...
+toolchain = {'name': 'CrayGNU', 'version': '21.02'}
+toolchainopts = {'dynamic': True, 'verbose': False}
+
+sources = [
+    {'git_config': {'url': 'https://github.com/score-p', 'commit': '%(version)s', 'repo_name': '%(name)s'}, 'filename': SOURCE_TAR_GZ},
+]
+
+# patches = ['%(name)s-%(version)s.patch']
+builddependencies = [
+    ('CMake', '3.20.1', '', True),
```
lucamar, line 19:
```suggestion
    ('CMake', '3.14.5', '', True),
```

lucamar: I would suggest to use the default CMake available on Dom, unless the new release is a requirement for this build

lucamar (changes requested):
Please @jgphpc see my suggested edit

lucamar (approved):
Thanks @jgphpc

PR #2321, opened 2021-05-10
WIP [daint dom] Qt 5.15.2 easyconfig creation

ChristopherBignamini:


easybuild/easyconfigs/q/Qt/Qt-5.15.2-CrayGNU-20.11.eb
```diff
@@ -0,0 +1,26 @@
+# contributed by Luca Marsella (CSCS)
+name = 'Qt'
+version_major_minor = '5.15'
```
lucamar, line 3:
```suggestion
```

lucamar: @ChristopherBignamini you don't need to define `version_major_minor`, it's a template defined by EasyBuild from the `version` parameter

lucamar (commented):
@ChristopherBignamini see my comment about `version_major_minor`

PR #2325, opened 2021-05-12
[daint dom eiger tsa] Fix TMPDIR 

lucamar:
If `XDG_RUNTIME_DIR` is not set, then `TMPDIR` should point to `tmp` as `BUILDPATH`.

### DIFF
--- a/easybuild/module/EasyBuild-custom/cscs
+++ b/easybuild/module/EasyBuild-custom/cscs
@@ -142,16 +142,12 @@ setenv EASYBUILD_INSTALLPATH                    $::env(EASYBUILD_PREFIX)
 #
 if { [ info exists ::env(EASYBUILD_TMPDIR) ] } {
     setenv EASYBUILD_TMPDIR                     $::env(EASYBUILD_TMPDIR)
-} elseif { ! [ info exists ::env(XDG_RUNTIME_DIR) ] } {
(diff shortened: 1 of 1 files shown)


lucamar:
We have already a workaround in place in the modulefile if `XDG_RUNTIME_DIR` is not defined:
```
# set XDG_RUNTIME_DIR if not defined on the current system
if { ! [ info exists ::env(XDG_RUNTIME_DIR) ] } {
    if { [ info exists ::env(SCRATCH) ] } {
        setenv XDG_RUNTIME_DIR                     $::env(SCRATCH)
    } else {
        setenv XDG_RUNTIME_DIR                     /tmp/$::env(USER)
    }
}
```

ChristopherBignamini (approved):
lgtm

kraushm (approved):
lgtm

lucamar:
Actually, I think that the change proposed by this pull request is meaningless: if `XDG_RUNTIME_DIR` is not set, it will be set by the piece of code already present in the modulefile in the following order of priority:
- `$SCRATCH`
- `/tmp/$USER`
The same occurs for `BUILDPATH`, so I propose to remove the additional test for `TMPDIR` and `BUILDPATH`.

kraushm (approved):
lgtm.

PR #2352, opened 2021-06-08
[eiger] Update spack configs for eiger

rasolca:
Version of modules and packages has been updated to match what is available after the maintenance.

### DIFF
--- a/spack/eiger/compilers.yaml
+++ b/spack/eiger/compilers.yaml
@@ -1,41 +1,5 @@
 compilers:
 - compiler:
-    spec: aocc@8.0.0
-    paths:
-      cc: cc
-      cxx: CC
-      f77: ftn
-      fc: ftn
-    flags:
-      cflags: null
-      cxxflags: null
-      fflags: null
-    operating_system: sles15
-    target: any
-    modules:
-    - PrgEnv-aocc
-    - aocc/8.0.0
-    environment: {}
-    extra_rpaths: []
-- compiler:
-    spec: aocc@2.2.0.1
-    paths:
-      cc: cc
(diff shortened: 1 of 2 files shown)


lucamar:
ok to test

spack/eiger/compilers.yaml
```diff
@@ -516,12 +154,12 @@ compilers:
     operating_system: sles15
     target: any
     modules:
-    - PrgEnv-gnu
-    - gcc/9.3.0
+    - PrgEnv-intel
+    - intel/8.0.0
```
haampie, line 158: This doesn't exist, I guess spack doesn't detect the right thing here..

spack/eiger/compilers.yaml
```diff
@@ -1,422 +1,60 @@
...
-    paths:
-      cc: cc
-      cxx: CC
-      f77: ftn
-      fc: ftn
-    spec: pgi@18.5.0
-    target: any
-- compiler:
+    - PrgEnv-aocc
+    - aocc/8.0.0
```
haampie, line 17: Same, doesn't exist...

haampie:
I've pushed some more changes. I can revert the `buildable: false` bits, but in my experience with spack using system libs by default was more often asking for problems than that it would make builds shorter.

rasolca:
Thank you @haampie.
Regarding `buildable: false` I kept them from the daint config.

victorusu (approved):
`lgtm`

victorusu:
@haampie, can we merge? what do you think?

haampie:
Yeah, should be good to go

PR #2357, opened 2021-06-14
[eiger pilatus] Fix EB custom modulefile for Lmod

lucamar:
I have realized that the EasyBuild-custom modulefile does not export correctly `EASYBUILD_PREFIX` for user `jenscscs` on Eiger and Pilatus (Cray with Lmod): the reason could be an issue with the script `tcl2lua.tcl` invoked by Lmod to convert Tcl to Lua syntax.

I have tried to fix the Tcl modulefile enclosed to this pull request removing the following line, which should not be necessary:
```
setenv EASYBUILD_PREFIX                     $::env(EASYBUILD_PREFIX)
```
However, the modified Tcl modulefile does not still work for user `jenscscs` on the systems with Lmod, therefore I provide the corresponding Lua modulefile which will only be loaded on systems with Lmod: the other systems will still load the Tcl modulefile.

### DIFF
--- a/easybuild/module/EasyBuild-custom/cscs
+++ b/easybuild/module/EasyBuild-custom/cscs
@@ -9,12 +9,12 @@ Production EasyBuild @ CSCS
 
 More information
 ================
- - Homepage: https://github.com/eth-cscs/production/wiki/User-instructions-for-EasyBuild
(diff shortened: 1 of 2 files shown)


kraushm (approved):
lgtm.

PR #2360, opened 2021-06-15
[daint-gpu] Add deepmd-kit for craygnu/20.11

rsarm:


twrobinson (approved):
Great, many thanks for this @rsarm

PR #2362, opened 2021-06-15
[daint dom eiger pilatus tsa] Include custom MNS and toolchains

lucamar:
I have implemented the change following the instructions on the [EasyBuild documentation](https://docs.easybuild.io/en/latest/Including_additional_Python_modules.html), since we have a custom module naming scheme and custom toolchains: using the include variables, we can avoid to copy manually the python scripts, therefore I have adapted [our Confluence page](https://confluence.cscs.ch/display/GENERAL/EasyBuild+at+CSCS) accordingly. 

You can check that the new custom modulefiles add the correct python scripts switching to the latest non-default `EasyBuild/4.4.0` (I have removed the custom python scripts from the standard installation folder) with the commands:
```
git clone git@github.com:lucamar/production.git
cd production
git checkout eb-custom
module use $PWD/easybuild/module
module load EasyBuild-custom-cscs
module switch EasyBuild EasyBuild/4.4.0

eb --avail-module-naming-schemes | grep -i lower
	LowercaseModuleNamingScheme

eb --list-toolchains | grep cpe
	cpeAMD: PrgEnv-aocc
	cpeCray: PrgEnv-cray
	cpeGNU: PrgEnv-gnu
	cpeIntel: PrgEnv-intel
```


kraushm (approved):
lgtm.

PR #2366, opened 2021-06-16
[eiger] CP2K with MKL

lucamar:
Update architecture file based on GitHub project.


lucamar:
Usual issue due to know bug in PE 21.04:
```
configure: error: Could not compile an MPI C program
```

PR #2370, opened 2021-06-21
[daint-gpu dom-gpu] Add missing ovito for 20.11

lucamar:
I provide the missing recipe, as discussed with @jfavre.


jfavre (approved):
Thanks @lucamar !

PR #2372, opened 2021-06-23
[eiger pilatus] Fix cscs Lua modulefile to avoid Lmod warning

lucamar:
The change to `cscs.lua` will fix the Lmod warning shown on Eiger and Pilatus for `EASYBUILD_INSTALLPATH` when `EASYBUILD_PREFIX` is not defined before loading the custom module:
```
Lmod Warning:  Syntax error in file:
/apps/pilatus/UES/modulefiles/EasyBuild-custom/cscs.lua
 with command: setenv, one or more arguments are not strings.
```

The new modulefile will also keep different prefix and installpath values if they are exported before loading `EasyBuild-custom/cscs`, without overwriting the installpath with the chosen prefix: this is relevant on Eiger and Pilatus with the hierarchical module naming scheme, if a user wants to build new local modules on top of the ones provided by CSCS staff.

### DIFF
--- a/easybuild/module/EasyBuild-custom/cscs.lua
+++ b/easybuild/module/EasyBuild-custom/cscs.lua
@@ -70,20 +70,21 @@ elseif system == "pilatus" then
 else
 	LmodError("System ", system, " is currently unsupported\n")
 end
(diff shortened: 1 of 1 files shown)


jgphpc (approved):

kraushm (approved):
lgtm.

PR #2376, opened 2021-06-30
[eiger pilatus] Deploy Reframe 3.6.3 on eiger/pilatus

teojgo:


jgphpc (approved):
I believe we will move to 21.06 soon

PR #2387, opened 2021-07-14
[daint dom] Fix MATLAB

lucamar:
I provide a fix for [UES-1617](https://jira.cscs.ch/browse/UES-1617) and [UES-1619](https://jira.cscs.ch/browse/UES-1619), replacing MATLAB `R2019a` with the latest `R2021a` in the production lists of Dom and Piz Daint:
- LooseVersion in the easyblock `matlab.py` is fixed with Python3 following [a similar issue](https://github.com/easybuilders/easybuild-easyblocks/issues/1792)
- I added a `FIXME` in the easyconfig file `MATLAB-R2021a.eb` since the `open()` function raises an error only with EasyBuild 4.4.0 and newer, not with 4.3.4 (`Failed to copy 'local_filehandle' easyconfig parameter: cannot serialize '_io.TextIOWrapper' object`)

### DIFF
--- a/easybuild/easyconfigs/m/MATLAB/MATLAB-R2021a.eb
+++ b/easybuild/easyconfigs/m/MATLAB/MATLAB-R2021a.eb
@@ -14,15 +14,12 @@ sources = ['/apps/common/UES/easybuild/sources/m/%s/%s.tar' % (name, version)]
 java_options = '-Xmx2048m'
 
 # read local license file
-import os as local_os
 local_licensefile = '/apps/common/UES/easybuild/sources/m/MATLAB/' + '%s-%s.txt' % (name, version)
-local_filehandle = local_os.open(local_licensefile, local_os.O_RDONLY)
-local_bytext = local_os.read(local_filehandle, 1000)
-local_text = local_bytext.decode('utf-8')
-# FIXME: the open() function raises the error below only with EasyBuild 4.4.0 and newer, not with 4.3.4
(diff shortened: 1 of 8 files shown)


easybuild/easyconfigs/m/MATLAB/MATLAB-R2021a.eb
```diff
@@ -0,0 +1,33 @@
...
+
+# read local license file
+import os as local_os
+local_licensefile = '/apps/common/UES/easybuild/sources/m/MATLAB/' + '%s-%s.txt' % (name, version)
+local_filehandle = local_os.open(local_licensefile, local_os.O_RDONLY)
+local_bytext = local_os.read(local_filehandle, 1000)
+local_text = local_bytext.decode('utf-8')
+# FIXME: the open() function raises the error below only with EasyBuild 4.4.0 and newer, not with 4.3.4
+# Failed to copy 'local_filehandle' easyconfig parameter: cannot serialize '_io.TextIOWrapper' object
+# with open(local_licensefile, 'r') as local_filehandle:
```
teojgo, line 24: Can you try the prior solution with the normal `open` and just set `local_filehandle` to `None` after the `with` block?

teojgo (changes requested):

easybuild/easyconfigs/m/MATLAB/MATLAB-R2021a.eb
lucamar: Thanks for the advice @teojgo, it works: I will push the change in the pull request

teojgo (approved):

PR #2390, opened 2021-07-14
[eiger pilatus] Fix EasyBuild-custom (Lua)

lucamar:
The pull request should fix a bug with the Lua modulefile, only used on Eiger and Pilatus:
- it will be possible to set `EASYBUILD_INSTALLPATH` and a different `EASYBUILD_PREFIX` before loading the module
- it should not be necessary to add the folder with already installed modules to the MODULEPATH, since that is done by EasyBuild


kraushm (approved):
lgtm.

PR #2391, opened 2021-07-14
[eiger pilatus] Remove PyCall with PE 21.06

lucamar:
The extension `PyCall` is removed to avoid build failure on Alps with PE 21.06.


victorusu:
@omlins, this package is listed since the beginning of the JuliaExtensions. Do you know if there is any particular reason for having this package?

omlins (approved):
We can remove PyCall from the pre-installed modules if it causes troubles. The reason why it was there is that it seems to be package that many users might want to use, nothing else.

PR #2392, opened 2021-07-14
[eiger pilatus] Add missing recipes of ParaView

lucamar:
Missing recipes of `ParaView` on Alps added to the production repository.


lucamar:
The failure on Eiger might be due to a temporary issue in the environment, since the module has been already installed successfully on the system.

PR #2393, opened 2021-07-14
[eiger pilatus] Add metadata 21.07 and simpler toolchain recipes

lucamar:
Add metadata file for PE 21.07 and simpler toolchain recipes (EasyBuild 4.4.0 or more recent required).


lucamar:
Recipes for PE 21.04 failed on Pilatus, as expected since that PE has been removed from the system.

PR #2397, opened 2021-07-15
[dom daint arolla tsa pilatus eiger] Fix for ReFrame autodetection

teojgo:


easybuild/easyconfigs/r/reframe/reframe-3.7.0.eb
```diff
@@ -47,5 +47,7 @@ sanity_check_paths = {
     'dirs':  ['bin', 'cscs-checks', 'cscs-checks/private', 'unittests',],
 }
 
-sanity_check_commands = ['reframe -V', 'python3 -m pytest --version']
+# FIXME: We run `reframe -l` to enforce autodetection the first time ReFrame runs
+# see https://github.com/eth-cscs/reframe/issues/2072
+sanity_check_commands = ['reframe -V', 'reframe -l', 'python3 -m pytest --version']
```
victorusu, line 52: @teojgo, do you also need to run the 'remote' autodetection?

teojgo: No, it seems to work without it.

victorusu: Does it make sense to also do it?

teojgo: I don't think so for the moment.

victorusu (approved):
`lgtm`

teojgo:
retest this please

easybuild/easyconfigs/r/reframe/reframe-3.7.0.eb
vkarak: As it is now, I don't think it would work for users if they request remote auto-detection, because it would still try to write to that directory. But at least it wouldn't block everything. I would vote for remote auto-detection here too, just to be on the safe side.

victorusu (commented):
@teojgo, what do you think?

teojgo:
Closing since this is going to be fixed with a patch version of ReFrame

PR #2398, opened 2021-07-15
[daint dom eiger pilatus tsa] UpdateEB pipeline and util.groovy

lucamar:
Adjusts the pipeline script of `UpdateEB` and the file `util.groovy`, removing outdated pipeline script for Pilatus.


jenkins/JenkinsfileUpdateEB
```diff
@@ -144,11 +143,8 @@ stage('Update Stage') {
                         $load_easybuild
 
                         # update buildpath and metadata file for external modules
-                        export EASYBUILD_BUILDPATH="\$EASYBUILD_BUILDPATH/$machineLabel"
+                        export EASYBUILD_BUILDPATH="/dev/shm/\$USER/$machineLabel"
```
victorusu, line 146: I think one should set this path as a variable like it was before. If one doesn't want to change the `${EASYBUILD_BUILDPATH}`, then I think another variable like `${_local_easybuild_buildpath} would fit.

victorusu (changes requested):

jenkins/JenkinsfileUpdateEB
teojgo: You can always use the `withEnv` like here: https://github.com/eth-cscs/production/blob/ff3722c3b11bb5128cfb3753c4a9afd258f4426f/jenkins/JenkinsfileTestingEB#L72 to avoid exporting manually inside the script.

teojgo (changes requested):

jenkins/JenkinsfileUpdateEB
teojgo: It's fine, let it be like that. Although I think that the USER would be defined as long as you logn to the node.

teojgo (approved):

victorusu (approved):
`lgtm`

PR #2415, opened 2021-07-22
[eiger] Fix PLUMED 2.7.1

lucamar:
I provide a fixed recipe for PLUMED with PE 21.05 and 21.06 that avoids linking the non-threaded release of `cray-libsci` (see [SD-52306](https://jira.cscs.ch/browse/SD-52306)).


lucamar:
The recipe does not seem to be effective with the user `jenscscs` as it is with my username. As a matter of fact, the test linked the non-threaded release of `cray-libsci`, that is not linked instead n my local build: 
```
$ lddtree /scratch/e1000/jenscscs/TestingEB/eiger/5b9224c-5166/software/PLUMED/2.7.1-cpeGNU-21.06/lib/libplumedKernel.so | grep libsci
    libsci_gnu_82.so.5 => /opt/cray/pe/lib64/libsci_gnu_82.so.5
```
Updating the recipe of the toolchain `cpeGNU/21.06` with the system default `gcc/10.2.0`, therefore I will avoid using Cray wrappers and rely on linear algebra routines provided by the dependency `GSL` instead.

victorusu (approved):
`lgtm`

PR #2417, opened 2021-07-23
WIP [pilatus] Updated recipes for PE 21.07

jenkins-cscs:


lucamar:
THe recipes have been already deployed on Pilatus by the update pipeline, therefore they can be merged without testing.

PR #2423, opened 2021-08-02
WIP [eiger] Updated recipes for PE 21.06

jenkins-cscs:


lucamar:
There is no need to test this pull request, since the update pipeline has already built `CubeLib` with PE 21.06 on Eiger.

PR #2427, opened 2021-08-06
[dom] Julia 1.6.2: make user hold registries and update versions (PE 20.11)

### LINKED ISSUE #2407: Remove global julia registry
Currently we have a site-wide julia registry when using `module load Julia`, since we're setting `JULIA_DEPOT_PATH`.

The upside of this is that everybody has the same registry, the downside is that users can't update the registry. This means that users can't get new packages after Julia has been installed on the system, other than unsetting JULIA_DEPOT_PATH.

Since Julia packages are often fast release cycles, it'd be great if any user can update the registry for themselves (that is: run `] up`).

(assigning @omlins so that he gets notified ;) )

omlins:
fixes #2407

### DIFF
--- a/easybuild/easyblocks/julia.py
+++ b/easybuild/easyblocks/julia.py
@@ -30,6 +30,7 @@ EasyBuild support for building and installing Julia packages, implemented as an 
 """
 import os
 import socket
+import shutil
 
 from easybuild.tools.config import build_option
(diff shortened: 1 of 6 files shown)


omlins:
This PR is replaced by: https://github.com/eth-cscs/production/pull/2508

PR #2441, opened 2021-08-19
[daint-mc] Add OpenFOAM v2106 for CrayGNU 20.11

kraushm:
Updated source URL as well as dependencies using cray-tpsl instead of Scotch built using EB.


lucamar (approved):
Thanks @kraushm

PR #2443, opened 2021-08-19
[daint] Update 7.0.UP02-20.11-daint-gpu

twrobinson:
Update of jupyterlab


lucamar (commented):
Thanks @twrobinson, but why are you changing default version?

twrobinson:
> Thanks for updating the package @twrobinson, but why are you changing default version?
> If the reason is due to a bugfix, then I would simply remove the older one from production.

As per Slack conversation.

lucamar (approved):

PR #2463, opened 2021-09-09
[daint dom] Remove HPX from production lists and clean up recipe folder

lucamar:
I have removed HPX from the production lists on Dom and Piz Daint (the current modules on the systems won't be removed, though). 
I have also moved outdated recipes to the deprecated folder, keeping only the ones corresponding to the Cray PEs that are available on the systems (20.08, 20.11 and 21.02).


msimberg (approved):
Thanks @lucamar!

PR #2469, opened 2021-09-10
Remove legacy spack

haampie:
This does a bit more than the PR title:

- Remove 0.15.x config files and code, it's been a year since 0.15 and it'll not take very long until 0.17 :crossed_fingers:. Most people tend to use spack's develop branch anyways. The main reason for this is that the script actually works, since 0.15 does not support parallel builds (so, multiple processes calling spack install, like reframe does).
- Use `-[cdt version]` as a suffix, so that we have a more visual `[cray compiler/package version]-[cdt version]` separation.


haampie:
Okay, Victor told me we do rely on "ancient" spack... let me change it s.t. spack uses the reframe stage dir for spack stage&install dirs.

PR #2474, opened 2021-09-15
Use the reframe stage dir for spack stage & install

haampie:
To avoid build errors with spack 0.15.x we are better off making spack use the reframe stage dir throughout


victorusu:
`lgtm`

PR #2484, opened 2021-09-21
Use short dir for reframe stage in spack test

haampie:
The store/database of spack moved inside the reframe stage dir for each
test, but that can cause errors about long path names. Simple solution
is to make reframe create a stage dir in /tmp


victorusu (approved):
`lgtm`

PR #2493, opened 2021-10-04
[daint, dom] Add ipcmagic to jupyterlab

rsarm:
Adds new version of ipcmagic to the production jupyterlab. Also bump jupyterhub component to match server VM. Tested on daint.


twrobinson (approved):
Great, thanks Rafa. I took liberty of upgrading the jupyterhub component to match the version on jupyter VM. This prevents a cosmetic-only warning in the user log file.

PR #2495, opened 2021-10-06
[eiger] jupyterlab-2.2.10-cpeGNU-21.08.eb

twrobinson:
upgrading widgets infrastructure to support custom kernels that use current pip install defaults. otherwise widgets will fail to render in jupyterlab
This fixes SD-53474
Compatibility table is shown here: https://github.com/matplotlib/ipympl


twrobinson:
retest this please

PR #2524, opened 2021-11-12
[dom daint tsa eiger pilatus] Add recipe for Skopeo 1.5.1

teojgo:


jgphpc (approved):
build and run fine on pilatus except that I get:

```
ERRO[0000] The storage 'driver' option must be set in /etc/containers/storage.conf, guarantee proper operation.
```

Should it be set ?
```
grep ^driver /etc/containers/storage.conf
driver = ""
```

PR #2527, opened 2021-11-16
[tsa] Updated CDO with ecCodes and libaec

lucamar:
I provide the recipes of CDO, ecCodes and libaec with the toolchains `foss` and `fosscuda`, version `2019b` (see [SD-53816](https://jira.cscs.ch/browse/SD-53816)).


lucamar:
The download of CDO often fails, therefore I have replaced the download link with the local tarball.

kraushm (approved):
lgtm, thanks.

PR #2529, opened 2021-11-16
[pilatus] configurable-http-proxy-4.2.3-cpeGNU-21.09.eb

twrobinson:
Adds configurable-http-proxy, which is a jupyterlab dependency.


lucamar (approved):
Thanks @twrobinson

PR #2530, opened 2021-11-16
[pilatus] graphviz-2.46.1-cpeGNU-21.09.eb

twrobinson:
Adds graphviz, which is a dep of jupyterlab


lucamar (approved):
Thanks @twrobinson

PR #2540, opened 2021-11-26
[dom daint pilatus eiger] Remove disabling of remote autodetection

teojgo:


vkarak (commented):
Lgtm, but we don't need to redeploy this module. We can remove the environment variable for the next release. What do you think?

teojgo:
> Lgtm, but we don't need to redeploy this module. We can remove the environment variable for the next release. What do you think?

Yeah, I just need this so that I base the next production PR to this.

vkarak (approved):

PR #2550, opened 2021-12-10
[dom-mc] spack_config: fix dom-mc install path

albestro:
What about the old install path? How does it get cleaned up?


ekouts (approved):

haampie:
this is superseded by #2563 #2564

PR #2551, opened 2021-12-10
[daint-mc] spack_config: fix install path on daint-mc

albestro:
Waiting for suggestions about what to do in #2550 before fixing also this one the same way.


haampie:
this is superseded by #2563 #2564

PR #2554, opened 2021-12-13
[pilatus] restore sanity check and update LD_LIBRARY_PATH

jfavre:
== COMPLETED: Installation ended successfully (took 9 mins 32 secs)
== Results of the build can be found in the log file(s) /scratch/e1000/jfavre/eiger/software/VTK/9.1.0-cpeCray-21.09-OSMesa-python3/easybuild/easybuild-VTK-9.1.0-20211213.110111.log

PR #2546, opened 2021-12-09
[eiger] fix LD_LIBRARY_PATH and restored sanity check

jfavre:
== COMPLETED: Installation ended successfully (took 16 mins 6 secs)
== Results of the build can be found in the log file(s) /scratch/e1000/jfavre/eiger/software/VTK/9.1.0-cpeCray-21.08-OSMesa-python3/easybuild/easybuild-VTK-9.1.0-20211209.110912.log
== Build succeeded for 1 out of 1

PR #2558, opened 2021-12-13
[dom-mc] Paraview with Mesa

lucamar:
I provide the recipe of ParaView with Mesa and dependencies with CDT 21.09 on Dom: I have removed the suffix `-python3`, since only Python3 is supported.

Please note that the build of ParaView would fail on GPFS due to the issue reported in [SD-53990](https://jira.cscs.ch/browse/SD-53990), since the `postinstallcmds` use `git clone`.


lucamar:
The failure is due to the link error with VTK, that has been fixed on Alps (see #2546 and #2554):
```
/usr/bin/ld: ../../lib64/libvtkCommonCore-pv5.9.so.5.9: undefined reference to `tbb::internal::get_initial_auto_partitioner_divisor()'
```

jfavre (approved):
I have not tried Mesa 21.3.1 but we can try

PR #2559, opened 2021-12-14
[dom-gpu] ParaView with EGL

lucamar:
I provide the recipe of ParaView with EGL and dependencies with CDT 21.09 on Dom: I have removed the suffix `-python3`, since only Python3 is supported.

Please note that the build of ParaView would fail on GPFS due to the issue reported in [SD-53990](https://jira.cscs.ch/browse/SD-53990), since the `postinstallcmds` use `git clone`.


jfavre (approved):
Thanks @lucamar

PR #2560, opened 2021-12-14
[dom] Update lists and recipes for Dom

lucamar:
I provide some minor updates for the new production lists and recipes with Cray PE 21.09 on Dom, including `NCO` @omlins and `numpy` @henrique (is it still needed? The recent `cray-python/3.9.4.1` features `numpy` release `1.20.1`).


henrique (approved):
lgtm, thanks Luca
I guess keeping both won't hurt...(it might be nice to run some numpy benchmarks and actually get some data on that)
anyway, AFAIK we can remove the numpy dependency from many eb recipes that already have cray-python, and don't really use numpy for performance (e.g. the deep learning stuff)

PR #2561, opened 2021-12-14
[dom] NAMD: Temporary fix for build on dom with Intel 21.09

ChristopherBignamini:


lucamar (approved):
Thanks @ChristopherBignamini

PR #2563, opened 2021-12-15
[daint] bump spack-config

haampie:


albestro:
do we have to do any cleanup about older install path?

haampie:
I think it's fine if they remain there or are removed only later, the default is now spack-config/1.1

lucamar:
Thanks @haampie: older modules are not removed automatically, we do that manually if needed.
 
Is there any connection to the pull requests #2504 and #2505 submitted by @albestro?

haampie:
They're not really related, we're moving away from storing config files in https://github.com/eth-cscs/production/tree/master/spack and instead generate them on the system through this `spack-config` package. The `spack-config` module is not yet documented but once we do that we don't really need the https://github.com/eth-cscs/production/tree/master/spack dir anymore.

albestro:
> Thanks @haampie: older modules are not removed automatically, we do that manually if needed.

Apart from the change of version, I want to highlight there was an error about misplaced things (iirc cpu modules installed in gpu folder, or something like that).

> Is there any connection to the pull requests #2504 and #2505 submitted by @albestro?

I think we can opt for closing them, since in the end this new thing will likely replace the previous method that was generating spack configurations via reframe runs.

easybuild/easyconfigs/s/spack-config/spack-config-1.1-cdt-20.11.eb
```diff
@@ -0,0 +1,42 @@
+easyblock = 'ConfigureMake'
+
+name = 'spack-config'
+version = '1.1'
+_cdt = 'cdt-20.11'
+
+homepage = 'https://github.com/eth-cscs/spack'
```
albestro, line 7: minor: do you want to fix also this (in both files)?

albestro (approved):
Just a minor request change, for the rest I'm ok with it. We can take care of older install paths separately.

PR #2568, opened 2021-12-16
[pilatus] Update Python packages

lucamar:
I provide updated recipes for Python packages on Pilatus with Cray PE 21.09 as in  #2566.


rsarm (approved):

jfavre (commented):
lgtm

PR #2569, opened 2021-12-17
[daint dom eiger pilatus tsa] Update compiler definitions and external metadata

lucamar:
Update Intel compiler options for oneAPI and add external metadata for CPE 21.10 and 21.11.


kraushm (approved):
lgtm, thanks!

PR #2571, opened 2021-12-17
[dom-gpu] PyTorch with 21.09

lucamar:
I have adapted the recipe of PyTorch 1.10.1 with the latest Cray PE 21.09 on Dom.


rsarm:
retest this please

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
```diff
@@ -0,0 +1,85 @@
...
+    ('wheel', '0.37.0')
+]
+
+dependencies = [
+    ('Boost', '1.77.0', '-python%(pymajver)s'),
+    ('cray-python', EXTERNAL_MODULE)
+]
+
+preinstallopts = ' export HOROVOD_BUILD_CUDA_CC_LIST=60; '
+preinstallopts += ' export HOROVOD_CUDA_HOME=/usr/local/cuda-11.2; '
```
henrique, line 27:
```suggestion
preinstallopts += ' export HOROVOD_CUDA_HOME=$CUDATOOLKIT_HOME; '
```
however, this wasn't necessary before, so we shouldn't need to define it?

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
```diff
@@ -0,0 +1,85 @@
...
+preinstallopts += ' export HOROVOD_GPU_OPERATIONS=NCCL; '
+preinstallopts += ' export HOROVOD_MPICXX_SHOW="CC --cray-print-opts=all"; '
+preinstallopts += ' export HOROVOD_NCCL_HOME=/opt/nvidia/hpc_sdk/Linux_x86_64/21.3/comm_libs/nccl; '
```
henrique, line 30:
```suggestion
preinstallopts += ' export HOROVOD_NCCL_HOME=$EBROOTNCCL; '
```

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
```diff
@@ -0,0 +1,85 @@
+# contributed by Luca Marsella (CSCS)
+easyblock = 'PythonBundle'
+
+name = 'PyTorch'
+version = '1.10.1'
+_ptcudaver = 'cu113'
```
henrique, line 6: This is probably fine, but should we use 11.1 and cudatoolkit/11.1.0_3.39-4.1__g484e319 for consistency?
```suggestion
_ptcudaver = 'cu111'
```
Available wheels: https://download.pytorch.org/whl/torch_stable.html

henrique (commented):
Thanks Luca

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
```diff
@@ -0,0 +1,85 @@
...
+    ('cloudpickle', '2.0.0'),
+    ('iniconfig', '1.1.1'),
+    ('Pillow', '8.4.0', {'modulename': 'PIL'}),
+    ('psutil', '5.8.0'),
+    ('pycparser', '2.21'),
+    ('PyYAML', '6.0', {'modulename': 'yaml'}),
+    ('pytest', '6.2.5', {
+        'source_tmpl': '%(name)s-%(version)s-py%(pymajver)s-none-any.whl'
+    }),
+    ('toml', '0.10.2'),
```
henrique, line 57: TOML? :) out of curiosity, but where does this comes from?
pytest is probably also not necessary... but scipy is for the reframe tests
I'm trying a similar recipe on 20.11 and can share it later on

lucamar: I had to add `toml` otherwise the build was failing, same for `pytest`. The package `scipy` is provided by `cray-python` instead and the ReFrame checks triggered by `pytorch_horovod_check.py` passed with my local modulefile on Dom.

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
lucamar: I wished to try the latest `CUDA 11.2` available on Dom and it makes the test pass with the extensions `cu113` provided by `torch` (tested locally with the modulefile in my sandbox).

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
lucamar: Thanks @henrique, I fixed it

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
henrique: hm... it might be fine, but if we can, it might be better to use the same cuda version everywhere. Here you are building Horovod with cuda 11.2 but using torch built with 11.3, which isn't even available for us yet. I guess that could cause some problems for users trying to build [torch extensions](https://pytorch.org/tutorials/advanced/cpp_extension.html) or some libraries like Nvidia APEX, MS DeepSpeed, etc

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
henrique: Fair enough, thanks!
if you have the module built, please make sure the reframe tests run with the provided version of scipy (I guess it should tho)

easybuild/easyconfigs/p/PyTorch/PyTorch-1.10.1-CrayGNU-21.09.eb
lucamar: I have switched the `torch` extensions to `cu111` as you suggested @henrique: I load the default `cudatoolkit` now, that is 11.0. The regressions tests pass also in this case, it seems that minor CUDA versions do not make any difference there.

henrique (approved):
Cheers Luca

PR #2573, opened 2021-12-20
[dom-gpu] Update Horovod for CrayGNU-21.09

rsarm:


easybuild/easyconfigs/p/prereq/prereq-tf-2.4-CrayGNU-21.09.eb
```diff
@@ -0,0 +1,65 @@
...
+
+homepage = 'https://pypi.python.org/pypi'
+description = "This module is a bundle of Python packages on Cray systems based on the module cray-python"
+
+toolchain = {'name': 'CrayGNU', 'version': '21.09'}
+toolchainopts = {'pic': True, 'verbose': False}
+
+dependencies = [
+    ('cray-python', EXTERNAL_MODULE),
+    ('numpy', '1.17.2'),
```
lucamar, line 15: @rsarm The module `cray-python` already provides `numpy` 1.20.1; if you want to load a module that we build due to performance reasons, there is the more recent version 1.21.4 available with the corresponding easyconfig file

easybuild/easyconfigs/h/Horovod/Horovod-0.21.0-CrayGNU-21.09-tf-2.4.0.eb
```diff
@@ -0,0 +1,34 @@
...
+toolchain = {'name': 'CrayGNU', 'version': '21.09'}
+toolchainopts = {'verbose': False}
+
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [
+    ('cray-python', EXTERNAL_MODULE),
+    ('cudatoolkit', EXTERNAL_MODULE),
+    ('numpy', '1.17.2'),
```
lucamar, line 19: @rsarm The module `cray-python` already provides `numpy` 1.20.1; if you want to load a module that we build due to performance reasons, there is the more recent version 1.21.4 available with the corresponding easyconfig file

lucamar:
The build failed since it could not find the package of the dependency `NCCL/2.8.3`: however, that should not be necessary any longer, since `NCCL` is provided by the module `nvidia` in `/opt/nvidia/hpc_sdk/Linux_x86_64/21.3/comm_libs/nccl` (see my comments above).

rsarm:
retest this please

lucamar (approved):
Thanks @rsarm

PR #2575, opened 2021-12-20
[dom] spack-config 1.2

haampie:


albestro (approved):
LGTM

minor comment: it's so bad that the version is embedded in the filename. can we do anything about it?

PR #2580, opened 2021-12-21
[pilatus] Paraview v510 cpecray 21.09

jfavre:


jenkins-builds/1.4.0-21.09-eiger
```diff
@@ -9,6 +9,7 @@
  GSL-2.7-cpeAMD-21.09.eb
  GSL-2.7-cpeCray-21.09.eb
  ParaView-5.9.1-cpeCray-21.09-OSMesa-python3.eb
+ ParaView-5.10.0-cpeCray-21.09-OSMesa.eb --set-default-module
```
lucamar, line 12:
```suggestion
 ParaView-5.10.0-cpeCray-21.09-OSMesa.eb
```

lucamar (changes requested):

jenkins-builds/1.4.0-21.09-eiger
lucamar: Thanks @jfavre: I have suggested to remove the old version from the production list, since the module won't be removed. Anyway, the flag to update the default is not needed, as Lmod takes the last module as default

jfavre: Thanks @lucamar for your tips

lucamar (approved):
Thanks @jfavre

PR #2598, opened 2022-01-05
[dom-gpu] Create jupyterlab-2.2.10-CrayGNU-21.09-batchspawner-cuda.eb

twrobinson:
- incorporated IPython and its deps 
- incorporated dask to help ensure compatibility with labextension 
- updated many python modules to work with 3.9. 
- most labextensions are working. Topbar extension is not yet working.
- ipyparaview functionality needs to be tested.
- removed bqplot (can be readded if requested by users)
I would like this to be merged to be able to test on Dom in prod-like environments.


lucamar (approved):
Thanks @twrobinson

PR #2613, opened 2022-01-14
[daint-mc] BAGEL with Cray

lucamar:
Tentative recipe of BAGEL with CrayIntel 20.11 for [SD-54169](https://jira.cscs.ch/browse/SD-54169): I had to use Boost 1.70.0 as a dependency since 1.75.0 was making the build fail. 
The configure option `--with-mpi=intel` is due to the limited choices available (`intel`, `mvapich` or `openmpi` only): the executable [needs to be tested beforehand](https://nubakery.org/quickstart/how_to_run_bagel.html).

### DIFF
--- a/easybuild/easyconfigs/b/BAGEL/BAGEL-1.2.2-CrayGNU-20.11.eb
+++ b/easybuild/easyconfigs/b/BAGEL/BAGEL-1.2.2-CrayGNU-20.11.eb
@@ -8,13 +8,14 @@ homepage = "http://www.nubakery.org"
 description = """BAGEL (Brilliantly Advanced General Electronic-structure Library)
 is a parallel electronic-structure program."""
 
-toolchain = {'name': 'CrayIntel', 'version': '20.11'}
+toolchain = {'name': 'CrayGNU', 'version': '20.11'}
 
 source_urls = ['https://github.com/nubakery/bagel/archive/']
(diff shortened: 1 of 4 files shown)


ajocksch (approved):

lucamar:
The build failed with several errors of the type below when calling Boost templates:
```
this candidate was rejected because mismatch in count of arguments
```

lucamar:
I have siwtched to `CrayGNU` and the build worked locally. Unfortunately, I get the following error at runtime:
```
   ERROR: EXCEPTION RAISED:  MPI_THREAD_MULTIPLE not provided
```
I think that selection `--with-mpi=intel` is the origin of the issue, therefore other options should be tested.

lucamar:
The error message at runtime comes from the file [mpi_interface.cc](https://github.com/qsimulate-open/bagel/blob/master/src/util/parallel/mpi_interface.cc):
```
#ifdef HAVE_MPI_H
  int provided;
  MPI_Init_thread(0, 0, MPI_THREAD_MULTIPLE, &provided);
  if (provided != MPI_THREAD_MULTIPLE)
    throw runtime_error("MPI_THREAD_MULTIPLE not provided");
```

lucamar:
I have defined the following runtime variable to allow multiple threads calling MPI:
```
export MPICH_MAX_THREAD_SAFETY=multiple
```

PR #2615, opened 2022-01-21
[dom-gpu] VASP 6.2.1

lucamar:
Recipe for the CUDA port of VASP 6.2.1 in production (it might still outperform the OpenACC port in some cases, therefore I keep it default although deprecated).


ChristopherBignamini (approved):
lgtm

PR #2616, opened 2022-01-21
[dom] spack-config update: add pgi compiler

albestro:
Testing on dom the latest spack-config which add `pgi` compiler to the provided config.


teojgo:
ok to test

haampie (approved):

PR #2619, opened 2022-01-24
[dom] Update JupyterLab release in production

twrobinson:
Adds Jupyterlab 3.2.8 and removes the temporary Jupyterlab 2, which will not be supported.


lucamar (approved):
Thanks @twrobinson

PR #2623, opened 2022-01-25
[dom] Update jupyterlab-3.2.8-CrayGNU-21.09-batchspawner-cuda.eb

twrobinson:
Adds nbresuse module which was missing and prevented the memory usage monitor from working


lucamar (approved):
Thanks @twrobinson

PR #2626, opened 2022-01-26
[pilatus] CP2K 9.1 with SIRIUS

lucamar:
I provide the recipe for CP2K 9.1 with the latest release of SIRIUS that includes ELPA.


mtaillefumier (approved):
looks fine to me. It is a CPU only build so things are much simpler.

PR #2630, opened 2022-01-27
[dom] Dom production lists as symlinks

lucamar:
Set Dom production lists as symbolic links to Piz Daint ones, as for Eiger and Pilatus: this helps keeping the production lists in sync for the same Cray PE release.


rsarm (approved):

lucamar:
I have updated the production list to use the environment variable `$APPS`and the variable `$system` defined and now exported by `production.sh` as well in the path, in order to avoid using the hard-coded machine name (local test worked): we could actually use only the `$system` variable that we define in `production.sh`, without depending on environment variables that might change.

kraushm (approved):
lgtm. Thanks!

teojgo (approved):

PR #2634, opened 2022-01-31
[daint] bump spack-config/1.5

haampie:


albestro (approved):
Just pointing out on daint we are doing a 1.3 -> 1.5. So this will add both `pgi` compiler and `pkg-config`.

teojgo (approved):

PR #2643, opened 2022-02-03
[dom-mc] Visit v3.2.2 for cdt21.09

jfavre:


jfavre:
== COMPLETED: Installation ended successfully (took 58 mins 59 secs)
== Results of the build can be found in the log file(s) /scratch/snx3000tds/jfavre/daint/software/Visit/3.2.2-CrayGNU-21.09/easybuild/easybuild-Visit-3.2.2-20220203.170948.log

easybuild/easyconfigs/v/Visit/Visit-3.2.2-CrayGNU-21.09.eb
```diff
@@ -0,0 +1,84 @@
...
+# * The original intent was to compile only the server-side components, disabling the
+# use of the GUI client for loggin nodes. Yet, this also removes the cli, preventing
+# us to run python-based batch-mode scripts. Thus, we build all components [default]
+
+
+# In order to use the pre-downloaded version use the source below
+sources = ['/apps/common/UES/easybuild/sources/%(nameletterlower)s/%(name)s/%(namelower)s-%(version)s.tar.gz']
+
+builddependencies = [
+    ('CMake', '3.14.5', '', True),
```
lucamar, line 56: Thanks @jfavre: do you need to keep using CMake 3.14.5 to buil VisIt, or can you also use a more recent version?

jfavre: Visit uses its own global builder to compile all its dependencies and it actually requires cmake 3.14.7 which is very close to our own 3.14.5. I was able to compile with our own module to reduce the number of dependencies by one. I have not tested other CMake because the compilation is very long (about 1 hour)

lucamar (approved):
Thanks @jfavre , then let's keep `CMake` release `3.14.5` for VisIt

jfavre:
if you are trying to eliminate cmake 3.14 from production, then I could switch to Visit's internal cmake distribution, change a config option and get rid of the Easybuild build dependency. Keep it in mind for the future.

PR #2645, opened 2022-02-07
[dom-gpu] Add ipyparaview and corresponding postinstallcmds

teojgo:


jfavre (commented):
would seem allright but I don't know the rest of the install procedure...

PR #2650, opened 2022-02-11
[dom-gpu] Update VASP default

lucamar:
VASP 6.3.0 OpenACC will be the default as the CUDA-C GPU port is deprecated.


ChristopherBignamini (approved):
lgtm

PR #2679, opened 2022-02-21
[daint dom] Link MPI-only ELPA with CP2K

lucamar:
I suggest to link MPI-only ELPA with CP2K on the Cray XC systems, since `cray-mpich` does not provide a sufficient level of threading support according to ELPA requirements. See [SD-54759](https://jira.cscs.ch/browse/SD-54759) for more details.


mtaillefumier (approved):
i thought GPU support in elpa was not really working. The toolchain compiles elpa without threading support on Cray systems actually. So I would suggest the same, elpa + mpi + gpu no openmp.

PR #2681, opened 2022-02-22
[daint,dom] spack-config 1.6

haampie:
- Adds `upstreams.yaml` as part of `module load spack-config` for reuse of binaries
- Makes `pkg-config` the default `pkgconfig` provider, since that's what cray compilers rely on by default

Purposely not changing the default version (requires some real life testing on daint).


albestro (approved):
LGTM

PR #2682, opened 2022-02-22
[dom] JupyterLab CSCS ipykernel (proof of concept)

twrobinson:
This is a "Proof of concept" that installs a minimal ipykernel that 
- includes cray-python Python modules
- includes ipykernel and its dependencies from a separate centrally-installed venv
- sources a file in user's home directory, which would include "module load" commands, for example.

This isolates the user environment from the JuptyerLab environment and has other advantages such as the fact that new "module load.." commands can be sourced from a simple kernel restart rather than stopping the JupyterLab server itself.

This is ready
- ipykernel and its dependencies are currently unversioned. This could be fixed by (a) a requirements.txt or (b) listing all deps versions on the pip line (c) many other ways (FIXED)
- the writing of the kernel.json and launcher files is ugly as hell (EDIT: but it works)
- would be nice to have the cscs logo as the kernel logos (FIXED)

(this is an attempt to fix "module load TensorFlow" which breaks JupyterLab)

Note: when testing this... the file that is sourced is jupyterlab-cscs.env, without a leading full stop 

It can be tested at https://jupyter-dev.cscs.ch


rsarm (commented):
Should we now remove ipcmagic and the related packages? https://github.com/eth-cscs/production/blob/bd2acb7854e713f6ee338a43498b9631a006cadd/easybuild/easyconfigs/j/jupyterlab/jupyterlab-3.2.8-CrayGNU-21.09-batchspawner-cuda.eb#L257-L261

I think docopt, and ipyparallel are needed only by ipcmagic. Not sure about pyexpect.

twrobinson:
> 

But I thought the idea was that we would leave the default kernel as it is, with a bunch of functionality that we know works together, and only "clean out" the new CSCS kernel.

twrobinson:
retest this please

twrobinson:
retest this please

rsarm (approved):

PR #2683, opened 2022-02-22
Revert "[daint,dom] spack-config 1.6"

lucamar:
Reverts eth-cscs/production#2681: I did not realised that @haampie renamed the 1.5 configuration file.
You also need to correct the entry in the production list, since [the CI/CD fails](https://jenkins.cscs.ch/blue/organizations/jenkins/ProductionEB/detail/ProductionEB/2163/pipeline) with the error below:
```
ERROR: One or more files not found: spack-config-1.5-cdt-21.09.eb
```
Otherwise, you could keep the older configuration files instead of renaming them: it is important that the most recent one is present in the [production list]().

### DIFF
--- a/easybuild/easyconfigs/s/spack-config/spack-config-1.5-cdt-21.09.eb
+++ b/easybuild/easyconfigs/s/spack-config/spack-config-1.5-cdt-21.09.eb
@@ -1,7 +1,7 @@
 easyblock = 'ConfigureMake'
 
 name = 'spack-config'
-version = '1.6'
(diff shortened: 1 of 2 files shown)


albestro:
> Otherwise, you could keep the older configuration files instead of renaming them: it is only important that the most recent one is present in the production list.

Ah, I was not aware of that. Thanks for pointing that out!

lucamar:
Let's close this pull request and merge #2684 instead. Thanks!

PR #2684, opened 2022-02-22
[dom,daint] Bring back spack-config/1.5

haampie:
Replaces #2683


albestro (approved):
I fully trust you @haampie, even because without any reference/diff I cannot check it. 😄

lucamar:
> @haampie do we have to fix also the production list as @lucamar said?

@albestro If you want to have the latest `spack-config/1.6` in production, you can open a new pull request: this one will only fix the CI/CD, as it restores `spack-config/1.5` that is currently appearing in the production lists.

PR #2688, opened 2022-02-23
[daint dom] Update PLUMED

lucamar:
I provide updated recipes for PLUMED last release with GNU and Intel compilers: I also update the production lists (2.7.3 should be the default, so I corrected it, adding the last release as non default) and clean-up the repository moving outdated recipes to the folder `deprecated`.


victorusu (approved):
`lgtm`

PR #2691, opened 2022-02-25
[daint-gpu] CP2K and SIRIUS with latest ELPA

lucamar:
I provide the recipes of CP2K and SIRIUS with the latest ELPA release `2021.11.002`.


lucamar:
The build fails since SIRIUS (CP2K dependency) is already built on Dom with the previous ELPA release (currently used in production on Piz Daint), therefore a conflict arises when loading the list of dependencies of CP2K.

PR #2693, opened 2022-02-25
[daint] Prepare switch of login meta-modules

lucamar:
Create login meta-modules for Cray PE 21.09 and remove obsolete ones before switching the symbolic link `/apps/daint/UES/easybuild/modulefiles`.


login/daint-gpu/21.09
```diff
@@ -33,23 +31,24 @@ if { [module-info mode remove] || [module-info mode switch] } {
...
- module unload cdt cdt-cuda
 }
 
 # load dependency and update MODULEPATH
 module load craype-haswell
 module use "${gpupath}"
 module use "${toolsgpupath}"
 
+# load additional modules
+#module load xalt
```
lucamar, line 42: @victorusu We are currently not loading `xalt`, should we load it by default?

jgphpc (approved):

login/daint-mc/21.09
```diff
@@ -31,15 +31,16 @@ if { [module-info mode remove] || [module-info mode switch] } {
...
- module unload cdt cdt-cuda
 }
 
 # load dependency and update MODULEPATH
 module load craype-broadwell
 module use "${mcpath}"
 module use "${toolsmcpath}"
 
+# load additional modules
+#module load xalt
```
lucamar, line 42: Same question here about `xalt` as above @victorusu

login/daint-gpu/21.09
victorusu: yes, we should

login/daint-mc/21.09
victorusu: yes, we should

login/daint-gpu/21.09
victorusu: Wait... @lucamar, @gppezzi and @vkarak, I am not sure... xalt is loaded by default from the system point of view... Here it is about "loading" it when `module load daint-{gpu,mc}`... I am not sure about this...
Because if the user is unloading it already, then it is a bit annoying to unload it another time, just because it has loaded `daint-{gpu,mc}`, no?
I think we should leave it as it is ... What do you think?

login/daint-mc/21.09
victorusu: I am not sure. Let's talk about this before changing the default behaviour..

victorusu (approved):
The PR `lgtm` the way it is, i.e. without loading xalt when `module load daint-{gpu,mc}`... I am not sure if we actually want to be on the way of those that `module purge` or `module unload xalt` in their `.bashrc`s...

kraushm (approved):
lgtm. Thanks!

login/daint-gpu/21.09
lucamar: That's fine with me, thanks for clarifying @victorusu

PR #2704, opened 2022-03-10
[daint-gpu] fixes ParaView for jupyter and plugin compilation

jfavre:


lucamar (approved):
Thanks @jfavre. As far as I can see among other changes, you have also downgraded `oidn` to rleease `1.4.1`: does it mean that the release 1.4.2 contributed to the issues with JupyterLab?

jfavre:
I resorted to compile ALL components of raytracing from source since one reason for the problems were the different TBB libraries. I now have all stuff relying only on the /opt/intel/oneapi/tbb/latest/ and version 1.4.1 of oidn is the version shipped (and thus supported by Kitware), and is also the version I could easily compile from source.

PR #2715, opened 2022-03-17
[daint-gpu dom-gpu] Add deepmd-kit-2.1.0

rsarm:


lucamar:
The build completed successfully in `TestingEB`, the status was not reported to GitHub though.

lucamar (commented):
We could also build `deepmd` with the [LAMMPS built-in mode](https://github.com/deepmodeling/deepmd-kit/blob/master/doc/install/install-from-source.md) using `-DLAMMPS_VERSION_NUMBER=20210929` with `CMake`. However, the [LAMMPS plugin-mode](https://docs.deepmodeling.org/projects/deepmd/en/master/install/install-lammps.html#install-lammps-plugin-mode) seems more straightforward with DeePMD-kit.

PR #2717, opened 2022-03-18
[daint] Update netcdf4-python

rsarm:
* Updates the recipe of netcdf4-python to address the issue in SD-54724
* Adds netcdf4-python-parallel


lucamar (changes requested):
Thanks @rsarm, I think that [setup-1.5.8.path](https://github.com/eth-cscs/production/pull/2717/files#diff-b7d28b37ba03c186a715397ab9dad796cdb18e89d3857f544ce76c8f643c6da4) should be renamed `setup-1.5.8.patch`

lucamar (approved):
Thanks @rsarm

PR #2739, opened 2022-05-04
[daint-gpu dom-gpu] Update 7.0.UP03-21.09-daint-gpu

jfavre:
update Visit to 3.2.2 to support Blueprint HDF5


lucamar:
Retest this please

jfavre:
> You are changing default, is it on purpose @jfavre? Please don't forget that the old module won't be removed automatically

yes, it is on purpose such that for the in-situ visualization class of next week, I can demonstrate the Blueprint reader. The Blueprint reader is the only change from the previous version and should not break anybody's workflow.

PR #2742, opened 2022-05-09
Namd 2.14 ph-fix

ChristopherBignamini:
Additional fix provided by NAMD developers added to the patch file


lucamar (approved):
Thanks @ChristopherBignamini: please inform NAMD users before applying the patch in production

PR #2754, opened 2022-06-20
[daint dom] Adding BCFtools and GATK .eb files for CrayGNU 21.09.

jgustavsen:
A follow-up from [SD-55649](https://jira.cscs.ch/browse/SD-55649). I have added two Easybuild files for BCFtools and GATK. 

This is my first contribution, so please let me know if something is missing. Thanks, Julia


jgustavsen:
Closing while investigating an issue I missed with BCFtools.

PR #2761, opened 2022-07-05
[eiger pilatus] Update ReFrame

lucamar:
I propose to update the default `reframe` module on Eiger and Pilatus to the latest release.


ekouts (approved):
Thanks, lgtm

PR #2785, opened 2022-09-13
[daint] Recipe of CLAPACK

lucamar:
I provide the recipe to build `CLAPACK` on Piz Daint (see [SD-56374](https://jira.cscs.ch/browse/SD-56374)).


lucamar:
The build succeeds on the login node (local installation) but fails on the compute nodes in `TestingEB`.

lucamar:
It seems that `buildininstalldir = 'True'` is needed to avoid building on `/tmp` of a compute node.

rsarm (approved):

PR #2786, opened 2022-09-16
[dom daint eiger pilatus tsa] Set xthostname=1 by default in reframe-cscs-tests

ekouts:
This is not necessary for the latest version of Reframe in the system but as soon we start using ReFrame 4.x we will need it. I am adding it so that we don't forget it since the new recipe will be deplyed automatically. I will also change [the template recipe](https://github.com/eth-cscs/cscs-reframe-tests/blob/main/JenkinsfileDeploy) to 22.09 after that.
```
def ebTemplate = '22.06'                        // The eb recipe of ReFrame to use as template
```
@teojgo am I forgetting anything?

### DIFF
--- a/easybuild/easyconfigs/r/reframe-cscs-tests/reframe-cscs-tests-22.09.eb
+++ b/easybuild/easyconfigs/r/reframe-cscs-tests/reframe-cscs-tests-22.09.eb
@@ -31,6 +31,7 @@ postinstallcmds = [
 modextrapaths = {
     'RFM_CONFIG_FILE': 'config/cscs.py',
     'RFM_CHECK_SEARCH_PATH': 'checks',
+    'RFM_AUTODETECT_XTHOSTNAME': '1',
 }
 
 sanity_check_paths = {


lucamar (commented):
Thanks @ekouts. The test on Dom might not pass as the system is still under maintenance and the runner on Eiger reported the following error message:
```
ERROR: ancient Tcl-only Environment Modules modules tool can not be used, 'modulecmd.tcl' command is not available; use --modules-tool to specify a different modules tool to use (EnvironmentModulesC, EnvironmentModulesTcl, EnvironmentModules, Lmod)
```
Sometimes running again solves the issue.

ekouts:
retest this please

teojgo:
I will merge once dom is back

teojgo (approved):

lucamar:
The failure on Eiger is expected due to the hostname change of the login nodes, therefore we can ignore it and merge.

ekouts:
Closing since we add manually the `RFM_AUTODETECT_XTHOSTNAME` in #2790 .

PR #2789, opened 2022-09-22
[daint-mc] URL change in tree

lukasgd:
Updating the URL as the FTP download fails and Gitlab is used for hosting according to http://mama.indstate.edu/users/ice/tree/.


lucamar:
retest this please

PR #2790, opened 2022-10-01
[daint eiger tsa] Add recipe for ReFrame CSCS tests version 22.10

jenkins-cscs:

PR #2792, opened 2022-10-06
[eiger pilatus] Fix system name

lucamar:
I update the system name match in the EasyBuild custom module addressing Eiger and Pilatus and removing outdated matches of dismissed systems: this should fix the error of [ProductionEB](https://jenkins.cscs.ch/view/Production/job/ProductionEB) on Eiger (see my comment in #2790 ).

### DIFF
--- a/easybuild/module/EasyBuild-custom/cscs
+++ b/easybuild/module/EasyBuild-custom/cscs
@@ -65,11 +65,10 @@ if { $status == 0 } {
 # SYSTEM SPECIFIC SETUP #
 #########################
 # current system name
-if { [ string match *esch* $::env(HOSTNAME) ] } {
- set system [regsub {[cl]n-[0-9]+} $::env(HOSTNAME) ""]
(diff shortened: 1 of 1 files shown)


teojgo (approved):

ekouts (approved):
It looks fine as it is, but would it be possible to use `/etc/xthostname`'s content, since it seems to be in all our systems?

lucamar:
> It looks fine as it is, but would it be possible to use `/etc/xthostname`'s content, since it seems to be in all our systems?

Thanks @ekouts, `/etc/xthostname` is defined on Dom, Eiger, Pilatus and Piz Daint, but not on Tsa: anyway we are not addressing the latter within `EasyBuild-custom`, so I think we can ignore it unless @kraushm has any objections.

lucamar:
I have included a command to match the hostnames of Arolla and Tsa, @kraushm. I have removed Escha/Kesch and Fulen that no longer exist.

kraushm (approved):
Thanks @lucamar, lgtm.

PR #2810, opened 2022-11-11
[eiger, pilatus] adding FFmpeg v5 for eiger

jfavre:


jfavre:
ERROR: libx264 not found
however the module exists and the library is there /apps/pilatus/UES/jenkins/1.4.0/software/x264/20191217-cpeGNU-21.12/lib/
so I don't have a clue.

jfavre:
retest this please

lucamar:
I think that the error reported by `TestingEB` is due to the issue described in [VCMSA-101](https://jira.cscs.ch/browse/VCMSA-101): as a matter of fact, the module `x264/20191217` is not available under the path of the default `cpeGNU/21.12`, which is loaded by Lmod instead of your local installpath with the module.

jfavre:
> I think that the error reported by `TestingEB` is due to the issue described in [VCMSA-101](https://jira.cscs.ch/browse/VCMSA-101): as a matter of fact, the module `x264/20191217` is not available under the path of the default `cpeGNU/21.12`, which is loaded by Lmod instead of your local installpath with the module.

you are most likely right. Unfortunately, I don't have the competence to fix VCMSA-101.

lucamar:
I will merge it, since you tested locally

PR #2818, opened 2022-11-21
[dom daint eiger pilatus tsa] Add recipe for ReFrame version 4.0.0-dev.3

jenkins-cscs:


lucamar:
Ok to test

ekouts:
I think we should probably skip this version since it has an important bug that will affect our tests: https://github.com/reframe-hpc/reframe/issues/2681

teojgo:
> I think we should skip this version since it has an important bug that will affect our tests: [reframe-hpc/reframe#2681](https://github.com/reframe-hpc/reframe/issues/2681)

I think it's ok because we don't deploy the dev versions in any case

lucamar:
Thanks for your comments @ekouts and @teojgo: I will just close the pull request without merging the recipe.

PR #2821, opened 2022-11-22
[daint-gpu, dom-gpu] fixes for SPH-EXA and adding explicit 3.9 numbering

jfavre:


lucamar (changes requested):
Thanks @jfavre: we started skipping the suffix `-python...` for recent recipes, is it possible to remove it from the recipe and dependencies?

lucamar:
I see that @henrique used the `versionsuffix`, so there won't be any confusion if you remove it. As a matter of fact, you are loading the default Python version with Cray PE 21.09, therefore I think that it makes sense to avoid setting explicitly the suffix: 
```
dependencies = [
    ('cray-python', EXTERNAL_MODULE),
```

PR #2823, opened 2022-11-25
[daint-gpu, dom-gpu] upgrading VMD from v1.9.4a54 to v1.9.4a57

jfavre:
== COMPLETED: Installation ended successfully (took 14 secs)
== Results of the build can be found in the log file(s) /scratch/snx3000/jfavre/daint/software/VMD/1.9.4a57/easybuild/easybuild-VMD-1.9.4-20221125.111622.log


lucamar (changes requested):
Thanks @jfavre: please note that in this way you need to force the re-build of the recipe to implement the change. I would suggest to change name of the recipe instead, according to the `versionsuffix` that you used: in this way, the change will be visible to users too

lucamar:
I have committed the name change that I have proposed directly to the branch of this pull request: please let me know if this is fine with you @jfavre .

lucamar (approved):

jfavre:
> I have committed the name change that I have proposed directly to the branch of this pull request: please let me know if this is fine with you @jfavre .

Yes, fine with me. Thanks.

PR #2824, opened 2022-11-25
[daint-gpu dom-gpu] Update and rename VMD-1.9.4-a57.eb to VMD-1.9.4a57.eb

lucamar:
Fix `versionsuffix`, recipe name and production list removing `-`.


jfavre (approved):
Thanks @lucamar !

PR #2849, opened 2023-02-01
[dom] Use different openmpi library

rsarm:
This should fix VCMSA-151


lucamar (commented):
Thanks @rsarm

lucamar (approved):
Thanks @rsarm

PR #2856, opened 2023-02-16
[daint-mc] Recipe for VASP 6.4.0 (multicore)

lucamar:
I provide the recipe for the latest relase 6.4.0 of VASP (multicore) and add it in production as non default.


ChristopherBignamini (approved):
lgtm

PR #2865, opened 2023-03-30
[daint, dom] spack-config: cleanup old versions

albestro:
~This new version of spack-config should fix a problem with a deprecated semantic for config files.~

~In particular, 1.7 was using a custom (and not updated anymore) spack version, that was generating config files with "target: any", which is not valid anymore.~

Actually, it was a regression https://github.com/spack/spack/issues/36628. For this reason I would take the chance, instead of updating, to just cleanup older versions of `spack-config`.


lucamar (commented):
Thaks @albestro, do you still need former versions 1.5 1.6 and 1.7 in the build list? If we remove them, the modules won't be removed from the system anyway

albestro:
> Thaks @albestro, do you still need former versions 1.5 1.6 and 1.7 in the build list? If we remove them, the modules won't be removed from the system anyway

I'm ok removing them 😉 AFAIK users should just use the default one, because it fixes previous problems (in general).

lucamar:
Ok to test

lucamar (approved):
Thanks @albestro

PR #2879, opened 2023-05-22
[daint-gpu] Add recipe of LAMMPS 28Mar2023

lucamar:
I provide the recipe to build the CUDA aware latest feature release of LAMMPS (28 Mar 2023).


simonpintarelli:
This seems to work for me: https://gist.github.com/simonpintarelli/30ef74cec9be6154e418f60d01ddf6achttps://gist.github.com/simonpintarelli/30ef74cec9be6154e418f60d01ddf6ac

I had to remove `-C .../most.cmake` (caused cmake processing error during configuration).

lucamar:
> Thank you. Yes, your recipe compiles. But I think it's not the right version, it's March 2022 not 2023.

Thanks, you're right, I had overlooked that the version is the old one. I have changed it now

simonpintarelli (approved):
LGTM

PR #2903, opened 2023-08-24
[eiger pilatus] Add GROMACS and PLUMED 

lucamar:
I provide the recipes for the latest release of GROMACS and PLUMED with the default toolchain cpeGNU (see [SD-59238](https://jira.cscs.ch/browse/SD-59238)).


lucamar:
The build failed due to a [known issue|https://jira.cscs.ch/browse/VCUE-99] with the configuration of `TestingEB` that does not prepend the local `MODULEPATH` correctly with Lmod: 
```
 Lmod has detected the following error: These module(s) or extension(s) exist but cannot be loaded as requested: "PLUMED/2.9.0"
 ```
 The local builds on the system are successful instead.

sekelle (approved):

PR #2909, opened 2023-09-25
[daint] remove options creating conflict with ParaView build

jfavre:
after discussion with @omlins , we agreed to remove the options which cause configuration errors with ParaView"


lucamar (approved):
Thanks @jfavre. Please let me remind you and @omlins that you need to force the rebuild of `adios-2.9.0-CrayGNU-21.09.eb` on the system in order to apply the changes, since the module `adios/2.9.1-CrayGNU-21.09` already exists (non default).

PR #2915, opened 2023-10-20
[daint dom] Julia 1.9.3 for CrayGNU-21.09

omlins:


lucamar (changes requested):
Thanks @omlins, I see that we would have 4 different releases of `Julia` in the build list:
```
 Julia-1.6.3-CrayGNU-21.09.eb                           --set-default-module
 Julia-1.7.2-CrayGNU-21.09.eb
 Julia-1.8.5-CrayGNU-21.09.eb
 Julia-1.9.3-CrayGNU-21.09.eb
 ```
 Do you need all of them? If you want to keep `1.6.3` as default, I would suggest to remove the intermediate ones `1.7.2` and `1.8.5`. Please note that the corresponding modules on the system won't be removed, unless you do that manually.

lucamar (approved):
Thanks @omlins: let's merge as it is for the time being, but we need to remove the old releases as soon as JupyterLab tests successfully with the latest Julia 1.9.3

PR #2926, opened 2023-11-08
[dom] Fix recipe of CDI on Dom

lucamar:
I have removed the dependencies below as the corresponding modules are already loaded by ecCodes:
```
#    ('cray-hdf5-parallel', EXTERNAL_MODULE),
#    ('cray-netcdf-hdf5parallel', EXTERNAL_MODULE),
```


jfavre (approved):
good catch. Thanks

PR #2930, opened 2023-11-09
[daint dom] Revert ecCodes dependencies

lucamar:
Revert to `cray-hdf5` and `cray-netcdf` to avoid the issues reported in [the regression tests](https://jenkins.cscs.ch/blue/organizations/jenkins/reframe-daint-production-daily/detail/reframe-daint-production-daily/2674/pipeline) for CDO and ParaView.


jfavre (commented):
I am afraid this will not be compatible with ADIOS which does require cray-hdf5-parallel. I need to do more testing

lucamar:
I have provided `ecCodes-parallel` in #2929 that loads the parallel libraries and therefore sold be compatible with `ADIOS`. The new module is already loaded by `ParaView` 5.12 on the system.

PR #2938, opened 2023-12-18
[daint,dom] Julia-1.9.4-CrayGNU-21.09

omlins:


omlins:
This julia recipe will create the module to be used in the jupyter recipes that are updated in a next step.

lucamar (approved):
Thanks @omlins: please note that you need to remove manually the modulefile `Julia/1.9.3-CrayGNU-21.09` from the system, as removing the recipe from the production list does not remove the modulefile

PR #2945, opened 2024-01-09
[daint-mc dom-mc] adding a new recipe for daint-mc ParaView5.12

jfavre:

PR #2948, opened 2024-01-12
[daint dom] Update git-lfs active recipes and deprecate outdated ones

lucamar:
Add command `git lfs install` to be executed upon loading the modulefile in orderto initialize `git lfs` (see my comment in #2945).


jfavre (approved):
Thanks @lucamar

PR #2949, opened 2024-01-19
[eiger pilatus] Update singularity-3.6.4-eiger.eb

lucamar:
Fix Lua footer for singularity module to avoid the following Lmod warning:
```
Lmod Warning:  Syntax error in file: /capstor/apps/cscs/eiger/UES/modulefiles/singularity/3.6.4-eiger.lua
 with command: prepend_path, one or more arguments are not strings.
```
The warning is due to the missing definition of `CRAY_LD_LIBRARY_PATH`, since the user environment does not load the Cray PE any longer by default upon login, but the user needs to do that with the `module load cray`.


teojgo (approved):
Thanks @lucamar

PR #2955, opened 2024-02-08
[daint-mc dom-mc] Remove non-default jupyterlab from daint-mc

lucamar:
Remove recipe `jupyterlab-3.4.5-CrayGNU-21.09-batchspawner.eb` from the list `7.0.UP03-21.09-daint-mc`, since the build fails on `dom-mc`.


omlins (approved):
LGTM

omlins:
Thanks @lucamar !

twrobinson:
Hi guys, what is the context here? Is it failing because it needs the fixes I put in the gpu version?

PR #2956, opened 2024-02-09
[dom-mc] jupyterlab-3.4.5-CrayGNU-21.09-batchspawner.eb

twrobinson:
Change the path to yarn temp file to avoid failures resulting from stale cache files. Increase the memory for nodejs.
Downgrade mistune package for compatibility.


lucamar (approved):
Thanks @twrobinson

PR #2973, opened 2024-03-12
[daint dom eiger pilatus] Files for CPE 23.05

lucamar:
I provide the configuration files required for CPE 23.05 currently installed on Pilatus, in view of the software stack update. 
Therefore, I need to detach the production list of Pilatus from Eiger, removing some builds that currently are not working (e.g.: `Scalasca`, due to the `PDT` dependency of `Score-P`). 
I also provide an updated `production.sh` to use a more recent EasyBuild release on Eiger and Pilatus.


jfavre (approved):
lgtm. Thanks @lucamar

PR #3001, opened 2024-05-07
[pilatus] Update CP2K architecture file on Pilatus

lucamar:
I provide an updated architecture file for CP2K 9.1 on Pilatus and the recipes of the dependencies (see [SPCI-265](https://jira.cscs.ch/browse/SPCI-265)).


lucamar:
The error is due to the download of the GMP package:
```
Couldn't find file gmp-6.2.1.tar.bz2 anywhere, and downloading it didn't work either
```

PR #2999, opened 2024-05-07
[daint, dom] Update adios-2.9.2-CrayGNU-21.09.eb

jfavre:
use the production version of Catalyst, instead of "rc4" version

PR #3009, opened 2024-05-08
[daint-mc] Remove versionsuffix from Catalyst dependency

lucamar:
Remove version suffix `-rc4` from `Catalyst` dependency of `ParaView` (see #2999): the [regression tests](https://jenkins.cscs.ch/blue/organizations/jenkins/reframe-daint-production-daily/detail/reframe-daint-production-daily/2879/pipeline) are successful (after the manual change of the corresponding modules on the system).


jfavre (approved):
thanks @lucamar

PR #3014, opened 2024-05-27
[daint-gpu] adding ParaView 5.12.1 as default for EGL-rendering

jfavre:
synching with the HDF5 provided by adios2 dependencies
removing NVIndeX no longer supported
making it the default


jfavre:
== COMPLETED: Installation ended successfully (took 39 mins 31 secs)
== Results of the build can be found in the log file(s) /scratch/snx3000/jfavre/daint/software/ParaView/5.12.1-CrayGNU-21.09-EGL/easybuild/easybuild-ParaView-5.12.1-20240527.214329.log

jfavre:
> Thanks @jfavre, I think that we can merge this pull request; are you sure to set as default the latest release `5.12.1` and not the older `5.11.2`?

yes, 5.12 should become the default. This is to encourage (or force) all users to move to 5.12 because on ALPS, this will be the only version available. Thanks for merging it.

PR #3029, opened 2024-07-18
[pilatus] Update jupyterlab-4.2.3-cpeGNU-23.12.eb

twrobinson:
Upgrade scipy as it was not compatible with the updated numpy

PR #3031, opened 2024-07-19
 [eiger] jupyterlab-4.2.3-cpeGNU-23.12

omlins:
Fix extra paths


lucamar (approved):
Please @omlins add the fix to #3029 as well: the latter is meant to fix the current JupyterLab module on Eiger, after testing on Pilatus.

PR #3040, opened 2024-07-26
[pilatus, eiger] making a custom Visit launcher to load modules

jfavre:
the internallauncher is modified to load the cpeGNU and Visit modules required for terminal-less client-server connections


lucamar (approved):
Thanks @jfavre

PR #3050, opened 2024-11-19
[eiger] Fix folder of ParaView recipe

lucamar:
The recipe was uploaded in the wrong folder.


jfavre (approved):
Thanks @lucamar for spotting this issue.

PR #3051, opened 2025-02-20
[eiger pilatus] add automake-1.16 for cpeGNU-23.12

finkandreas:


finkandreas:
I bumped m4 dependency to 1.4.19, since 1.4.18 failed building. I did not have any build failures doing it, and I think autoconf should also be happy with a newer m4 at runtime.

lucamar:
ok to test

teojgo (changes requested):
@finkandreas Can you add the 1.4.19 M4 in the PR?

finkandreas:
It is already in the master branch, I did not create or change it:
https://github.com/eth-cscs/production/blob/master/easybuild/easyconfigs/m/M4/M4-1.4.19-cpeGNU-23.12.eb

I probably should remove the M4-1.4.18 from the PR, since it's unused :see_no_evil:

teojgo:
ok to test

teojgo (approved):

